PR #23, opened 2026-07-21
Add cooktop and range hood support

themaanda:


custom_components/localthings/registry/capabilities/cooktop.py
```diff
@@ -0,0 +1,192 @@
...
+            icon='mdi:stove',
+        ),
+    ),
+)
+
+
+# NA9300K exposes physical slots 0, 1, 3, 4, and 5.  Preserve Samsung's slot
+# numbers instead of guessing at front/back/left/right placement.  A user can
+# safely identify and rename each entity by briefly lighting one burner.
+_NA9300K_OPERATION_SLOTS = (0, 1, 3, 4, 5)
```
mbillow: `_operation_slots(options)` above already derives the live set of burner slots from `x.com.samsung.da.options` at runtime (it's used for `any_burner_active`), but the per-slot `SensorDesc` entities here are still built from this hardcoded tuple at import time. Each entity does have an `exists_fn` that hides it when its slot is absent, so a device reporting *fewer* burners degrades fine — but a cooktop reporting burner slots outside `(0, 1, 3, 4, 5)` (e.g. a 6-zone model) would silently get no entity for those slots, and this wouldn't trip the unbound-href coverage check since `/mode/vs/0` is still claimed by `COOKTOP_MODE`.

Not a blocker for this PR (only one device is verified so far), but worth a short comment flagging the limitation, and/or widening this to a generous static superset (e.g. `range(0, 8)`) so `exists_fn` has more headroom to hide/show correctly on other models. Filed the real fix (a data-driven `entities_fn` hook so the registry can generate the right number of entities from the live resource instead of guessing a bound) as #31.

---

custom_components/localthings/registry/capabilities/range_hood.py
```diff
@@ -0,0 +1,295 @@
...
+            value_fn=_active_alarm_codes,
+        ),
+    ),
+)
+
+
+def _hood_fan_write(payload, rep, href=None):
+    kind, value = payload
+    if kind == 'power':
+        return ['power', '0'], {'value': bool(value)}
```
mbillow: This unconditionally PUTs to `/power/0`. The fan's read side (`fan.py`'s `is_on`) checks `/power/0` first and falls back to `/power/vs/0` if the OCF-standard resource is absent, and that fallback pair is the established house style elsewhere (`common.POWER_GENERIC`/`POWER_VS_FALLBACK`). On the verified hood both resources exist so this works today, but a hood variant exposing only `/power/vs/0` would have its power-on/off writes silently fail (PUT to a resource that doesn't exist) with no error surfaced to the user.

Worth mirroring the read-side fallback here — check which of `/power/0` / `/power/vs/0` is present on `rep`/`resources` and target whichever one the device actually has.

---

mbillow (commented):
Nice work — this is a well-scoped, well-tested addition that follows the registry/Capability/discovery conventions closely and reuses existing shared capabilities (`fridge.FIRMWARE_UPDATE`, `common.ENERGY_METER`) rather than reinventing them.

**Strengths:**
- Cooktop deliberately ships **no write path** for burners/power, with a docstring explaining why ("must not be remotely ignited by an automation") and a dedicated test asserting zero `write_fn`s across every cooktop entity — genuinely good safety-conscious design.
- The new range hood's lamp switch write correctly compares `payload == 'On'` — actually more correct than a few existing `fridge.py` switches (`ICEMAKER_NIGHTTIME`, `STATUS_LOCK`, `DEFROST_DELAY`) that use `'On' if p else 'Off'` against a payload that `switch.py` always passes as a non-empty string, so those always encode "On" regardless of which way the user toggled. Pre-existing bug elsewhere, not introduced here — just noting the new code avoided it.
- Test coverage matches (in places exceeds) the bar set by #24: scrubbed fixtures, golden state-key fixtures, end-to-end `discover()`+`flatten()` tests asserting zero unbound hrefs for both new families, entity-layer tests for the new `FanEntity`, and write-contract assertions including rejection of unsupported values.
- `for_device_by_resources`'s heuristics are appropriately conservative (require multiple co-occurring markers) and guarded by a negative test against an oven-shaped `/mode/vs/0`.

**Left two inline comments** on things worth addressing before merge:
1. `range_hood.py` — the fan's power write lacks the `/power/0` → `/power/vs/0` fallback that the read side and the rest of the codebase establish.
2. `cooktop.py` — the hardcoded burner-slot tuple means a differently-configured cooktop could silently lose entities for its own layout without tripping the coverage check.

Filed two follow-up issues that came out of digging into this: #31 (framework support for dynamically generating entities from array-encoded multi-slot resources, which would be the real fix for the burner-slot limitation) and #32 (an existing `FLEX_ZONE` bug on TP1X-class fridges, unrelated to this PR but surfaced while comparing conventions against #24's fixture — not blocking here, just tracking it separately).

Minor/non-blocking: `COOKTOP_POWER`'s boolean On/Off field is modeled as a `SensorDesc` rather than `BinarySensorDesc`, inconsistent with the pattern used one capability later in the same file (`COOKTOP_CONNECTED`). And `range_hood._active_alarm_codes` forks `common`'s near-identical helper rather than extending it — justified by fixture evidence, but worth a cross-reference comment or hoisting if other families hit the same "deleted row" noise later.

No security concerns — no `eval`/`exec`, no new deserialization paths, fixtures properly scrubbed of serials/device IDs.

---

themaanda:
Thanks for the review! 

Both concerns have been addressed. Hood power writes now support both power resources, and cooktop burner slots have been widened with live-device gating. All 395 tests and GitHub checks pass.

mbillow:
Thanks! I just cut a release, so this will likely go out with some other fixes tomorrow.

PR #34, opened 2026-07-22
test: remove timing race from coordinator observe-mode tests

aaearon:
Several coordinator observe-mode tests delivered the subscription notification from a background thread with a wall-clock `time.sleep(0.005)`, racing it against the coordinator's own update cycle. When that cycle (failed poll -> session close -> retry -> downgrade -> cache sweep -> executor hop) took longer than the sleep, `try_enter_observe_mode` had already cleared `_notified` before the notify landed, so the notification was dropped and the mode stayed `poll` -> intermittent test failures.

Fix: `FakeObserveSession` gained an opt-in `notify_on_subscribe` mapping; `subscribe()` now delivers the representation synchronously via `on_notification` (inside the grace-period window by construction). The thread + sleep blocks were removed. Tests that need "no notification" set it to `None`.

Scope: test-only, no production/behavior change to the coordinator or observe logic. Existing assertions are unchanged, not weakened.

Verification: `pytest tests/localthings/test_coordinator.py tests/localthings/test_observe.py -q` passes. Full `pytest tests/ -q` passes except one unrelated, pre-existing socket test that is environment-specific.

### DIFF
--- a/tests/localthings/conftest.py
+++ b/tests/localthings/conftest.py
@@ -5,6 +5,7 @@ import json
 from pathlib import Path
 from unittest.mock import patch
 
+import cbor2
 import pytest
 from pytest_homeassistant_custom_component.common import MockConfigEntry
 
@@ -150,12 +151,24 @@ class FakeObserveSession:
         self.subscribed: list[str] = []
         self.fail_hrefs: set[str] = set()
         self.closed = False
+        # When set to a rep dict, subscribe() immediately delivers that rep
+        # as an OBSERVE notification for the href — what a real device does
+        # when it answers a subscription with the current representation.
+        # `try_enter_observe_mode` clears its notified set before it
+        # subscribes, so this is the only way to deliver a notify that
+        # reliably lands inside the grace period: a notify raced in from
+        # another thread can be wiped by that clear (or arrive after the
+        # grace period ends) depending on scheduling. Set to None to model
+        # a device that answers subscriptions but never notifies.
+        self.notify_on_subscribe: dict | None = None
 
     def subscribe(self, path_segs):
         href = '/' + '/'.join(path_segs)
         if href in self.fail_hrefs:
             raise ConnectionError("subscribe failed")
         self.subscribed.append(href)
+        if self.notify_on_subscribe is not None and self.on_notification is not None:
+            self.on_notification(href, cbor2.dumps(self.notify_on_subscribe))
         return b'\x01'
 
     def refresh_observes(self, paths):
--- a/tests/localthings/conftest.py
+++ b/tests/localthings/conftest.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import json
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


mbillow (commented):
## Review: test: remove timing race from coordinator observe-mode tests

### Overview
Test-only change. Replaces the background-thread + `time.sleep(0.005)` notification pattern in six coordinator observe-mode tests with an opt-in `FakeObserveSession.notify_on_subscribe` mapping that delivers the OBSERVE representation synchronously from inside `subscribe()`. Net −46 lines, assertions unchanged.

### Correctness — the fix is sound
- `ObserveManager.try_enter_observe_mode` runs: `_notified.clear()` → subscribe loop → `time.sleep(grace_period_s)` → `fraction = len(set(_notified) & subscribed) / len(subscribed)`. A notify only counts if it lands in the window *after* the clear and *before* the fraction check.
- The old thread fired the notify on a wall-clock delay racing the coordinator's own update cycle (failed poll → reconnect → downgrade → sweep → executor hop). When that cycle outran the 5 ms sleep, the notify was dropped → intermittent failures. Real, not theoretical.
- Delivering the notify synchronously inside `session.subscribe()` (line 144, after the clear at 139, before the sleep at 154) lands it in the window deterministically. This is also *more faithful* to a real OCF device, which answers a subscribe with the current representation.
- `notify_on_subscribe = None` for the two "no notify on resubscribe" cases correctly models a device that answers subscriptions but never notifies, so mode stays `poll`. Sequencing is right: set `{'notified': True}` to enter observe, then reset to `None` before the reconnect.
- Import cleanup is safe: every `threading`/`time` reference in `test_coordinator.py` was inside a removed block (the only remaining hit is the word "time" in a docstring); `cbor2` is retained and still used.

### Non-blocking observations
1. **href normalization (informational, not a defect).** The fake now delivers the notify with a reconstructed href `'/' + '/'.join(path_segs)`, while `subscribed` keys on the original href string. These intersect only if the hrefs are canonical (no trailing/double slashes, no empty segments). `_hot_hrefs + _warm_hrefs` are canonical, so it holds — and it actually mirrors what the real session would report. Worth being aware of if non-canonical hrefs ever enter those lists.
2. **`test_observe.py` still races.** The same `threading.Thread` + `time.sleep(0.005)` "notify during grace period" pattern this PR removes remains throughout `test_observe.py` (e.g. lines ~86, 136, 191, 209, 242, 281, 310). Those tests drive `ObserveManager` directly without the coordinator update cycle contending, so they're less exposed, but the underlying wall-clock dependency is identical. Reasonable to leave out of this PR (scope is coordinator tests), but a good follow-up to apply the same `notify_on_subscribe` approach there.
3. **Type hint.** `dict[str, Any] | None` is fine; `Any` is just the rep passed to `cbor2.dumps`.

### Verdict
Clean, well-documented, low-risk. Removes genuine flakiness without weakening any assertion and with a more realistic device model. No changes requested.

---

Issue #32, opened 2026-07-22
fridge: FLEX_ZONE misbinds on TP1X hardware and corrupts x.com.samsung.da.modes on write; CV_FDR_* flex-zone content selector unmodeled


mbillow:
### Summary

`registry/capabilities/fridge.py`'s `FLEX_ZONE` capability is scoped in comments/docstring to RF9000-class fridges (`CV_TTYPE_RF9000A_*` mode tokens), but its `exists_fn` doesn't actually check for that prefix — only that `x.com.samsung.da.supportedOptions` is non-empty. Since `/mode/vs/0` has exactly one capability registered against it in `refrigerator.py`, `FLEX_ZONE` unconditionally claims that href for *every* fridge, including TP1X-class hardware, which uses a different, non-overlapping mode-flag scheme.

### Evidence

The TP1X_REF_21K fixture added in #24 (`tests/fixtures/refrigerator_tp1x_ref_21k_us_device.json`) reports:

```json
"x.com.samsung.da.modes": ["CVN_CONVERTIBLE_ZONE", "CV_FDR_WINE", "WATERFILTER_DISABLE"],
"x.com.samsung.da.supportedOptions": ["CV_FDR_WINE", "CV_FDR_DELI", "CV_FDR_BEVERAGE", "CV_FDR_MEAT", "CV_FDR_SOFT_FREEZER"]
```

None of these start with `CV_TTYPE_RF9000A_`. Tracing through the current code:

```python
FLEX_ZONE = Capability(
    href='/mode/vs/0',
    entities=(
        SelectDesc(key='flex_zone_mode', ...,
                   exists_fn=lambda rep, resources: bool(rep.get('x.com.samsung.da.supportedOptions')),
                   value_fn=lambda modes: next((m for m in (modes or [])
                                                 if m.startswith('CV_TTYPE_RF9000A_')), None),
                   write_fn=_flex_zone_write),
    ),
)

def _flex_zone_write(p, rep, href=None):
    modes = list(rep.get('x.com.samsung.da.modes') or [])
    modes = [m for m in modes if not m.startswith('CV_TTYPE_RF9000A_')]
    modes.append(p)
    return ['mode', 'vs', '0'], {'x.com.samsung.da.modes': modes}
```

On TP1X hardware:
[...]

PR #35, opened 2026-07-22
feat: add AI energy level select and self-check error sensor for TP1X refrigerator

aaearon:
Adds two small, additive fridge capabilities that were left on the table after #24 added TP1X support.

## What's added

- **AI energy level select** (`/energy/ailevel/vs/0`, `ai_energy_level`): exposes the fridge's AI energy-saving level as a select. This resource was previously in the global ignore list because every dump seen so far only reported one entry in `supportedAiLevel` (no real choice to offer). The capability now binds the href but gates the entity behind `supportedAiLevel` actually containing more than one value, so hardware with only a single level continues to fall through cleanly with no entity created, while hardware that does offer multiple levels gets a usable select instead of a permanently-ignored resource.
- **Self-check error sensor** (existing `SELF_CHECK` capability, `selfcheck_error`): adds a diagnostic sensor for `x.com.samsung.da.error`, the error-code list from the last self-check run. It's gated by `exists_fn` since not every fridge reports the field. Combined with the existing `selfcheck_status`/`selfcheck_result`/`selfcheck_start` entities, this gives a full picture of self-check state instead of an incomplete one.

## Scope

This is purely additive on top of #24. It intentionally does not touch `FLEX_ZONE` / `/mode/vs/0` (that's #32, already fixed on `main`) or `/icemaker/status/0`.

## Test coverage

- `selfcheck_error` is exercised against the real captured `refrigerator_tp1x_ref_21k_us` fixture — that dump's `/selfcheck/vs/0` does include `x.com.samsung.da.error: ["DA_ERROR_NONE"]`, so this is real-fixture coverage. The golden for that fixture (`tests/fixtures/golden/refrigerator_tp1x_ref_21k_us.json`) has been regenerated via real discovery to include the new key, and `TestTp1xNativeDuplicateResources.test_us_fixture_has_complete_coverage` now asserts the value.
- `ai_energy_level` is synthetic-only: the real TP1X_REF_21K_US dump's `supportedAiLevel` only has one entry (`["1"]`), so `exists_fn` correctly hides the select on that hardware and it can't be exercised through the real fixture. The existing scrubbed `tests/fixtures/refrigerator_device.json` synthetic fixture already had a single-entry `/energy/ailevel/vs/0` resource (to exercise the "hidden" path); I extended its `supportedAiLevel` to `["1", "2"]` so it also exercises the "shown" path, and regenerated `tests/fixtures/golden/refrigerator.json` accordingly. `TestAiEnergyLevel.test_synthetic_fixture_has_complete_coverage` documents and asserts this.
- New unit tests in `tests/test_fridge_capabilities.py` (`TestSelfCheckError`, `TestAiEnergyLevel`) cover the `exists_fn` gating, `value_fn`/`write_fn` behavior, and the href for both capabilities directly, plus zero-unbound-hrefs checks against the fixtures above.

`pytest tests/ -q`: 290 passed, 1 failed (`tests/localthings/test_config_flow.py::test_find_live_ports_detects_silent_port`, the pre-existing WSL loopback-socket flake — unrelated to this change).

May need a trivial rebase once other open fridge/energy work lands, but there's currently nothing open that overlaps these two hrefs.

### DIFF
--- a/custom_components/localthings/registry/by_type/refrigerator.py
+++ b/custom_components/localthings/registry/by_type/refrigerator.py
@@ -20,6 +20,7 @@ REGISTRY = DeviceRegistry(
         dishwasher.DIAGNOSIS,
         fridge.ICEMAKER_NIGHTTIME,
         fridge.FLEX_ZONE,
+        fridge.AI_ENERGY_LEVEL,
         fridge.REFRIGERATION,
         fridge.AUTOFILL,
         fridge.WELCOME_LIGHTING,
--- a/custom_components/localthings/registry/capabilities/fridge.py
+++ b/custom_components/localthings/registry/capabilities/fridge.py
@@ -242,6 +242,14 @@ SELF_CHECK = Capability(
         SensorDesc(key='selfcheck_result', field='x.com.samsung.da.result',
                    name='Self-check result', icon='mdi:clipboard-check-outline',
                    entity_category='diagnostic'),
+        # List of error codes from the last self-check; joined for display.
+        # Not every fridge reports the field, hence the exists_fn.
+        SensorDesc(key='selfcheck_error', field='x.com.samsung.da.error',
+                   name='Self-check error', icon='mdi:alert-circle-outline',
+                   entity_category='diagnostic',
+                   exists_fn=lambda rep, resources: (
+                       not rep or 'x.com.samsung.da.error' in rep),
+                   value_fn=lambda v: ', '.join(v) if isinstance(v, list) else v),
         ButtonDesc(key='selfcheck_start', field='', name='Start self-check',
                    payload='Start', icon='mdi:play-circle-outline',
                    entity_category='diagnostic',
@@ -577,6 +585,31 @@ FLEX_ZONE = Capability(
                        rep.get('x.com.samsung.da.supportedOptions')),
                    rep_fn=_flex_zone_current,
                    write_fn=_flex_zone_write),
+    ),
+)
+
+# ---------------------------------------------------------------------------
+# AI energy-saving level
+#
+# supportedAiLevel is a single-entry list ('1' only) on some hardware, where
+# a select would offer no real choice — gated to >1 supported level.
+# ---------------------------------------------------------------------------
+
+AI_ENERGY_LEVEL = Capability(
+    href='/energy/ailevel/vs/0',
+    poll_tier='cold',
+    entities=(
+        SelectDesc(key='ai_energy_level', field='aiLevel',
+                   name='AI energy level', icon='mdi:leaf',
+                   translation_key='ai_energy_level',
+                   entity_category='config',
+                   options_field='supportedAiLevel',
+                   exists_fn=lambda rep, resources: (
+                       not rep or (
+                           isinstance((sl := rep.get('supportedAiLevel')), (list, tuple))
+                           and len(sl) > 1)),
+                   write_fn=lambda p, rep, href=None: (
+                       ['energy', 'ailevel', 'vs', '0'], {'aiLevel': p})),
     ),
 )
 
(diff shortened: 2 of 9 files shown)


mbillow (commented):
## Review: AI energy level select + self-check error sensor (TP1X fridge)

Reviewed the full diff against the surrounding registry/discovery code. This is a clean, well-scoped, purely-additive change with strong test coverage. No blocking issues found.

### Overview
- Promotes `/energy/ailevel/vs/0` from the global ignore list to a real `AI_ENERGY_LEVEL` capability exposing an `ai_energy_level` select, gated behind `supportedAiLevel` having >1 entry so single-level hardware creates no entity.
- Adds a `selfcheck_error` diagnostic sensor (`x.com.samsung.da.error`) to the existing `SELF_CHECK` capability, gated by `exists_fn` and joining the error-code list for display.
- Updates `strings.json` / `translations/en.json`, and regenerates both affected goldens.

### Correctness — verified
- **"Falls through cleanly" claim holds.** In `discovery.py`, an href is marked `matched` as soon as any capability is registered for it (before `exists_fn` runs), so moving this resource out of `ignored.py` into a real capability keeps it from being logged as unbound even when the entity is gated off. `flatten()`/discovery apply `exists_fn` afterward to hide the entity. Confirmed the TP1X golden correctly gains only `selfcheck_error` (single AI level → no `ai_energy_level`), while the synthetic `refrigerator.json` golden gains both.
- **`exists_fn` gating is robust.** The AI-level guard `isinstance((sl := rep.get('supportedAiLevel')), (list, tuple)) and len(sl) > 1` correctly rejects a stray scalar (a 5-char string would otherwise pass `> 1`), and the `not rep` empty-stub carve-out matches the established `/device/0` not-yet-fetched pattern used elsewhere. Both edge cases are explicitly tested.
- **Conventions followed.** `translation_key` + `options_field` select shape matches `BEVERAGE_ZONE`/`PANTRY_ZONE`; the state-translation block is correctly nested under `select` (not `sensor`) and `_display()` lowercasing of `'1'`/`'2'` lines up with the string keys; `write_fn` signature matches the module's other selects.

### Test coverage
Excellent. New `TestSelfCheckError` / `TestAiEnergyLevel` cover `exists_fn` (present/absent/empty-stub/non-list-scalar), `value_fn` list-join and scalar passthrough, href, and `write_fn`, plus zero-unbound-hrefs assertions against both the real TP1X fixture and the synthetic fixture. The synthetic-vs-real-fixture split is well documented in the PR body and test docstrings.

### Minor / non-blocking suggestions
1. **`strings.json` only defines levels `"1"` and `"2"`.** If real hardware ever surfaces `supportedAiLevel` with a `"3"+` entry, that option will render as its raw value. Fine for now given no such dump exists, but worth a note next to the strings block.
2. **`value_fn=lambda v: ', '.join(v) ...` on an empty list** yields an empty display string rather than something like `DA_ERROR_NONE`. Cosmetic only; devices seem to report `["DA_ERROR_NONE"]` rather than `[]`.
3. **Walrus-in-lambda** (`(sl := rep.get(...))`) is a touch dense versus the simpler `exists_fn`s elsewhere in this module, though the dedicated test documents the intent well. Purely stylistic.

Note: the PR mentions one pre-existing unrelated test flake (`test_find_live_ports_detects_silent_port`, WSL loopback socket) — not introduced by this change.

Overall: LGTM.

---

mbillow:
None of the feedback is blocking, just nits. Happy to merge if you don’t think changes are useful.

aaearon:
Thanks for the thorough review! Incorporated the two actionable nits in 0cc9d8e:

1. **AI energy level labels** — added a comment next to the `AI_ENERGY_LEVEL` `SelectDesc` noting that only levels `"1"`/`"2"` have translated labels in strings.json/translations, and a higher `supportedAiLevel` entry will render as its raw value until a label is added.

2. **`selfcheck_error` empty list** — `value_fn` now returns `None` (no value / unknown) for an empty error list instead of an empty display string. Went with `None` rather than synthesizing a `DA_ERROR_NONE` string so we don't invent a code the device didn't report — as you noted, real devices seem to send `["DA_ERROR_NONE"]` rather than `[]`, so this is just belt-and-suspenders. Added a test for the empty-list case.

Left the walrus in `ai_energy_level`'s `exists_fn` as-is per your note (the dedicated test documents the intent). No goldens changed.

Issue #17, opened 2026-07-21
Samsung AC unit


edo2313:
### Appliance model

ARTIK051_PRAC_20K

### What's missing or wrong

Hi, first of all thanks for this amazing integration! 🥇 
I attached the diagnostics file for the AC that I have, let me know if you need the SmartThings app screenshots (there isn't a lot of data there, it seems that the diagnostic json is already pretty self-explanatory)

### Diagnostics file

[config_entry-localthings-01KY1RRB96TVSPK9Q0TT6RXB6J.json](https://github.com/user-attachments/files/30221622/config_entry-localthings-01KY1RRB96TVSPK9Q0TT6RXB6J.json)

mbillow:
I’ve got something that functionally should work, just need to code review and do some cleanup. Thanks for providing this! 

Hopefully I should have a release cut for you to try by the end of the day.

mbillow:
Can you give 0.7.0 a spin and see how it works?

edo2313:
Thanks for the update, seems to work fine!
The only strange thing (if I understant correctly how the underlying library works) is that there is a pretty long delay (20/30 secs maybe? I need to measure it) between the state change and the new values in homeassistant, altough the commands are instant.
E.g: I turn it on in homeassistant -> the AC turns on istantly -> In hass i still see it off for about 20/30 seconds

mbillow:
Thanks for confirming v0.7.0 worked and for flagging the lag! Found and fixed it in **v0.10.0**: the AC's power/temperature/fan/swing/preset state was only refreshing on the ~30s full-device poll instead of getting the same instant push updates the rest of the climate entity gets — that's exactly the 20-30s delay you saw between a command landing and HA showing it. It's now on the same fast-refresh tier as the rest of the climate entity.

Please update to v0.10.0 — commands should reflect in HA right away now.

---
_Generated by [Claude Code](https://claude.ai/code)_

DevineCZ:
I am using the latest version 0.10.1 and unfortunately the error with the delay in reporting the on/off status or temperature change is still there. The air conditioner responds immediately to turning it on, off, changing the temperature, or changing the fan speed, but the status changes after more than 30 seconds. According to my estimate, after 40-60 seconds.

mbillow:
[...]

PR #36, opened 2026-07-22
Fix AC coverage for TP1X-class firmware (oneUiVersion "7.0 Air conditioner")

paddawan2002:
Fixes the air-conditioner coverage gap on newer Samsung AC firmware (Tizen Lite, `oneUiVersion` "7.0 Air conditioner"; e.g. model `TP1X_DA-AC-RAC-01011`). Addresses #17 for this model class.

### Problem

This firmware does **not** expose the OCF-standard `/temperature/current/0` + `/temperature/desired/0` pair that the AC registry (written for the older `ARTIK051_PRAC` unit in #17) reads from. Instead it reports temperature under the vendor `/temperatures/vs/0` resource, whose `x.com.samsung.da.items[0]` carries current/desired/minimum/maximum/increment/unit — a resource the registry previously **ignored** as a duplicate.

Result on these units: `current_temperature` and `target_temperature` are `null`, and 14 resources are unbound, so the "device_gap" (incomplete capability coverage) repair fires.

### Changes

- **`climate.py`** — temperature (`current`/`target`/`range`/`step`/`unit`) falls back to `/temperatures/vs/0` `items[0]` when the OCF-standard hrefs are absent. Pure fallback: older units that expose the OCF pair are unaffected.
- **`registry/capabilities/airconditioner.py`** — `/temperatures/vs/0` moved from ignored to a consumed climate sibling; added a **Display light** switch on `/light/vs/0`; the 13 remaining TP1X-only housekeeping/opaque hrefs added to the AC-scoped ignore list with one-line reasons.
- **`registry/by_type/airconditioner.py`** — registers the new `DISPLAY_LIGHT` capability.

### DIFF
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -41,6 +41,7 @@ from .registry.capabilities.airconditioner import (
     HREF_TEMP_CURRENT as TEMP_CURRENT_HREF,
     HREF_TEMP_DESIRED as TEMP_DESIRED_HREF,
     HREF_TEMP_CONTROL as TEMP_CONTROL_HREF,
+    HREF_TEMPS_VS as TEMPS_VS_HREF,
     HREF_WIND_STRENGTH as WIND_STRENGTH_HREF,
     HREF_WIND_DIRECTION as WIND_DIRECTION_HREF,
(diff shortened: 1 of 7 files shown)


mbillow:
Thanks for the contribution!

PR #65, opened 2026-07-24
Feature/add completion time

Metal-Eagle:
First of all, thanks for the nice work on this! I had some things that were missing so I added them.

## 🛠️ Description

This PR introduces new completion time sensors for Samsung DA appliances within operational capabilities, updates washer/dryer cycle definitions with missing course codes, and updates unit test coverage along with golden regression fixtures.

### Key Changes:

* **Operational Capability (`operational.py`):**
  * Added `_parse_remaining_time`, `_completion_time`, and `_completion_minutes` helper functions to parse cycle remaining time from `x.com.samsung.da.remainingTime` or `remainingTime`.
  * Added `completion_time` sensor entity (`mdi:timer-sand`).
  * Added `completion_minutes` sensor entity (`mdi:clock-outline`, unit: `min`, device class: `duration`) with automatic minute rounding for remaining seconds.

* **Localization & Translations (`strings.json` & `en.json`):**
  * Added translation mappings for new laundry cycle course codes (missing on the Samsung WW11DB8B95GB washer):
    * `2b` -> **AI Wash**
    * `2a` -> **Jeans**

* **Test Coverage & Golden Fixtures (`tests/`):**
  * Added unit test coverage for new course codes (`2b`, `2a`) in `test_laundry_capabilities.py`.
  * Added `TestCompletionTime` test suite in `test_operational_capability.py` covering string parsing, fallback keys, minute conversions, and empty payload handling.
  * Updated `golden` state keys fixtures for appliances featuring operational state capabilities (`dishwasher`, `dryer`, `washer`, `washer_dryer_combo`, `washer_flexwash`).

And again, thanks for the great work you are doing!


mbillow:
This is already exposed, or it should be through the “finish_time” sensor. What benefits come from exposing this information in triplicate? Are you not seeing finish time on your device?

Thanks for the additional cycle translations, those will be helpful regardless of the above questions!

Metal-Eagle:
No, the finish time only shows while the washer is running, but I'd like to see the estimated completion time before starting the program.

Maybe we could add this as a disabled-by-default sensor so users can enable it if needed? Or alternatively, we could update the existing Estimated finish sensor so it doesn't turn into Unknown when idle.

custom_components/localthings/registry/capabilities/operational.py
```diff
@@ -70,6 +70,34 @@ def _finish_time(remaining_str):
...
 
+def _parse_remaining_time(rep):
+    return (
+        rep.get('x.com.samsung.da.remainingTime')
+        or rep.get('remainingTime')
+    )
+
+
+def _completion_time(rep):
+    return _parse_remaining_time(rep)
```
mbillow: Two naming/indirection things worth tidying here:

- `_parse_remaining_time` doesn't actually parse — it's a `.get()` with a fallback key. Something like `_remaining_time_raw` reads truer.
- `_completion_time` is a pure pass-through (`return _parse_remaining_time(rep)`), so it's two names for one function. The `rep_fn=_completion_time` below could just point at the get-with-fallback directly and drop this wrapper entirely.

Related: the `remainingTime` (non-prefixed) fallback key — do any of the DA_WM_ boards actually report it that way? The existing `finish_time` only reads `x.com.samsung.da.remainingTime`, so if the bare key never appears in practice it's dead-code fallback worth removing for consistency.

---

custom_components/localthings/registry/capabilities/operational.py
```diff
@@ -70,6 +70,34 @@ def _finish_time(remaining_str):
...
+        parts = [int(p) for p in raw.split(':')]
+        if len(parts) == 3:
+            h, m, s = parts
+            return h * 60 + m + (1 if s > 0 else 0)
+        elif len(parts) == 2:
+            m, s = parts
+            return m + (1 if s > 0 else 0)
+    except (ValueError, TypeError):
+        pass
+    return None
```
mbillow: This is the third hand-rolled `HH:MM:SS` splitter in this file — `_delay_hours` (line 38) and `_finish_time` (line 64) already do the same colon-split-and-int dance. Worth factoring the parse into one shared helper, e.g.:

```python
def _remaining_seconds(raw):
    """'HH:MM:SS' (or 'MM:SS') -> total seconds, or None."""
    if not isinstance(raw, str):
        return None
    try:
        parts = [int(p) for p in raw.split(':')]
    except ValueError:
        return None
    if len(parts) == 3:
        h, m, s = parts
    elif len(parts) == 2:
        h, m, s = 0, *parts
    else:
        return None
    return h * 3600 + m * 60 + s
```

Then `finish_time` and `completion_minutes` both build on it (`minutes = ceil(seconds / 60)`), and there's one place that knows the wire format. As-is, the two implementations can drift.

---

custom_components/localthings/registry/capabilities/operational.py
```diff
@@ -123,6 +151,15 @@ def _finish_time(remaining_str):
                             or rep.get('x.com.samsung.da.progress') == 'Finish'
                        else _finish_time(rep.get('x.com.samsung.da.remainingTime'))
                    )),
+        SensorDesc(key='completion_time', name='Completion time',
+                   icon='mdi:timer-sand',
+                   exists_fn=lambda rep, resources: _completion_time(rep) is not None,
+                   rep_fn=_completion_time),
```
mbillow: This entity overlaps heavily with the existing `finish_time` sensor three lines up. Both derive from the same `remainingTime` field; `finish_time` is a `timestamp` sensor, which the HA frontend already renders as a live relative countdown ("in 1h 25m") and which automations can compare against. `completion_time` exposes the same information as a raw `"01:25:30"` string with no device class — it can't be graphed, compared, or used numerically, so it's strictly less useful than the timestamp that already exists.

My suggestion is to **drop `completion_time`** and keep only `completion_minutes` below, which is a genuinely distinct representation (a plain number for gauges/automations). That takes us from three views of one field back to two meaningfully-different ones.

(If it stays, it needs the same active/`Finish` gating as `finish_time` — see the note on `completion_minutes`.)

---

custom_components/localthings/registry/capabilities/operational.py
```diff
@@ -123,6 +151,15 @@ def _finish_time(remaining_str):
...
+        SensorDesc(key='completion_minutes', name='Remaining minutes',
+                   icon='mdi:clock-outline', unit='min',
+                   device_class='duration', state_class='measurement',
+                   exists_fn=lambda rep, resources: _completion_minutes(rep) is not None,
+                   rep_fn=_completion_minutes),
```
mbillow: Functional bug: this (and `completion_time`) reads `remainingTime` unconditionally, but `finish_time` right above deliberately does **not**. Per the comment at lines 144–146:

> Samsung firmware leaves a stale remainingTime after a cycle ends, and freezes it at '00:01:00' when progress reaches 'Finish'.

So once a cycle finishes, `completion_minutes` will report `1` forever and `completion_time` will report `"00:01:00"` forever — exactly the stale-value trap `finish_time` is written to avoid. This needs the same guard: return `None` unless `machine_state` maps to `'active'` and `progress != 'Finish'`. Factoring that predicate out (both `finish_time` and this can share it) would keep the three time entities consistent.

---

mbillow (commented):
Thanks for this — the missing course codes and the desire for a "minutes remaining" number are both welcome, and the golden/unit-test coverage is appreciated. I've left inline notes; the summary of what I'd like to see before merge:

**1. Three representations of one field.** After this PR, `x.com.samsung.da.remainingTime` is surfaced by three entities: `finish_time` (existing timestamp), `completion_time` (raw string), and `completion_minutes` (int). That's a lot of overlap. `finish_time` already renders as a live countdown and is automation-friendly, so the raw-string `completion_time` doesn't add anything over it. I'd drop `completion_time` and keep just `completion_minutes` — see the inline note.

**2. Stale-value gating (the one real bug).** The new entities read `remainingTime` unconditionally, but `finish_time` deliberately gates on `active` + `progress != 'Finish'` because the firmware freezes `remainingTime` at `00:01:00` after a cycle ends. Without that guard the new sensors will show `1` / `00:01:00` forever post-cycle. Details inline.

**3. Helper shape.** `_parse_remaining_time` doesn't parse (it's a get-with-fallback), `_completion_time` is a pure pass-through wrapper, and `_completion_minutes` re-implements colon-splitting that two other helpers in this file already do. A single shared `_remaining_seconds()` parser plus a shared "is the cycle live?" predicate would remove the indirection and keep all three time entities from drifting.

**Smaller items:**
- `operational.py` and `test_operational_capability.py` both lose their trailing newline — worth restoring (likely a lint failure, and it's what's making CI `unstable`).
- The `2b` / `2a` translation additions are correct — lowercase matches the "translation keys are the lowercased id" convention (cf. the `1f` entry and the note in `laundry.py`). Only nit: the `parse_edit_course_list` test feeds *lowercase* `2b2a` into the raw string, but real firmware sends uppercase hex; it passes only because `hex_pairs` is case-preserving. Using `2B2A` in the test input would match what the device actually reports.

Happy to help with any of these if useful.

---

Metal-Eagle:
Thanks for the details! I tried to fix everything on my end, and for the parts I couldn't quite figure out, I used a bit of AI help since I'm still learning Python.

I've updated the PR to address all of your feedback:

1. **Entity Cleanup:** I dropped the redundant `completion_time` raw-string sensor entirely and kept only `completion_minutes`. 
2. **Helper Refactoring:** I removed the redundant wrappers and implemented a single, shared `_remaining_seconds()` parser. Now `delay_hours`, `finish_time`, and `completion_minutes` all share the exact same `HH:MM:SS` parsing logic. 
3. **Stale-Value Gating:** To fix the firmware bug where the time freezes at `00:01:00` post-cycle, `completion_minutes` now returns `0` when `progress == 'Finish'`. 
    * *(Note: I set it up this way rather than gating strictly on the `active` state so that users can still see the estimated cycle time when the machine is in the `Ready` state before pressing start, which was my main goal for adding this!)*
4. **Test & Fixture Updates:**
    * Updated the test input in `test_laundry_capabilities.py` to use uppercase `2B2A` to match the real firmware.
    * Removed `completion_time` from all tests and the 5 golden JSON fixtures.
    * Added tests for the new `completion_minutes` gating/fallback logic.
5. **Formatting:** Restored the missing trailing newlines in `operational.py` and `test_operational_capability.py` to keep the linter happy.

Everything should be passing cleanly now. Let me know if you need any other adjustments

mbillow (approved):

mbillow:
Would you mind rebasing and implementing the new translation/i18n interface, please? Sorry to put another hurdle between you and getting this merged, but we did some fairly large refactoring today to support better internationalization.

Metal-Eagle:
No problem! I just did a rebase and also added the Dutch translations for the new washing machine entities.

custom_components/localthings/strings.json
```diff
@@ -0,0 +1,268 @@
+{
```
mbillow: This entire file is no longer necessary. It was deleted from the repo.

custom_components/localthings/translations/en.json
```diff
@@ -25,823 +25,856 @@
...
+          "name": "{instance_name} setpoint"
+        },
+        "oven_setpoint": {
+          "name": "Setpoint"
+        },
+        "sound_volume": {
+          "name": "Sound volume"
         }
       },
-      "cooktop_state": {
```
mbillow, line 449: Not sure if this just got reordered a bunch or what, but this diff is way too complicated. It should only include the specific things for your change.

custom_components/localthings/strings.json
Metal-Eagle: My bad, I noticed it was just updated!

custom_components/localthings/translations/en.json
Metal-Eagle: It was using the old strings.json file, but I think I've fixed it now.

Metal-Eagle:
Now the tests are failing, so I'm trying to fix them.

Metal-Eagle:
I believe it's all working now. Thanks for your patience and help.

PR #68, opened 2026-07-24
Add full translation support

hmmbob:
## Summary

- add complete Dutch translations for LocalThings
- route static entity names through Home Assistant translation keys instead of hard-coded entity names
- add translated names for dynamic appliance instances such as compartments, ice makers, doors, and burners through placeholders
- translate known select options, enum states, climate modes, presets, config and options flows, repair text, and user-facing validation errors
- reuse Home Assistant common strings and device-class translations where available
- preserve exact Samsung values for writes and keep safe readable/raw fallbacks for unknown future vendor values
- ship resolved English and Dutch runtime catalogs while retaining Home Assistant common-string references in `strings.json`

## Why

The existing Dutch catalog translated only strings that were already exposed by the integration. Many entity descriptions still assigned an English name directly, dynamic instances assembled English names at runtime, and several vendor states and validation errors bypassed Home Assistant's translation mechanism.

This change makes the integration's static UI surface translatable through the normal Home Assistant backend translation architecture while leaving genuinely dynamic device-provided names and unknown vendor codes intact.

Home Assistant resolves `[%key:…%]` references while building Core's translation catalogs, but it loads a custom integration's shipped `translations/*.json` files directly. The runtime catalogs therefore contain the resolved text; a regression test rejects unresolved references so they can no longer leak into the UI.

## Compatibility

- unique IDs and internal state keys are unchanged
- existing entity IDs are not migrated or renamed
- device discovery, protocol communication, command payloads, and write paths are unchanged
- unknown or device-provided runtime values remain visible and round-trip to the appliance without lossy normalization
- unknown washer/dryer course tables use a generic translated entity name and raw course codes instead of borrowing labels from another hardware generation

## Validation

- Pytest: 457 passed
- Hassfest: passed
- translation catalog topology, placeholder parity, and resolved-runtime-reference checks: passed
- HACS integration checks: 7/9 passed; the remaining two checks concern fork-level repository settings (Issues disabled and missing repository topics), not integration contents

Latest validation run: https://github.com/hmmbob/localthings/actions/runs/30119969943


mbillow:
Awesome, thanks! Err… Dankjewel! 😉

You mentioned strings you can’t yet translate. I’m happy to rip them out into translation keys if there are any on your mind! You’re also more than welcome to do that in this PR as well. Let me know.

hmmbob:
> Awesome, thanks! Err… Dankjewel! 😉
> 
> You mentioned strings you can’t yet translate. I’m happy to rip them out into translation keys if there are any on your mind! You’re also more than welcome to do that in this PR as well. Let me know.

Codex is already churning!

mbillow:
I don’t _love_ some of the choices it made here. They are going to make maintenance a real chore. Let me see if I can figure out a better way to make this possible.

mbillow:
I pulled your changes here into a branch and iterated from there. I think we landed in a much cleaner and less duplicative spot. I'm going to close this, but your changes will be merged in #72.

Thanks again for everything here, really appreciate the effort. This work should make future language additions really simple (just json changes).

PR #70, opened 2026-07-24
feat: laundry firmware-flag gating, fridge vendor temp writes, course…

vmvarga:
## Summary
Adds firmware-flag-aware behavior for Samsung WD7000B-class washer-dryers and fixes temperature writes for TP1X_REF_21K-class refrigerators (RB38C705CWW/EF). All changes are generic and flag-based — they activate only when the device firmware reports the relevant fields, so existing devices are unaffected.
### Laundry (WD7000B / WD90DG5B15BEEC)
- **Power switch gating on `isModelSettingPowerOnOff`:** When `/wm/setinfo/vs/0` reports this flag as `"false"`, the writable power switch is replaced with a read-only binary sensor. The device ACKs power writes but ignores them — exposing a non-functional switch is misleading.
- **Auto-bypass remote control gate for settings writes:** When `isModelSettingWithoutSC` is `"true"`, the remote control validation is skipped for settings writes (wash temp, spin, course options, etc.) but still enforced for operational commands (`/operational/state` — start/pause/stop). Confirmed: these devices accept settings writes with Smart Control off.
- **Alarm filter:** `_active_alarm_codes` now filters out items with `state == "Deleted"` instead of showing stale alarm codes.
- **Table_02 course translations:** Added 4 codes confirmed on a WD90DG5B15BEEC: `34`=Mixed, `3a`=Drum Clean, `7f`=Wool/Delicate, `87`=Download.
### Fridge (TP1X_REF_21K / RB38C705CWW/EF)
- **Vendor-path temperature writes:** Samsung fridges expose both OCF-standard `/temperature/desired/*` and vendor `/temperatures/vs/0`. The OCF resource ACKs POSTs but silently ignores them — only the vendor path commits the setpoint change. New `TEMP_SETPOINT_VENDOR` pattern capability writes to `/temperatures/vs/0` when present; falls back to direct OCF writes when it's absent.
## Devices tested
| Device | Model | Board | Detection |
|--------|-------|-------|-----------|
| Samsung WD90DG5B15BEEC | WD7000B | DA_WM_TP2 | `for_device_by_model` (WD prefix → washer) |
| Samsung RB38C705CWW/EF | TP1X_REF_21K | TP1X | `for_device_by_model` (_REF_ → refrigerator) |
## Test plan
- [x] Probe script validates power gating, SC bypass, and alarm filtering logic
- [x] Unit tests for `remote_control_required_for_write` in coordinator
- [x] Unit tests for `model_allows_power_on_off` / `model_setting_without_sc` in common capabilities
- [x] Fridge temperature write confirmed via debug panel (CoAP 2.04, desired value changed)
- [x] Washer settings writes confirmed working without Smart Control enabled
- [x] Course name translations verified on device

### DIFF
--- a/custom_components/localthings/registry/by_type/refrigerator.py
+++ b/custom_components/localthings/registry/by_type/refrigerator.py
@@ -33,8 +33,7 @@ REGISTRY = DeviceRegistry(
     ]),
     pattern_capabilities=[
         fridge.TEMP_CURRENT_GENERIC,
-        fridge.TEMP_SETPOINT_VENDOR,
-        fridge.TEMP_SETPOINT_GENERIC,
+        fridge.TEMP_SETPOINT,
         fridge.ICEMAKER_GENERIC,
         fridge.DOOR_GENERIC,
     ],
--- a/custom_components/localthings/registry/by_type/refrigerator.py
+++ b/custom_components/localthings/registry/by_type/refrigerator.py
@@ -33,6 +33,7 @@ REGISTRY = DeviceRegistry(
     ]),
     pattern_capabilities=[
         fridge.TEMP_CURRENT_GENERIC,
+        fridge.TEMP_SETPOINT_VENDOR,
         fridge.TEMP_SETPOINT_GENERIC,
         fridge.ICEMAKER_GENERIC,
         fridge.DOOR_GENERIC,
--- a/custom_components/localthings/translations/en.json
+++ b/custom_components/localthings/translations/en.json
@@ -55,6 +55,9 @@
       "power_state": {
         "name": "Power state"
       },
+      "power_switch": {
+        "name": "Power"
+      },
       "rapid_freezing": {
         "name": "Rapid freezing"
       },
@@ -379,16 +382,20 @@
           "27": "Rinse+Spin",
           "28": "Drain/Spin",
           "29": "Drum Clean+",
+          "2a": "Jeans",
+          "2b": "AI Wash",
           "2d": "Silent Wash",
           "2e": "Baby Care",
           "2f": "Activewear",
           "30": "Cloudy Day",
           "32": "Shirts",
           "33": "Bedding",
+          "34": "Mixed",
           "36": "Wash+Dry",
           "37": "Air Wash",
           "38": "Cotton Dry",
           "39": "Synthetics Dry",
+          "3a": "Drum Clean",
           "53": "Heavy Duty",
           "55": "Activewear",
           "57": "Delicate",
@@ -398,11 +405,11 @@
           "7c": "Whites",
           "7d": "Bedding/Waterproof",
           "7e": "Self-Clean",
+          "7f": "Wool/Delicate",
           "86": "Deep Wash",
+          "87": "Download",
           "8f": "Intense Cold",
-          "96": "Less Microfiber",
-          "2b": "AI Wash",
-          "2a": "Jeans"
+          "96": "Less Microfiber"
         }
       },
       "washer_dry_level": {
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -564,7 +564,10 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
                     translation_domain=DOMAIN,
                     translation_key=error,
                 )
-        result = write_fn(payload, rep, href)
+        try:
+            result = write_fn(payload, rep, href, resources)
+        except TypeError:
+            result = write_fn(payload, rep, href)
         if result is None:
             self._log.warning("write_fn rejected payload %r for %s", payload, href)
             return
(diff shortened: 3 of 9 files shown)


mbillow:
The `isModelSettingWithoutSC` find is really interesting! Nice one. 

I’ll take a deeper look at this later. Can you do me a favor and amend these commits to remove the coauthor. I’m working on some AGENT.md and CONTRIBUTING.md additions to solidify this, but I’d love to keep agentic coauthors and committers out of the log.

splitsec2:
Independent confirmation of the fridge temperature fix. I applied just the fridge portion of this PR (`TEMP_SETPOINT_VENDOR` + `_temp_setpoint_vendor_write` and its registration) on top of v0.11.1. My fridge reports the same `TP1X_REF_21K` platform token as yours but is a different SKU (RF29DB…), so this is a second model in the class.

**Before:** setpoint writes went to the OCF `/temperature/desired/` path, which the fridge ACKs but silently ignores — the number entity snapped back on the next poll and the temperature never changed.

**After (vendor `/temperatures/vs/0` write):** both setpoints now commit on the device, not just optimistically in HA. Verified each by setting it, then confirming it held through a device re-poll rather than reverting:
- cooler 3 °C → 2 °C — held
- freezer −18 °C → −19 °C — held

Haven't exercised the laundry / `isModelSetting*` changes — just the fridge temperature path. 👍

mbillow:
> setpoint writes went to the OCF /temperature/desired/ path, which the fridge ACKs but silently ignores — the number entity snapped back on the next poll and the temperature never changed.

Interesting, I've verified that the temperature changes previously worked on my fridge, so this is definitely an inconsistency between firmwares / models. I'll have to make sure this still works on my fridge before merging. 

@vmvarga Would you mind rebasing this? It has been a busy day and we did a pretty robust overhaul of the i18n platform / patterns of the component.

vmvarga:
@mbillow Removed coauthor and rebased

custom_components/localthings/translations/en.json
```diff
@@ -402,7 +405,11 @@
           "8f": "Intense Cold",
           "96": "Less Microfiber",
           "2b": "AI Wash",
-          "2a": "Jeans"
+          "2a": "Jeans",
+          "34": "Mixed",
+          "3a": "Drum Clean",
+          "7f": "Wool/Delicate",
+          "87": "Download"
```
mbillow: Trying to keep these sorted. Can you order them accordingly?

probe/test_wm_setinfo_flags.py
```diff
@@ -0,0 +1,135 @@
+"""Probe: validate setinfo-driven power/SC gating without Home Assistant.
```
mbillow: Did you mean to include this file?

custom_components/localthings/translations/en.json
vmvarga: fixed

custom_components/localthings/registry/by_type/refrigerator.py
```diff
@@ -33,6 +33,7 @@
     ]),
     pattern_capabilities=[
         fridge.TEMP_CURRENT_GENERIC,
+        fridge.TEMP_SETPOINT_VENDOR,
```
mbillow: This would create completely duplicate set point entities, which would be confusing. I verified my fridge allows writes to both the /temp/desired and /temps/vs endpoints. Let’s just update the current capability to use your write function and get rid of this duplication.

splitsec2:
+1 on collapsing to a single capability — the duplicate setpoint entities are confusing, and right now the "vendor wins" behavior only happens because it's registered first and wins a `unique_id` collision, which is fragile.

One thing worth keeping from the two-capability version, though: the `match_fn` guard on `/temperatures/vs/0`. If the generic write_fn is repointed at the vendor function *unconditionally*, any model that has a working `/temperature/desired/` write but doesn't expose `/temperatures/vs/0` would start POSTing to a resource that isn't there. We've only got two data points so far — @mbillow's fridge (writes to both) and mine (only the vendor endpoint commits; generic is silently ignored) — so neither of us has hit a generic-only model, but it's a plausible third case.

Cleanest way to get both: keep the single capability, but move the vendor-vs-generic choice *into* the write_fn — use the `/temperatures/vs/0` payload when that resource is present, else fall back to the current `/temperature/desired/` write. No duplication, and the guard survives.

vmvarga:
fridge.py — Collapsed TEMP_SETPOINT_VENDOR + TEMP_SETPOINT_GENERIC into a single TEMP_SETPOINT capability with a smart _temp_setpoint_write function that:

Checks if /temperatures/vs/0 exists in resources at write time
If yes → uses the vendor path (Samsung convention: item "0" = Freezer, "1" = Cooler)
If no → falls back to the generic OCF direct-write path
resources=None default ensures backwards-compatibility
coordinator.py — Added backwards-compatible try/except TypeError around the write_fn call to pass resources as a 4th argument. Old-style 3-arg write_fns continue to work unmodified (the TypeError triggers the fallback to the 3-arg call).

refrigerator.py — Replaced fridge.TEMP_SETPOINT_VENDOR + fridge.TEMP_SETPOINT_GENERIC with just fridge.TEMP_SETPOINT.

test_fridge_capabilities.py — Renamed TestTempSetpointGeneric to TestTempSetpoint and updated the reference from TEMP_SETPOINT_GENERIC to TEMP_SETPOINT.

PR #71, opened 2026-07-24
fix(fridge): switch turn-off silently sends 'On' (truthy-string guard)

splitsec2:
## The bug

Six fridge `SwitchDesc` write_fns build their OCF payload like this:

```python
{'ice.night.status': 'On' if p else 'Off'}
```

The switch platform hands `write_fn` the **literal string** `'Off'` on
turn-off (and `'On'` on turn-on). `'Off'` is a non-empty string, so it's
truthy — `'On' if p else 'Off'` always evaluates to `'On'`. The upshot is
that turning any of these switches **off** silently re-sends `On`, and they
can never actually be turned off from the UI.

Affected descriptors (all in `registry/capabilities/fridge.py`):

- `ICEMAKER_NIGHTTIME` → `ice.night.status`
- `STATUS_LOCK` (the `_status_lock_write` helper) → `x.com.samsung.da.ado.devicecontrol` and `x.com.samsung.da.device.sound`
- `DEFROST_DELAY` → `x.com.samsung.da.delayDefrost`
- `WELCOME_LIGHTING` → `status`
- `CABINET_LIGHT` dim → `light.dimming.status`
- `ICEMAKER_STATUS_FALLBACK` → `x.com.samsung.da.iceMaker`

## The fix

Compare `p == 'On'` explicitly, which is the pattern the other capability
files (and the pass-through helpers like `_cabinet_light_write`) already
use. No behaviour change on turn-on; turn-off now correctly sends `Off`.

## Test

Adds a `TestSwitchOffIsNotInverted` regression test that drives every
affected `write_fn` with both `'On'` and `'Off'` and asserts the payload
matches. Verified it **fails** against the current truthy-string code and
**passes** with this fix (full suite: 25 passed).

### DIFF
--- a/custom_components/localthings/registry/capabilities/fridge.py
+++ b/custom_components/localthings/registry/capabilities/fridge.py
@@ -92,7 +92,7 @@ ICEMAKER_NIGHTTIME = Capability(
                    value_fn=lambda v: v == 'On',
                    write_fn=lambda p, rep, href=None: (
                        ['icemaker', 'nighttime', 'vs', '0'],
-                       {'ice.night.status': 'On' if p else 'Off'})),
+                       {'ice.night.status': 'On' if p == 'On' else 'Off'})),
     ),
 )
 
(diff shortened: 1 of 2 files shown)


mbillow:
Great catch! Thanks for the contribution!

Issue #56, opened 2026-07-23
Samsung AirPurifier


ericksoegiantoro-a11y:
### Appliance model

AX60R5080WD/SE

### What's missing or wrong

Hi! First of all, thanks for all the work on LocalThings. I recently tried connecting my Samsung air purifier (AX60R5080WD/SE, internal model ARTIK051_TVTL_18K) and was pleasantly surprised that it already communicates over the local protocol.
I noticed that Samsung air purifiers don't seem to be supported yet. If support is ever something you'd be interested in adding, I'd be happy to help test it.
From looking at the device, it appears to expose resources for things like:
Power
Fan mode (Auto, Sleep, 1, 2, 3)
PM1.0
PM2.5
PM10
Odor
It also seems to expose filter status, alarms, and a few other resources, but I think the list above would already make a great starting point.
I've attached a diagnostics file, and if it would help I can also provide other logs if needed.

### Diagnostics file

[config_entry-localthings-01KY7TR3DWWQC55T9YEF272YDN.json](https://github.com/user-attachments/files/30314667/config_entry-localthings-01KY7TR3DWWQC55T9YEF272YDN.json)

Nylizz:
Heyo, 
Also have the similar (identical?) model, so here's my diagnostics as well:

[config_entry-localthings-01KY86V47DQ2PVHRB0361QWJEE.json](https://github.com/user-attachments/files/30320138/config_entry-localthings-01KY86V47DQ2PVHRB0361QWJEE.json)

mbillow:
Thanks both for the diagnostics dumps — really helpful having two independent captures of the same internal model (`ARTIK051_TVTL_18K`, AX60R5080WD/SE). I've pushed initial air purifier support on branch `claude/device-support-issue-56-shy2vg`.

**What's now supported:**
- Power on/off
- Alarms (filter alarm, error codes)
- Air quality sensors: dust, fine dust, super fine dust, odor, clean level
- Filter status ("filter progress" %)
- Diagnosis status
- Display light on/off (parsed out of `/mode/vs/0`)
- Energy consumption (if your unit reports it)
- Kids lock / remote control (if your unit reports them)

Both dumps now resolve to this new registry with zero "unbound" resources, so the coverage-gap repair should clear once this lands.

**What's still incomplete, and where I need your help:**

[...]

PR #85, opened 2026-07-26
Add Samsung AVT-WW-TP1 air purifier support (fixes #84)

### LINKED ISSUE #84: Samsung air purifier AVT-WW-TP1-23-AXX500 detected as `unknown` — full capability map + tested implementation
## Summary

The Samsung air purifier **AVT-WW-TP1-23-AXX500** is not recognised by
`for_device_by_model`: it lands on `device_type: unknown`, raises the "incomplete
capability coverage" repair, and leaves **12 hrefs unbound** — no fan, no filter, no
air-quality sensors, none of the AI Purify controls.

Root cause is a single missing model-routing rule. This board is an air purifier that
happens to share almost nothing structurally with the existing `ARTIK051_TVTL` purifier
family (issue #56), so once routed it also needs its own capability set.

I've implemented the routing rule plus the full capability map and verified it against
the physical unit. Patch is on a branch and I'm happy to open a PR — details below.

## Environment

- LocalThings: `main` (`aff7647`), Home Assistant 2026.7.4
- Device: `AVT-WW-TP1-23-AXX500`, firmware `21102800,21102100`, SW `02530A260327`
- `x.com.samsung.da.airconOptionList`: `["WELCOMECARE", "AI_PURIFY"]`
- Attached: the scrubbed `/device/0` dump in the repo's **test-fixture** shape
  (`{"device0": [{href, rep}, ...]}`), not the diagnostics-download wrapper — it's the
  file I added as `tests/fixtures/air_purifier_avt_device.json`, so it drops straight
  into the harness. Serials/MACs/SSID are redacted and the region code is zeroed.

## Bug 1 — `for_device_by_model` has no rule for the `AVT-` board family

`modelNum` is `AVT-WW-TP1-23-AXX500|10238341|7000023C…`, `description` is
`AVT-WW-TP1-23-AXX500`. It reports **no `oneUiVersion`** and carries **no `_TVTL_`
token**, so none of the existing branches match and `key` stays `None`.

Fix mirrors the existing `AHD-` range-hood rule — a `startswith` on the board-family
prefix:

```python
if key is None and model_identity.startswith('AVT-'):
    key = 'air_purifier'
```

`startswith` (rather than a substring scan) matters here: the `-WW-` colour/variant code
further into the string must not be confused with the washer consumer prefix, which is
read from `description`'s first token.

## Bug 2 — the AVT board needs its own capability set

Routing alone isn't enough. Compared to the `ARTIK051_TVTL` family already in
`capabilities/air_purifier.py`, this board:

| Function | TVTL family | AVT-WW-TP1 |
|---|---|---|
| Fan speed | `/airflow` (unmodelled) | `/wind/strength/vs/0` |
| Filter | `/consumable/vs/0` | `/filter/hepafilter/vs/0` |
| AI Purify | — | `/airlevelcheck/vs/0` |
| Power | `/power/*` switch | fan owns on/off |

Both sub-families coexist in one registry: each capability binds only when its href is
actually present, and the power switch gains a `match_fn` so it stays on fan-less TVTL
boards and drops on AVT (where the fan entity owns on/off). The TVTL golden fixture is
unchanged.

### The 12 unbound hrefs, and how they're now covered

| href | Treatment |
|---|---|
| `/wind/strength/vs/0` | Fan entity — `preset_modes` auto/low/medium/high/sleep, from `supportedModes` + `modesName` |
| `/filter/hepafilter/vs/0` | Filter usage %, status, usage (h), life remaining (h), capacity, reset button |
| `/airlevelcheck/vs/0` | AI Purify — sensing switch, mode select, interval, do-not-disturb window, diagnostics |
| `/sensors/vs/0` | PM10 / PM2.5 / PM1.0 / Odor / Air quality |
| `/devicespecificinfo/vs/0` | `device_active` binary sensor |
| `/option/muteonce/vs/0` | Mute-once switch (reuses `airconditioner.MUTE_ONCE` as-is) |
| `/mode/convenient/vs/0` | ignored — `supportedModes` is `["Off"]` only, no choice to offer |
| `/availablecontrolsets/vs/0` | ignored — opaque hex control-set bitmap |
| `/da/softreset/vs/0` | ignored — soft-reset plumbing |
| `/dnd/autosleep/vs/0` | ignored — `visible: "false"` on this unit, so the feature isn't available |
| `/humidity/0`, `/humidity/vs/0` | ignored — both empty `{}`, same treatment as the TVTL family already uses |

Two more hrefs weren't unbound but changed treatment:

- `/wirelessinfo/vs/0` — globally ignored, but this family exposes a usable
  `connectedApSsid`, so the global ignore is filtered out for the `air_purifier` registry
  and a Wi-Fi network sensor binds instead.
- `/mode/vs/0` — the existing `Light_*` switch already bound; this board additionally
  carries a `Pollution_*` flag, gated on with `exists_fn` so TVTL is unaffected.

Result: `unbound_hrefs: []`, ~30 entities.

### `/sensors/vs/0` particulates

Each `items[]` entry carries a two-element `value` — `["11","1"]` etc. Index 0 is a
numeric concentration, index 1 a coarse level. Live readings are monotonic across the
three particle sizes (`Dust 11 ≥ FineDust 9 ≥ SuperFineDust 5`), consistent with µg/m³,
so they're surfaced with HA's `pm10` / `pm25` / `pm1` device classes. `Odor` and
`CleanLevel` are unitless indices.

### `/airlevelcheck/vs/0` — AI Purify

This is the interesting one and has no equivalent in the existing purifier family. It
runs **while the appliance is powered off** — `periodicSensingActivationState: "On"` with
`/power` reading Off, and `lastSensingTime` still advancing — which is worth knowing when
deciding what the power entity should mean on this board.

| Field | Exposed as |
|---|---|
| `periodicSensingActivationState` | switch — AI Purify (Sensing), the master on/off |
| `+ autoExeState` | select — folded into one control: `off` / `sensing_only` / `auto_purify` / `st_alarm` |
| `periodicSensingInterval` | number, minutes in the UI ↔ seconds on the wire |
| `periodicSensingSkipStatus` | switch — do-not-disturb |
| `periodicSensingSkipTime` | `"HHMMHHMM"` split into two `time` entities; each write reads the other half back out of the live rep |
| `sensingState` / `lastSensingTime` / `lastSensingLevel` | diagnostic sensors |

`autoExeState` is asymmetric: it accepts `Airpurify`/`Alarm` only while
`periodicSensingActivationState` is already `On`, and rejects them while it's `Off`. A
standalone auto-action select is therefore unusable from a cold start, which is why the
two fields are folded into one select that writes both in a single PUT.

## Side finding — `alarm_code` shows `ErrorCode_OFF` instead of "no alarm"

Not AVT-specific, so not fixed in my patch. `common.ALARMS` surfaces the raw code
without filtering, so a healthy device reads:

```json
{"x.com.samsung.da.code": "ErrorCode_OFF", "x.com.samsung.da.state": "Deleted"}
```

as an active alarm code. `range_hood` already handles this with `_active_alarm_codes`;
the shared capability could filter on `state == "Deleted"` and/or
`code == "ErrorCode_OFF"` the same way.

## Verification

Registry/translation tests pass (144), including a new golden fixture for this board and
a no-regression check on the TVTL golden. Write contracts confirmed against the physical
unit:

| Control | Wire contract | Result |
|---|---|---|
| Fan preset | `{"x.com.samsung.da.modes": "<code>"}` scalar (not a list); all five codes incl. Sleep `91` | ✅ |
| Fan power | `/power/0` `{"value": bool}` — the preset write powers the unit on as a side effect, which is why the fan owns on/off here | ✅ |
| Pollution indicator | single-token `option_write` on `/mode/vs/0` — `Light_*` and the opaque `OptionCode_27514` survive the write untouched | ✅ |
| Sensing interval | scalar seconds; arbitrary values honoured (60 s drove sensing every ~60 s, not just the app's 10/30/60 min) | ✅ |
| AI Purify (Sensing) switch | `periodicSensingActivationState` On/Off | ✅ |
| AI Purify (Mode) select | one PUT setting both fields. Tested from a cold `off`: `off → auto_purify` lands `{On, Airpurify}`, `→ st_alarm` lands `{On, Alarm}` | ✅ |
| Do-not-disturb switch | `periodicSensingSkipStatus` On/Off | ✅ |
| Do-not-disturb window | each half writes independently and preserves the other: start → `07452300`, then end → `07452200` | ✅ |
| Reset HEPA filter button | `{"x.com.samsung.da.filterResetType": ["replaceable"]}` | ⚠️ see below |

Every ✅ was read back from the resource's raw `rep` after the write's settle window
expired, and the key cases were additionally re-checked after a full config-entry reload
so the value came from a freshly-swept device read rather than the optimistic cache.

Corroborating behaviour: with the skip window active and covering the current time,
`lastSensingTime` stopped advancing despite `periodicSensingInterval` being 600 s — the
device really is honouring the window we wrote.

### Open question on the filter reset

This is the one control I can't call confirmed.

`filterResetType` **reads** as an array (`["replaceable"]`). The button first wrote the
scalar `"replaceable"`, by analogy with the other reset-style buttons in the codebase
(all of which target scalar fields). On hardware that **silently no-ops**: the device
accepts the PUT, nothing is logged, and `filterUsage` stays put.

It now writes the array form `["replaceable"]` instead — matching the shape the device
reports, and consistent with `laundry.option_write`, which already establishes that a
field reading as a list (`x.com.samsung.da.options`) is written back as a list. The
scalar was the odd one out. Re-test pending.

If the array form also no-ops, the remaining candidates are a direct
`{"x.com.samsung.da.filterUsage": "0"}`, or a write-only command field that doesn't
appear in the read rep at all — SmartThings models this as `samsungce.airPurifierFilter`
with a `resetFilter()` command distinct from the `resetType` attribute, which hints the
OCF side may have a separate trigger. **If anyone knows the actual reset contract for
this resource, I'd rather be told than keep probing** — every failed guess that turns
out to be right zeroes a real filter counter irreversibly.

## Not modelled

- `startSensingOnce` on `/airlevelcheck/vs/0` — a one-shot "sense now" trigger. The href
  is bound, so it doesn't show as unbound coverage, but there's no entity for it yet.
- `sensingState` reads `NonProcessing` / (presumably) `Processing` with no state labels
  in the catalog, so it renders raw. It's a disabled-by-default diagnostic, so low
  priority, but worth labelling if the value set gets confirmed.
- `CleanLevel` is currently surfaced with unit `CAQI`. On this unit it looks more like a
  1–4 national grade than a 0–100 index — `lastSensingLevel` reads `Kr1` alongside
  `CleanLevel: 1` at PM10 11 µg/m³ — so the unit is probably wrong and an `enum` with
  labelled grades would be more honest. Flagging rather than changing, since it would
  break existing statistics.

## PR?

Branch is ready against `main` (registry rule, capabilities, fan dispatch, a shared
`extra_attributes` field on `SamsungEntityDescription`, en/nl translations, tests +
fixtures). Happy to open it if you'd like it in this shape — or split the routing fix out
first if you'd rather take that separately.

## Scrubbed `/device/0` dump

<details>
<summary><code>tests/fixtures/air_purifier_avt_device.json</code> (serial / MACs / SSID redacted, region code zeroed)</summary>

```json
{
  "device0": [
    {
      "rt": [
        "x.com.samsung.devcol"
      ],
      "if": [
        "oic.if.baseline",
        "oic.if.ll",
        "oic.if.b"
      ]
    },
    {
      "href": "/realtimenotiforclient/vs/0",
      "rep": {
        "x.com.samsung.da.timeforshortnoti": "0",
        "x.com.samsung.da.longnotisubscription": "true",
        "x.com.samsung.da.periodicnotisubscription": "true"
      }
    },
    {
      "href": "/airlevelcheck/vs/0",
      "rep": {
        "x.com.samsung.da.periodicSensingActivationState": "On",
        "x.com.samsung.da.periodicSensingInterval": "600",
        "x.com.samsung.da.startSensingOnce": "On",
        "x.com.samsung.da.sensingState": "NonProcessing",
        "x.com.samsung.da.lastSensingTime": "1784978700",
        "x.com.samsung.da.lastSensingLevel": "Kr1",
        "x.com.samsung.da.autoExeState": "Off",
        "x.com.samsung.da.supportedAutoExeState": [
          "Off",
          "Airpurify",
          "Alarm"
        ],
        "x.com.samsung.da.periodicSensingSkipStatus": "Off",
        "x.com.samsung.da.periodicSensingSkipTime": "00000000"
      }
    },
    {
      "href": "/mode/convenient/vs/0",
      "rep": {
        "x.com.samsung.da.supportedModes": [
          "Off"
        ]
      }
    },
    {
      "href": "/wind/strength/vs/0",
      "rep": {
        "x.com.samsung.da.modes": "0",
        "x.com.samsung.da.supportedModes": [
          "0",
          "2",
          "3",
          "4",
          "91"
        ],
        "x.com.samsung.da.modesName": [
          "Auto",
          "Mid",
          "High",
          "Turbo",
          "Sleep"
        ]
      }
    },
    {
      "href": "/filter/hepafilter/vs/0",
      "rep": {
        "x.com.samsung.da.filterUsage": "8",
        "x.com.samsung.da.filterUsageResolution": "1",
        "x.com.samsung.da.filterStatus": "normal",
        "x.com.samsung.da.filterCapacity": "8762",
        "x.com.samsung.da.filterCapacityUnit": "Hour",
        "x.com.samsung.da.filterResetType": [
          "replaceable"
        ]
      }
    },
    {
      "href": "/dnd/autosleep/vs/0",
      "rep": {
        "x.com.samsung.da.startTime": "00:00:00",
        "x.com.samsung.da.endTime": "00:00:00",
        "x.com.samsung.da.visible": "false",
        "x.com.samsung.da.useTimeSetting": "false",
        "x.com.samsung.da.functionState": "false"
      }
    },
    {
      "href": "/availablecontrolsets/vs/0",
      "rep": {
        "x.com.samsung.da.sets": "021D0100000A0000000000000000",
        "x.com.samsung.da.id": "VTL",
        "x.com.samsung.da.version": "1.0"
      }
    },
    {
      "href": "/da/softreset/vs/0",
      "rep": {
        "x.com.samsung.da.softwarereset": "false"
      }
    },
    {
      "href": "/alarms/vs/0",
      "rep": {
        "x.com.samsung.da.items": [
          {
            "x.com.samsung.da.id": "0",
            "x.com.samsung.da.description": "Alarm",
            "x.com.samsung.da.alarmType": "Device",
            "x.com.samsung.da.code": "ErrorCode_OFF",
            "x.com.samsung.da.triggeredTime": "2026-07-25T11:29:22",
            "x.com.samsung.da.state": "Deleted"
          }
        ]
      }
    },
    {
      "href": "/energy/consumption/vs/0",
      "rep": {
        "x.com.samsung.da.cumulativePower": "31500",
        "x.com.samsung.da.cumulativeSavedPower": "0",
        "x.com.samsung.da.cumulativeDate": "1785011348",
        "x.com.samsung.da.cumulativeDateUTC": "1784978948",
        "x.com.samsung.da.cumulativeUnit": "Wh"
      }
    },
    {
      "href": "/energy/consumption/0",
      "rep": {}
    },
    {
      "href": "/mode/vs/0",
      "rep": {
        "x.com.samsung.da.options": [
          "Light_On",
          "Pollution_Off",
          "OptionCode_27514"
        ]
      }
    },
    {
      "href": "/power/vs/0",
      "rep": {
        "x.com.samsung.da.power": "On"
      }
    },
    {
      "href": "/power/0",
      "rep": {
        "value": true
      }
    },
    {
      "href": "/sensors/vs/0",
      "rep": {
        "x.com.samsung.da.items": [
          {
            "x.com.samsung.da.id": "0",
            "x.com.samsung.da.description": "Sensor for Dust",
            "x.com.samsung.da.type": "Dust",
            "x.com.samsung.da.value": [
              "5",
              "1"
            ]
          },
          {
            "x.com.samsung.da.id": "1",
            "x.com.samsung.da.description": "Sensor for FineDust",
            "x.com.samsung.da.type": "FineDust",
            "x.com.samsung.da.value": [
              "5",
              "1"
            ]
          },
          {
            "x.com.samsung.da.id": "2",
            "x.com.samsung.da.description": "Sensor for Odor",
            "x.com.samsung.da.type": "Odor",
            "x.com.samsung.da.value": [
              "1"
            ]
          },
          {
            "x.com.samsung.da.id": "3",
            "x.com.samsung.da.description": "Sensor for CleanLevel",
            "x.com.samsung.da.type": "CleanLevel",
            "x.com.samsung.da.value": [
              "1"
            ]
          },
          {
            "x.com.samsung.da.id": "4",
            "x.com.samsung.da.description": "Sensor for SuperFineDust",
            "x.com.samsung.da.type": "SuperFineDust",
            "x.com.samsung.da.value": [
              "5",
              "1"
            ]
          }
        ],
        "x.com.samsung.da.total": "5"
      }
    },
    {
      "href": "/information/vs/0",
      "rep": {
        "x.com.samsung.da.modelNum": "AVT-WW-TP1-23-AXX500|10238341|7000023C001611CC0D02080200080000",
        "x.com.samsung.da.description": "AVT-WW-TP1-23-AXX500",
        "x.com.samsung.da.serialNum": "**REDACTED**",
        "x.com.samsung.da.otnDUID": "**REDACTED**",
        "x.com.samsung.da.diagProtocolType": "BLE_OCF",
        "x.com.samsung.da.diagLogType": [
          "errCode",
          "dump"
        ],
        "x.com.samsung.da.diagDumpType": "file",
        "x.com.samsung.da.diagEndPoint": "SSM",
        "x.com.samsung.da.diagMnid": "0AJT",
        "x.com.samsung.da.diagSetupid": "AP0",
        "x.com.samsung.da.diagMinVersion": "3.0",
        "x.com.samsung.da.diagTsId": "DA01",
        "x.com.samsung.da.items": [
          {
            "x.com.samsung.da.id": "0",
            "x.com.samsung.da.description": "Version",
            "x.com.samsung.da.type": "Software",
            "x.com.samsung.da.number": "02530A260327",
            "x.com.samsung.da.newVersionAvailable": "0"
          },
          {
            "x.com.samsung.da.id": "1",
            "x.com.samsung.da.description": "Version",
            "x.com.samsung.da.type": "Firmware",
            "x.com.samsung.da.number": "21102800,21102100",
            "x.com.samsung.da.newVersionAvailable": "0"
          },
          {
            "x.com.samsung.da.id": "2",
            "x.com.samsung.da.description": "Version",
            "x.com.samsung.da.type": "Firmware",
            "x.com.samsung.da.newVersionAvailable": "0"
          }
        ]
      }
    },
    {
      "href": "/file/information/vs/0",
      "rep": {
        "x.com.samsung.timeoffset": "+09:00"
      }
    },
    {
      "href": "/configuration/vs/0",
      "rep": {
        "x.com.samsung.da.region": "0000000000",
        "x.com.samsung.da.airconOptionList": [
          "WELCOMECARE",
          "AI_PURIFY"
        ],
        "x.com.samsung.da.countryCode": ""
      }
    },
    {
      "href": "/humidity/0",
      "rep": {}
    },
    {
      "href": "/humidity/vs/0",
      "rep": {}
    },
    {
      "href": "/devicespecificinfo/vs/0",
      "rep": {
        "x.com.samsung.da.deviceActive": true
      }
    },
    {
      "href": "/otninformation/vs/0",
      "rep": {
        "x.com.samsung.da.target": "",
        "x.com.samsung.da.newVersionAvailable": "false",
        "x.com.samsung.da.newVersionNo": "00000000",
        "x.com.samsung.da.currentVersionInfo": "00000000",
        "otnStatus": "None",
        "flashingProgress": "",
        "otnTarget": "main",
        "otnCompleteDate": "noHistory",
        "otnList": [
          {
            "type": "WIFI",
            "modelId": "AVT-WW-TP1-23-AXX500",
            "versions": [
              "12260327"
            ],
            "visVersion": "260327"
          },
          {
            "type": "Micom",
            "modelId": "",
            "versions": [
              "21102800",
              "21102100"
            ],
            "visVersion": "211028"
          },
          {
            "type": "Micom",
            "modelId": "",
            "versions": [
              "21102800",
              "21102100"
            ],
            "visVersion": "211028"
          }
        ]
      }
    },
    {
      "href": "/timezone/vs/0",
      "rep": {
        "timezoneid": "Asia/Seoul",
        "offset": "+09:00",
        "DST": "OFF"
      }
    },
    {
      "href": "/connectionconfig/vs/0",
      "rep": {
        "autoReconnectionMinVersion": "1.0",
        "autoReconnection": "true",
        "autoReconnectionProtocolType": [
          "helper_hotspot",
          "ble_ocf"
        ],
        "supportedWiFiAuthType": [
          "OPEN",
          "WEP",
          "WPA-PSK",
          "WPA2-PSK",
          "SAE"
        ],
        "supportedWiFiCryptoType": [
          "TKIP",
          "AES",
          "WEP-64",
          "WEP-128"
        ],
        "supportedWiFiFreq": [
          "2.4G"
        ],
        "calmConnectionCare": {
          "version": "1.0",
          "role": [
            "things"
          ]
        }
      }
    },
    {
      "href": "/option/muteonce/vs/0",
      "rep": {
        "muteonce": "Off"
      }
    },
    {
      "href": "/wirelessinfo/vs/0",
      "rep": {
        "macaddressWiFi": "**REDACTED**",
        "macaddressBLE": "**REDACTED**",
        "connectedApSsid": "**REDACTED**"
      }
    },
    {
      "href": "/quickcontrol/info/vs/0",
      "rep": {
        "supportedVersion": "1.0"
      }
    }
  ]
}
```

</details>

kkqq9320:
> Review feedback is in: the Wi-Fi SSID sensor and its carve-out from the global
> `/wirelessinfo` ignore are gone, and so is `extra_attributes` along with its one
> use — `entity.py` and `registry/entities.py` are no longer touched at all. Commits
> are rewritten with no co-author trailers; author and committer are me throughout.
>
> For context on the close/reopen: I closed this to rework it, and the rework caught
> that the shared `AIR_QUALITY` capability had been renaming and re-uniting entities
> on the ARTIK051_TVTL family, which I don't own. That's fixed — the section below
> spells out exactly what TVTL does and doesn't pick up.

Implements the AVT-WW-TP1 air purifier support described in #84 — that issue has the
resource map, the reasoning per href, and the hardware verification matrix. This PR is
the code.

## What it does

1. **Routes the board.** `for_device_by_model` gains an `AVT-` prefix rule, mirroring the
   existing `AHD-` range-hood rule. Without it this model resolves to `unknown` with 12
   unbound hrefs.

2. **Adds the AVT capability set** to the existing `air_purifier` registry, below a
   banner comment that separates the two board families. Every new capability binds only
   when its href is present, so a board picks up only what it exposes.

3. **One shared change:** `fan.py` dispatches on the bound href, so hood fans keep
   `LocalThingsRangeHoodFan` and `/wind/strength/vs/0` gets a preset-only
   `LocalThingsAirPurifierFan`. Nothing else outside the air-purifier registry moves —
   `entity.py` and `registry/entities.py` are untouched.

Result: `unbound_hrefs: []`, 28 entities, `device_type: air_purifier`.

## What this does to the existing ARTIK051_TVTL family

Worth its own section, because `air_purifier.py` is shared and I don't have TVTL
hardware. **No existing entity changes name, unit, or icon.** `en.json` is purely
additive against `main` — the diff removes no line from it.

The one behavioural change TVTL picks up is on the three particulate sensors, which gain
`device_class` (pm10 / pm25 / pm1), `µg/m³`, and `state_class=measurement`. That is
evidenced rather than guessed: index 0 of each `items[]` entry falls monotonically with
particle size on **both** families' dumps (TVTL 11/9/5, AVT 11/9/5), which is what
concentrations do and what an N-step grade would not. `odor` gains `state_class` only.
Splitting the capability per-board instead costs one `match_fn` if you prefer that.

The other thing that could have reached TVTL, checked and doesn't: `common.POWER` is
replaced with a `match_fn`-gated pair so the power switch drops on boards where the fan
owns on/off. TVTL has no `/wind/strength`, so it keeps its switch, and its golden fixture
is unchanged.

An earlier revision of this branch also renamed `clean_level` to "Air quality" with a
`CAQI` unit, and the dust sensors to "Dust PM10"-style names. Both are reverted: renaming
entities that have shipped since #56 isn't mine to do, and I don't believe the CAQI unit
anyway (see below).

**Heads-up on the golden fixtures:** they compare `state_keys` only, so none of the above
would have shown up there. A unit, `device_class`, or name change on a shared capability
passes golden regression untouched — I only caught my own by diffing. Widening the golden
format is out of scope here, but it's a real gap.

## Verification

`pytest tests/` — new `test_air_purifier_avt_capabilities.py` (16 tests) plus a golden
fixture for this board added to `test_golden_regression.py`. TVTL golden unchanged.

Write contracts confirmed against the physical unit — the full matrix is in #84. The
short version: fan presets (scalar `modes`, all five codes), pollution indicator
(single-token `option_write`, doesn't clobber the packed array), sensing interval, the
AI Purify sensing switch, the folded mode select (tested from a cold `off`, so the single
combined PUT is genuinely exercised), the do-not-disturb switch, and both halves of the
`HHMMHHMM` skip window. Each was read back from the raw `rep` after the settle window
expired, with the key cases re-checked after a config-entry reload.

## No filter-reset button — I couldn't work out the contract

There is no reset control in this PR. `/filter/hepafilter/vs/0` clearly supports a reset
somehow, but I couldn't find the write that triggers it, so I left it out rather than
ship a button that appears to work and doesn't.

What I tried on the hardware, and what each did:

| Write | Result |
|---|---|
| `filterResetType: "replaceable"` (scalar), appliance off | 2.04, `filterUsage` unchanged |
| `filterResetType: ["replaceable"]` (list), appliance off | 2.04, `filterUsage` unchanged |
| both of the above again with the appliance powered on | 2.04, `filterUsage` unchanged |
| `filterReset: "01"` — the field dishwashers carry on `/filter/waterfilter/vs/0` | 2.04, `filterUsage` unchanged |
| `filterLastResetDate: "<timestamp>"` — SmartThings' `custom.hepaFilter` has that attribute, this resource doesn't | 2.04, `filterUsage` unchanged |
| `filterUsage: "0"` directly | 2.04, `filterUsage` unchanged |
| `GET /filter/hepafilter/0` (OCF-standard variant) | 4.04 |

Every write is acknowledged and discarded, so the response code tells you nothing — only
the side effect does. Worth knowing if anyone else probes this: a 2.04 here is not
success.

For reference, the vendor's own `resetHepaFilter()` — invoked through SmartThings on the
same physical unit — **does** work, and moves exactly one field on this resource:
`filterUsage` 8 → 0, with no new key appearing. So a real reset is visible here; I just
don't know what triggers it, and it may not be a write to this resource at all.

I've recorded the excluded candidates in a module comment and added a test that keeps the
button from reappearing on a hunch. If you know the actual contract, I'll happily add it.

## Cross-check against HA core's own SmartThings integration

This unit is also paired to SmartThings, so the same physical appliance shows up twice in
my Home Assistant — once through this integration and once through core's `smartthings`.
That turned out to be a useful control, because two of the choices here match what core
already does for this exact model:

| core's `smartthings` entity | this PR's local entity |
|---|---|
| `switch.*_do_not_disturb` (off) | `switch.*_do_not_disturb` (off) |
| `time.*_do_not_disturb_start_time` 13:00 | `time.*_do_not_disturb_start` 13:00 |
| `time.*_do_not_disturb_end_time` 23:00 | `time.*_do_not_disturb_end` 23:00 |
| `fan.*` presets auto/low/medium/high/sleep, `supported_features: 56` | identical |

So "Do not disturb" for the sensing-skip window, and a preset-only fan labelled
low/medium/high rather than the device's raw Mid/High/Turbo, aren't my invention — core
reaches the same two answers for the same appliance. The matching 13:00/23:00 values also
confirm the DND entities are bound to the same underlying feature.

## Judgement calls

- **"AI Purify (Sensing)" / "AI Purify (Mode)"** are mine, and the parenthetical
  disambiguators don't fit HA's style guide especially well. These are new entities, so
  renaming them breaks nothing — pick whatever you prefer.

- **The shared `AIR_QUALITY` change reaches TVTL and I can't test that board.** Covered
  above; the evidence is that index 0 falls monotonically with particle size on both
  families' dumps, but it is still the one part of this PR touching hardware I don't own.

- **`CleanLevel` has no unit, unchanged from `main`.** Flagging rather than proposing: on
  this board it looks like a 1–4 national grade rather than a 0–100 index —
  `lastSensingLevel` reads `Kr1` alongside `CleanLevel: 1` at PM10 11 µg/m³. `CAQI` and
  `AQI` both look wrong on that reading, and an `enum` would need the grade set confirmed
  first.

Also noted in #84 but deliberately not touched here: `common.ALARMS` reports
`ErrorCode_OFF` as an active alarm on a healthy device (the item carries
`state: "Deleted"`). It's cross-family, so it didn't belong in a device-support PR.


kkqq9320:
Closing this for now — I want to rework how the air-purifier fan and the AVT capability set are structured before asking anyone to review it. The device report in #84 stands on its own; I'll open a fresh PR once the shape is right. Sorry for the noise.

custom_components/localthings/registry/by_type/air_purifier.py
```diff
@@ -1,25 +1,46 @@
...
+from ..capabilities import air_purifier, airconditioner, common, dishwasher, ignored
 from ._base import DeviceRegistry, _build
 
 REGISTRY = DeviceRegistry(
     name='air_purifier',
     capabilities=_build([
-        *ignored.IGNORED,
+        # Drop the global /wirelessinfo ignore -- this family exposes the SSID
+        # (air_purifier.WIRELESS below) instead.
+        *[c for c in ignored.IGNORED if c.href != '/wirelessinfo/vs/0'],
```
mbillow: Lots of devices expose which SSID they are connected to using the href. I'd really rather show it for all devices (move `WIRELESS` to `common.py` and remove the ignore) or not show it here. Honestly, I don't see a huge amount of value in the sensor, but I'm not too opinionated either way.

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -172,3 +242,313 @@ def _light_write(payload, rep, href=None):
...
+        NumberDesc(key='sensing_interval',
+                   field='x.com.samsung.da.periodicSensingInterval',
+                   icon='mdi:timer-cog', entity_category='config',
+                   native_min=1, native_max=60, step=1, unit='min',
+                   value_fn=_interval_minutes, write_fn=_interval_write),
+        SwitchDesc(key='periodic_air_sensing',
+                   field='x.com.samsung.da.periodicSensingActivationState',
+                   icon='mdi:radar', entity_category='config',
+                   extra_attributes={
+                       'description': 'Runs the selected AI Purify (Mode) '
```
mbillow: What is the goal here? Extra attributes aren't really visible in the UI. This is just a static string so I can't see it being useful in automations. It also won't get properly translated for users in different languages.

mbillow:
Also, please amend commits to remove agent author information. Committer and author name should be yourself, no coauthor extras.

mbillow (changes requested):

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: The goal was simply to avoid confusion. From the switch name alone, it isn't obvious that it continues to run even when the appliance is powered off, so I thought a short description might help.

That said, I agree with all three of your points. HA hides attributes behind an extra click, a static string isn't useful for automations or templates, and attribute values aren't part of the translation catalog. Since the SmartThings app already explains this behavior, the extra attribute doesn't really add any value.

custom_components/localthings/registry/by_type/air_purifier.py
kkqq9320: Dropped it. The sensor and the per-registry carve-out are both gone, and *ignored.IGNORED is back to its original behavior. You're right that exposing it for one device family while every other device keeps the global ignore isn't the right approach.

kkqq9320:
Closing this.

Upstream has since landed the routing for this board independently (74f34b9, issue #190), and #151/#130 brought the fan, HEPA filter and air-quality capabilities this PR was carrying — so almost all of its diff is now redundant. Re-probing the original dump against current main confirms it: the board resolves to the air_purifier registry with zero unbound hrefs out of the box.

What's actually left is the `/airlevelcheck/vs/0` cluster, which is still covered as scheduler plumbing even though two dumps of this board (#84 and #190) show it drives the feature the SmartThings app calls AI Purify. I've rewritten just that on top of current main as a single commit, with the writes verified on hardware. Opening it separately rather than force-pushing here, since the scope and title of this PR no longer match what's left.

Thanks for the review on the earlier shape — the SSID sensor and `extra_attributes` points both stuck, and neither is in the new work.

Issue #54, opened 2026-07-23
Washing machine without remote control


niker:
### Appliance model

SAMSUNG RS68A884CB1/EF

### What's missing or wrong

Hello,
I was wondering if you could clear up something.
The washing machine has a button that needs to be pressed before remote control like a custom program is possible, you are then basically forced to pick all the parameters and bypass the built-in programs.

But the SmartThings app can modify some of the settings even out of that mode - specifically the default amount of detergent and softener (high/medium/low/none) - that setting then applies to the official built-in programs even without remote control mode.

The problem is when I try to change those settings out of the remote control mode, the integration returns:

`Remote control is turned off on this device. Check your appliance's manual for how to enable remote control before Home Assistant can control it.`

I wonder if this is something hardcoded in the integration or if it's really trying to send the command and getting rejected on the appliance.
It is possible that this is only available over cloud control, I'm just curious if it's even possible to do locally. 

And thanks for the great integration, other than this bit my washing machine and fridge both work flawlessly :)

### Diagnostics file

[config_entry-localthings-01KY79N1VFBPGD5S6YAMF9MR0Z.json](https://github.com/user-attachments/files/30307677/config_entry-localthings-01KY79N1VFBPGD5S6YAMF9MR0Z.json)

mbillow:
This is what I was worried about when releasing v0.10.0. I couldn’t find anything on my dishwasher that would change without remote control on, but I knew Samsung would never make it that easy and make this consistent. 🫠

The answer is it’s us artificially blocking the request because I thought it would be better to display a user friendly error instead of the device just rejecting most (I assumed all) writes. 

I’ll see if something better is possible but would you prefer no friendly error at all (writes protected by remote control lock get silently dropped) or by default we block all writes with an override in the device options to manually disable that check for the entire device / or maybe even specific entities?

niker:
First, I have no idea if the command would succeed, it is possible that Samsung locked it to the cloud somehow, but I suppose there's only one way to know for sure 🙄 Hit me up with a modified version and I'll test it.

If there really is no ACK from the appliance I can only think of a hardcoded lookup table with entities known to work OR having some kind of override list in configuration.yaml that would allow users to "send anyway" for a specific entity name. Or both and slowly build the database for everyone.

mbillow:
Thanks for flagging this, @niker — and for offering to test it directly!

[...]

Issue #53, opened 2026-07-23
Windfree Air Conditioner - windfree mode missing


pejannl:
### Appliance model

Windfree Airco - TP1X_DA-AC-RAC-01001_0000 (AR70F09D1DWNST)

### What's missing or wrong

The Windfree capabilities are not listed in the device controls.

I have added the localthings file and the smartthings file so you can compare (dunno if that is necessary at all).

I have three or four different windfree air conditioners. They all report (in localthings) as TP1X_DA-AC-RAC-01001_0000.

I also noticed that I can switch the air conditioners on, even change mode with some effort, but can not switch them off again.

I asked claude ai to compare the two devices. here's what claude said:

SmartThings vs localthings - Living Room AC (TP1X_DA-AC-RAC-01001_0000)

SmartThings - 12 entities:
climate.air_living_room
select.air_living_room_dust_filter_alarm_threshold
sensor.air_living_room_volume
sensor.air_living_room_energy
sensor.air_living_room_power
sensor.air_living_room_energy_difference
sensor.air_living_room_power_energy
sensor.air_living_room_energy_saved
sensor.air_living_room_humidity
sensor.air_living_room_temperature
switch.air_living_room_display_lighting
switch.air_living_room_sound_effect

localthings - 18 entities:
climate.living_room_air_living_room
sensor.samsung_airconditioner_tp1x_da_ac_rac_01001_0000_connection_mode_3
sensor.living_room_air_living_room_alarm_code
sensor.living_room_air_living_room_current_limit_level
sensor.living_room_air_living_room_energy
sensor.living_room_air_living_room_energy_saved
sensor.living_room_air_living_room_filter_usage
sensor.living_room_air_living_room_filter_status
sensor.living_room_air_living_room_self_check_status
sensor.living_room_air_living_room_self_check_result
sensor.living_room_air_living_room_self_check_error
binary_sensor.living_room_air_living_room_current_limit_enabled
binary_sensor.living_room_air_living_room_firmware_update_available
switch.living_room_air_living_room_display_light
switch.living_room_air_living_room_air_purification
switch.living_room_air_living_room_auto_clean
switch.living_room_air_living_room_mute_once
button.living_room_air_living_room_start_self_check

In SmartThings, not in localthings:
sensor.air_living_room_volume
sensor.air_living_room_power
sensor.air_living_room_energy_difference
sensor.air_living_room_power_energy
sensor.air_living_room_humidity
sensor.air_living_room_temperature
[...]

Issue #38, opened 2026-07-22
Incomplete capability coverage for Samsung Airconditionor


sameerismair:
### Appliance model

AR50F18C1JHNJO

### What's missing or wrong

This device is missing full capability coverage. Either its appliance type wasn't recognized, or some of the resources it exposes aren't modeled yet. It'll keep working with whatever is already supported. You can help expand support by going to Settings > Devices & Services > Samsung Airconditioner (TP1X_DA-AC-RAC-01001_0000) > the menu (top right) > Download diagnostics, then filing it with the linked issue template.

### Diagnostics file

[config_entry-localthings-01KY4YF06H1ZDP87MX0YVFH3K6.json](https://github.com/user-attachments/files/30269565/config_entry-localthings-01KY4YF06H1ZDP87MX0YVFH3K6.json)

mbillow:
Thanks for the diagnostics dump — really appreciate you taking the time to file this, it made it possible to add real support here.

Your unit (`TP1X_DA-AC-RAC-01001_0000`) already reported a recognized `oneUiVersion`, so the appliance type resolved correctly — the gap was in our air-conditioner capability coverage itself. I've pushed a fix (branch `claude/ac-support-triage-3zyjj9`) that adds:
- **Display light** switch (`/light/vs/0`)
- **Mute once** switch (`/option/muteonce/vs/0`)
- **Self-check** status/result/error sensors + a "Start self-check" button (`/selfcheck/vs/0`)
- **Current limit** level, exposed as read-only sensors (`/electriccurrent/vs/0`) — see below

A few resources in your dump describe real features we don't fully understand yet, so we left them unmapped rather than guess at a write contract against live HVAC hardware. If you're willing, screenshots of the SmartThings app for any of these would help us model them properly in a follow-up:
- **AI Sleep** (`/aisleep/vs/0`) — the app's AI Sleep control screen, if your unit has one.
- **Circuit-breaker / current limit** (`/electriccurrent/vs/0`) — we exposed the reported level as read-only, but don't know what the values (`3`–`9`) mean (amps? tiers?) or how it's normally set. A screenshot of wherever this is configured (device settings, or the physical remote's menu) would help.
- **Welcome cooling/heating** (`/welcome/temperature/vs/0`) — if there's a "Welcome" feature toggle in the app.
- **Reservation schedule** (`/reserverulesets/vs/0`) — if you have any on/off timers set up, a screenshot of that schedule screen.

[...]

PR #91, opened 2026-07-26
Add cool-only global RAC support: WindFree, Auto mode, and display light

pedroperosin:
## Summary
Adds support for Samsung TP1X cool-only global room air conditioners (e.g. `TP1X_DA-AC-RAC-01001`, `airconOptionList` includes `AI_RAC_GLOBAL_COOLONLY_3.0`), which previously loaded as an **unknown** device with only common capabilities. Confirmed against two live units — one cool-only, one heat-capable — plus all existing AC fixtures.

## Changes
**Detection** (`registry/by_type/__init__.py`) — these boards ship no `swVersionInfo` block, so `oneUiVersion` is empty and `for_device()` cannot route them; their `modelNum` spells the RAC token with hyphens (`-AC-RAC-`), which the existing `_RAC_` underscore match missed. Match the hyphenated `-RAC-` token in `for_device_by_model()`.

**Climate** (`climate.py`)
- Map device `Auto` -> `HVACMode.AUTO` (was `HEAT_COOL`, which renders Heat/cool and implies a two-setpoint range these single-setpoint / cool-only units do not have).
- Expose **WindFree** as a preset: on these boards it is the convenient mode `Nano` (`NanoSleep` = WindFree + sleep), in place of `Smart`. Confirmed by diffing a WindFree-off dump (`Off`) vs on (`Nano`).

**Display light** (`registry/capabilities/airconditioner.py`)
- Display-light **switch** for boards that encode the panel light in `/mode/vs/0`''s options blob instead of a `/light/vs/0` switch (mutually exclusive across all observed dumps). Read + single-token `SingleCommand` write, both accounting for the token being **inverted** vs its name (`Light_Off` = panel lit) — confirmed by a live toggle; mirrors `air_purifier.py`''s `Light` switch.
- Ignore `/stepcontrol/vs/0` and `/remotedeviceinfo/vs/0` to clear the coverage-gap repair.

**Tests / fixtures** — new scrubbed fixture + golden for the cool-only unit, detection/coverage/write regression tests, and the three regenerated AC goldens that gain `display_light`.

## Validation
HA-free discovery harness across all six AC fixtures: each resolves to the airconditioner registry, zero unbound hrefs, goldens match; no non-AC fixture misroutes; en/nl translations mirror. All behavior (climate control, WindFree, Auto label, display-light toggle that sticks without disturbing other state) verified live on hardware.

> Note: the full `pytest` suite needs `pytest-homeassistant-custom-component` (Python >= 3.13) and my machine has 3.12, so I validated with the HA-free registry harness. CI should run the full suite.

### DIFF
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -24,6 +24,7 @@ from homeassistant.components.climate import (
     ClimateEntity,
     ClimateEntityFeature,
     HVACMode,
+    PRESET_NONE,
 )
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import UnitOfTemperature
@@ -96,25 +97,17 @@ _DEVICE_TO_SWING: dict[str, str] = {
 }
 _SWING_TO_DEVICE = {v: k for k, v in _DEVICE_TO_SWING.items()}
 
-# Preset (convenient mode): Off/Sleep map onto HA standard presets; Quiet/Smart/
-# Speed and WindFree are custom (translated). On cool-only global RAC variants
-# (e.g. TP1X_DA-AC-RAC-01001) Samsung's WindFree still-air cooling is exposed
-# through this convenient-mode resource under the device code 'Nano' (and
-# 'NanoSleep' for WindFree + sleep) in place of the 'Smart' code seen on other
-# boards -- confirmed empirically by diffing a dump with WindFree off (code
-# 'Off') against one with it on (code 'Nano'). Codes absent from this map are
-# dropped from preset_modes (see _read_modes), so a board that lists neither
-# Smart nor Nano simply won't offer them.
-_DEVICE_TO_PRESET: dict[str, str] = {
-    'Off': 'none',
-    'Sleep': 'sleep',
-    'Quiet': 'quiet',
-    'Smart': 'smart',
-    'Speed': 'speed',
-    'Nano': 'windfree',
-    'NanoSleep': 'windfree_sleep',
-}
-_PRESET_TO_DEVICE = {v: k for k, v in _DEVICE_TO_PRESET.items()}
+# Preset (convenient mode): resolved dynamically from the device's own
+# /mode/convenient/vs/0 supportedModes -- no per-model table. The device 'Off'
+# code maps to HA's PRESET_NONE ("no preset active"); every other code is exposed
+# as its lowercased self and labelled in translations
+# (entity.climate.airconditioner.state_attributes.preset_mode.state.<code>), so
+# any board's convenient modes surface without code changes and an unlabelled
+# code renders as its raw value until a label is added. (Samsung's WindFree
+# still-air cooling shows up here as the 'Nano'/'NanoSleep' codes on cool-only
+# global RAC boards -- that's just a translation label, not a hard-coded mode.)
+def _preset_to_ha(code) -> str:
+    return PRESET_NONE if code == 'Off' else str(code).lower()
 
 
 async def async_setup_entry(
@@ -316,11 +309,12 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
 
     @property
     def preset_mode(self):
-        return self._read_mode(CONVENIENT_HREF, _DEVICE_TO_PRESET)
+        code = _first(self._rep(CONVENIENT_HREF).get(_MODES_FIELD))
+        return _preset_to_ha(code) if code is not None else None
 
     @property
     def preset_modes(self) -> list[str]:
-        return self._read_modes(CONVENIENT_HREF, _DEVICE_TO_PRESET)
+        return [_preset_to_ha(c) for c in self._supported(CONVENIENT_HREF)]
 
     # -- writes -------------------------------------------------------------
 
@@ -378,4 +372,9 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
         await self._set_mapped('swing', _SWING_TO_DEVICE, swing_mode)
 
     async def async_set_preset_mode(self, preset_mode: str) -> None:
-        await self._set_mapped('preset', _PRESET_TO_DEVICE, preset_mode)
+        # Reverse-resolve against the unit's own supportedModes (codes aren't a
+        # fixed transform of the HA value -- e.g. 'NanoSleep' -> 'nanosleep').
+        for code in self._supported(CONVENIENT_HREF):
+            if _preset_to_ha(code) == preset_mode:
+                await self.coordinator.async_send_command(self._bound, ('preset', code))
+                return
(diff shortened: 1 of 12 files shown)


custom_components/localthings/climate.py
```diff
@@ -88,13 +92,22 @@
...
-# Speed are custom (translated).
+# Speed and WindFree are custom (translated). On cool-only global RAC variants
+# (e.g. TP1X_DA-AC-RAC-01001) Samsung's WindFree still-air cooling is exposed
+# through this convenient-mode resource under the device code 'Nano' (and
+# 'NanoSleep' for WindFree + sleep) in place of the 'Smart' code seen on other
+# boards -- confirmed empirically by diffing a dump with WindFree off (code
+# 'Off') against one with it on (code 'Nano'). Codes absent from this map are
+# dropped from preset_modes (see _read_modes), so a board that lists neither
+# Smart nor Nano simply won't offer them.
 _DEVICE_TO_PRESET: dict[str, str] = {
```
mbillow: We really should just be reading this dynamically from the device instead of hard coding it. I've been meaning to update the skill to specifically say not to do things like this. It creates implementations that are way too single device model specific.

mbillow:
Overall, this looks good. If you could rip out that static mode array and add the dynamic select option resolution, that would be awesome.

One more ask, please amend any commits to remove agentic committer details. Commits should be made as you (both committer and author name) with no co-authors.

Thanks for the contribution!

pedroperosin:
Thanks for the review! Addressed both points:

- **Dynamic presets:** ripped out the static convenient-mode table. `preset_modes`/`preset_mode` now resolve from `/mode/convenient/vs/0`'s `supportedModes` at runtime — the device `Off` code maps to `PRESET_NONE`, every other code is exposed lowercased, and the labels live in `translations` (`entity.climate.airconditioner.state_attributes.preset_mode.state.<code>`). So any board's convenient modes surface without a per-model table, and an unlabelled code renders as its raw value until a label is added.
- **Commits:** reauthored — author and committer are me, no co-authors.

All verified against the fixtures/goldens and both live board variants (cool-only and heat-capable).

One aside, not included here to keep the PR scoped: I noticed a minor cosmetic lag where the card briefly shows the last cached setpoint right after a power-on (the unit restores its own state, and temp/power aren't push-subscribed on these boards, so it re-syncs on the next poll or interaction). Happy to open a follow-up for a short delayed re-poll after power-on if that's of interest.

mbillow:
Thanks for the quick turnaround on the presets — that's exactly what I was after. Four things I'd like addressed in this PR before it goes in.

### 1. Vendor temperature write should carry only `{id, desired}`

We've verified on other devices that an `x.com.samsung.da.items` write only needs the array item with the right `x.com.samsung.da.id` plus the field being changed — the device merges the rest itself. Same fact as the `options` case in #54, different array. So the vendor branch shouldn't be echoing `current`/`minimum`/`maximum`/`increment`/`unit` back at the unit; `current` in particular is a sensor reading we have no business writing.

The catch is that these two things are coupled, so please don't do only the first half. `async_send_command` currently reconciles partial writes for exactly one array shape — `x.com.samsung.da.options`, via `merge_options_field`. `observe.apply()` merges field-by-field, so a bare `{'x.com.samsung.da.items': [{id, desired}]}` would replace the whole cached `items` field and drop current/min/max/unit for the full settle window. On vendor boards that's the card losing its current temperature and range for tens of seconds after every setpoint change.

What I'd like:

- a `merge_items_field(cached, new_items)` next to `merge_options_field` in `registry/capabilities/common.py`, matching items by `x.com.samsung.da.id` and merging fields within the matched item (append if the id isn't present)
- wire it into `async_send_command` alongside the existing `options` pre-merge, so the optimistic cache entry stays complete
- `_climate_write`'s `temperature` branch then returns just `{'x.com.samsung.da.id': <id>, 'x.com.samsung.da.desired': <value>}`

That should also let `async_set_temperature` drop the `payload[2]` passthrough and go back to a plain `('temperature', temp)` — the platform no longer needs to hand the current item down for the write to be RMW-safe.

### 2. `StateCache.get()` doesn't take a default

`test_climate_power_write_applies_to_its_own_href_not_bound_href` fails on `main`'s test env:

```
E   TypeError: StateCache.get() takes 2 positional arguments but 3 were given
```

The signature is `get(self, href) -> Optional[dict]`. The two rewritten assertions use the dict-style two-arg form; they want `(coordinator._cache.get(href) or {})`, which is the idiom the coordinator itself uses. Only the assertions are wrong — with them fixed the test passes and the write correctly targets `/power/vs/0`, so the behaviour is fine.

Worth setting up an env that can run the suite, since CI hasn't run on this branch either. Python 3.13 works (`python3.13 -m venv .venv && .venv/bin/pip install -r requirements-dev.txt`); 3.12 is what's blocking you.

### 3. Missing preset labels on boards we already support

Resolving presets dynamically means every board's convenient-mode codes now surface, and five of them have no label — so existing users get raw lowercased codes in the dropdown:

| fixture | untranslated |
|---|---|
| `airconditioner_caww_tp2` | `longwind` |
| `airconditioner_tp1x_da_ac_rac_01011` | `motionindirect`, `drycomfort` |
| `airconditioner_tp2x_rac_20k` | `motionindirect`, `motiondirect` |

Please add those to `en.json`/`nl.json` alongside the two new WindFree ones. The goldens don't catch this because presets aren't in `state_keys` — if you can see a cheap way to assert "every fixture's convenient codes have a label", that'd be a nice guard, but I won't hold the PR on it.

### 4. Please describe the write-path commit

`3db5366` changes the power write target, adds board-dependent temperature channel selection, and flips `_is_on()`'s precedence. That's a real fix and I want it — from the fixtures, `/power/0` is absent on three of six known AC boards, and our optimistic apply would have been materialising a phantom `/power/0` entry that nothing ever overwrote, which lines up with the "can turn on but not off" report in #38. But none of it is in the PR description, and it's touching the write path we've iterated on across #9/#17/#27/#38/#54. Please call it out in the summary so it gets read as the behavioural change it is.

Everything else looks good — detection fallback, the Auto mapping, and the dynamic presets are all clean, and 476/477 tests pass. Happy to take the display-light polarity question (inverted reading confirmed on one unit, now applied to four board families) as a follow-up rather than blocking this.

---

mbillow:
Correction to item 4 above: the "can turn on but not off" report is in #53, not #38. The #38 reference in the list of write-path issues is right.

---

mbillow (changes requested):

mbillow:
Thanks again for the contribution and for turning around the preset feedback so quickly! Since main has moved a fair bit since this branch was opened and the remaining items touch several files, I picked up the last four points myself and finished it on top of current main: #114.

It carries your commits' intent forward (WindFree, Auto mode, display light, the hyphenated detection fallback) plus the four items from the review above — slim vendor-temperature writes with a proper read-side merge, the `StateCache.get()` test fix, the missing preset labels, and a description of the write-path change. Closing this one in favor of #114; really appreciate the work that went into it.

---

PR #129, opened 2026-07-27
Add beep, tropical night mode, filter hours/threshold, air-quality and sw/fw version for AC

blka:
Adds the ARTIK051_PRAC-class AC entities that aren't in the current registry, layered additively on top of the existing airconditioner registry (no changes to the display-light / mute-once / WindFree / humidity / current-temperature entities already shipped).

## New entities

| Entity | Platform | OCF source | Write |
|---|---|---|---|
| `beep` | switch | `/mode/vs/0` option `Volume_*` | single-token `option_write` |
| `tropical_night_mode` | number 0–16 | `/mode/vs/0` option `Sleep_<N>` | single-token `option_write` |
| `air_filter_usage_hours` | sensor (duration, h) | `/filter/airdustfilter/vs/0` `filterUsage` | read-only |
| `air_filter_threshold` | sensor (duration, h) | `filterDesiredUsage` (500 h) | read-only |
| `clean_level`, `odor`, `dust`, `fine_dust` | sensor (diagnostic) | `/sensors/vs/0` items[] | read-only |
| `software_version`, `firmware_version` | sensor (diagnostic) | `/information/vs/0` items[] | read-only |

## Non-obvious bits worth flagging for review

- **Single-token options write.** Writing the full `x.com.samsung.da.options` array to `/mode/vs/0` (RMW with one slot swapped) is accepted (2.04) but **autonomously reverts on ARTIK051_PRAC within ~10 s**. A one-element body `{"x.com.samsung.da.options": ["Volume_100"]}` persists. `beep` and `tropical_night_mode` reuse the existing `option_write` single-token merge (same mechanism already used by the display-light switch), so this isn't a new write path — just calling out why these two controls use it.
- **`beep` token polarity:** `Volume_Mute` = off, `Volume_100` (any non-Mute) = on.
- **Air-quality sensors have no `device_class`/`state_class`** on purpose: `/sensors/vs/0` advertises no unit, and the CleanLevel/Odor/Dust/FineDust semantics aren't confirmed as numeric measurements vs categorical codes. Left as diagnostic scalars until a populated reading + unit is observed (don't-guess). `SuperFineDust` is present in the dump but intentionally not modeled.
- **`/information/vs/0`** is globally ignored as identity plumbing; the AC registry drops that ignore so `INFO` is the sole capability on the href (verified by test — `len(caps) == 1`).
- **`air_filter_threshold` is read-only locally.** Setting the alarm threshold is cloud-only (`samsungce.dustFilterAlarm`); the 500 h read is local. Same for filter reset (`custom.dustFilter`) and audio-feedback volume level (`samsungce.airConditionerAudioFeedback`) — not exposed here because no local OCF resource accepts them on this board (all candidate hrefs 4.04).

## Scope note

`AIR_FILTER` is reused by the dehumidifier registry, so `air_filter_usage_hours` and `air_filter_threshold` also surface there — correct, since dehumidifier exposes the same `/filter/airdustfilter/vs/0` with `filterUsage`/`filterDesiredUsage`. Golden regression updated for it.

## Tests

- 20 new tests in `test_airconditioner_capabilities.py` (read/write/gating/value_fn for each entity, including the absent-token gating cases across fixtures — e.g. `airconditioner_tp1x_da_ac_rac_01011` carries no `Volume_` token → no `beep`; `airconditioner_window_ac` carries no `Sleep_` → no `tropical_night_mode`).
- Golden regression regenerated for the 8 AC fixtures + dehumidifier.
- Translations: `en.json` + `nl.json` (every-language-mirrors-english invariant).
- Full suite: 608 passed.


mbillow (changes requested):
**Overall:** Well-constructed, genuinely additive PR. The write path reuses the existing single-token `option_write` merge rather than inventing a new one, the exists_fn gating is per-token so entities only materialize where the board actually reports them, and the golden regeneration across all 8 AC fixtures + dehumidifier is exactly the right way to prove that. The PR body is unusually good about flagging its own soft spots. I ran the suite locally and confirmed 608 passed.

My findings are concentrated in the two capabilities that read `items[]` arrays, where the "first match wins" reduction discards data the fixtures show is actually distinct.

### 1. `firmware_version` silently collapses multiple distinct Firmware items

`_info_item_number` returns the first `type == 'Firmware'` item, but half the fixture corpus reports several, each a different component:

| Fixture | Firmware items | Value surfaced |
|---|---|---|
| `tp1x_da_ac_rac_01011` | 3 | `02756C25082500,FFFFFFFFFFFFFF` (2 dropped) |
| `tp1x_rac` | 3 | `02678A24092500,FFFFFFFFFFFFFF` (2 dropped) |
| `tp1x_rac_coolonly` | 2 | first (1 dropped) |
| `tp2x_rac_20k` | 2 | first — **and the second item has no `x.com.samsung.da.number` key at all** |

These are separate MCUs (they parallel the `Outdoor` item that other boards report as its own type), each with its own `x.com.samsung.da.id`. Collapsing them into one `firmware_version` means the entity's meaning depends on device item ordering, and `test_info_item_number_picks_first_match` codifies that behavior without justifying it. This is the same ambiguity the "don't guess" rule exists for — which the PR invokes correctly for the air-quality `device_class`, but not here. Suggest keying on `x.com.samsung.da.id` (instance-style, like the fridge compartments) or at minimum documenting why item 0 is authoritative.

### 2. `air_filter_threshold` may not actually be cloud-only

The PR states setting the threshold is cloud-only via `samsungce.dustFilterAlarm`. But three fixtures (`tp1x_da_ac_rac_01011`, `tp1x_rac`, `dehumidifier`) advertise, on the same resource being read:

```
'filterDesiredUsage': '500',
'supportedFilterDesiredUsage': ['180', '300', '500', '700']
```

A supported-values enum sitting next to the current value on a live local resource is the exact shape of a locally-writable option — and `SelectDesc.options_field` exists for precisely this pattern. The PR says "all candidate hrefs 4.04", but `/filter/airdustfilter/vs/0` isn't a candidate href that 404'd; it's already bound and read. Was a POST of `filterDesiredUsage` to that resource actually attempted? If so, worth saying so in the comment; if not, this looks like a `SelectDesc` rather than a read-only sensor.

### 3. Hardcoded `unit='h'` where the device reports the unit

Both new filter sensors hardcode `unit='h'`, but the resource carries `filterCapacityUnit` (`'Hour'` on every fixture), and `_filter_usage_percent`'s own docstring right above already calls this out: *"reports `filterUsage` as a raw count in `filterCapacityUnit`"*. `SensorDesc.unit_fn` is the codebase's documented "static default, live override" hook for this. With `device_class='duration'`, a board reporting a different unit wouldn't just mislabel — it would produce wrong long-term statistics.

### 4. `air_filter_usage_hours` state class

`filterUsage` is a lifetime counter that increases monotonically and resets to zero on filter replacement. That's `state_class='total_increasing'`, not `'measurement'` — `total_increasing` is what makes HA handle the reset correctly in statistics. (I realize `air_filter_usage` uses `'measurement'` for the percentage, so there's local precedent, but the percentage is a derived ratio and this one is the raw counter.)

### 5. `INFO` entities have no `exists_fn` — inconsistent with `AIR_QUALITY` in the same PR

`AIR_QUALITY` correctly gates each sensor on its item type via `_has_sensor_type`. `INFO` doesn't, so `_is_included` falls back to `field in rep` — `x.com.samsung.da.items` is always present, so both version sensors always bind. A board with no `Firmware` item, or whose Firmware item lacks `number` (as `tp2x_rac_20k`'s second item already does), gets a permanently-`unknown` entity. Same one-liner as `AIR_QUALITY`:

```python
exists_fn=lambda rep, res: _info_item_number(rep.get('x.com.samsung.da.items'), 'Firmware') is not None
```

### 6. `_sensor_item_value` drops the second array element without comment

The PR deletes this ignore rationale:

> `# All-zero and ambiguously encoded on this model (2-value arrays); the 'don't guess' rule`

The 2-value arrays are still there — `Dust`, `FineDust`, and `SuperFineDust` all report `["0", "0"]` while `CleanLevel`/`Odor` report `["0"]`. `_sensor_item_value` takes `v[0]` and discards `v[1]` silently. The PR removes the note flagging this as ambiguous without resolving the ambiguity. What is the second element, and what's the evidence the first is the reading?

Relatedly, the "all-zero, so no `device_class`" premise is already falsified in the corpus: `tp1x_da_ac_rac_01011` reports `CleanLevel: ["1"]` **and** a corroborating top-level scalar `x.com.samsung.da.cleanLevel: "1"` on the same resource. That top-level field is decent evidence `clean_level` is numeric and could take `value_fn=_int` + `state_class='measurement'`. Note that `str()`-coercing now locks these into string states — switching to numeric later breaks recorder history for anyone who's run this.

### 7. `beep` write discards the volume level

`_beep_write` always writes `Volume_100` for On. The cloud counterpart (`samsungce.airConditionerAudioFeedback`) is a level, and the token is `Volume_<n>`, so off→on would clobber a user's intermediate setting. `rep` is already in hand — reading the previous non-Mute token via `_option_token(rep, 'Volume')` and restoring it (falling back to `100`) is cheap insurance. Every fixture only shows `Volume_Mute`/`Volume_100`, so this is speculative, but the fix costs two lines. Same reasoning applies to `_beep_on`'s `tok != 'Volume_Mute'`, which would read `Volume_0` as on.

### Smaller items

- **Test gap:** `test_air_quality_sensors_from_sensors_vs_items` only asserts the all-zero WindFree case. `tp1x_da_ac_rac_01011`'s `clean_level == '1'` is the only non-zero air-quality reading in the whole corpus and the one assertion that would actually catch a `value_fn` regression.
- **Naming collision:** `_AC_IGNORED` now means two different things in two files — a list of `Capability` objects in `by_type/airconditioner.py`, a list of href strings in `capabilities/airconditioner.py`. The new comment "Removed from `_AC_IGNORED` below so AIR_QUALITY is the sole cap" is now ambiguous about which. Suggest `_IGNORED_LESS_INFO` for the new one.
- **Unmodeled siblings:** `Outdoor` appears in 4 fixtures with real values (`20091600,10000400`) and `Touch IC` in 2, both skipped while `Software`/`Firmware` are modeled. Same for `SuperFineDust`, which the body says is deliberate but doesn't say why. If these opaque scalars are safe to expose, the omitted ones are equally safe — worth one sentence either way.
- **Golden nuance (pre-existing, not introduced here):** `caww_tp2` reports no `filterDesiredUsage` at all, yet its golden gained `air_filter_threshold`. That's because `flatten` applies `exists_fn` but not `_is_included`'s `field in rep` fallback, so the golden lists a key HA would never create an entity for. Not this PR's bug, but this PR is where it becomes visible.

### What's good

- Single-token `option_write` reuse instead of a new write path, with the ~10 s revert behavior documented at the call site.
- Per-token `exists_fn` gating, verified against fixtures that lack the token — the `beep`-absent-on-`tp1x_da_ac_rac_01011` and `tropical_night`-absent-on-`window_ac` tests are the right shape.
- `/information/vs/0` de-ignoring is properly scoped to the AC registry, with `len(caps) == 1` locking the invariant. No redaction impact — `redact.py` is key-name based, so `serialNum`/`otnDUID` stay redacted regardless.
- Dehumidifier scope call-out is correct; it has its own `_DHM_IGNORED` covering `/sensors/vs/0`, so no coverage gap from the AC change.
- Translations mirrored in both catalogs.

---

Items 1, 2, 3, and 5 are the ones I'd want addressed before merge; the rest are discussion or follow-up.

---

blka:
Pushed 51fa1c5 addressing 1-7 + the smaller items. 612 tests pass.

**1. INFO per-item versions.** Each `/information/vs/0` version item is now exposed per `(type, ordinal)` via `_info_version(items, type_, ordinal)` + `_has_info_version` exists_fn, instead of collapsing to a single first-Firmware value. `firmware_version` is the first Firmware MCU; `firmware_version_2`/`_3` the additional MCUs (tp1x boards report 3); `outdoor_unit_version` and `touch_ic_version` (window AC) where present. An item with no `x.com.samsung.da.number` (tp2x_rac_20k's second Firmware) is suppressed by exists_fn rather than binding unknown.

**2. `air_filter_threshold` is locally writable.** You were right -- I had not POSTed `filterDesiredUsage` to `/filter/airdustfilter/vs/0`; the "all candidate hrefs 4.04" was for the `samsungce.dustFilterAlarm` capability, not this resource. Confirmed live on ARTIK051_PRAC (192.168.20.60): POST `filterDesiredUsage=700` -> 2.04, read-back 700, persisted; restored to 500. Converted to a `SelectDesc` with `options_field='x.com.samsung.da.supportedFilterDesiredUsage'` and a `write_fn` POSTing the scalar field. It only binds where the device advertises the enum; ARTIK051_PRAC (this unit) doesn't advertise it, so the writable field stays unexposed there rather than guess the valid set.

**3. `unit_fn`.** `air_filter_usage_hours` now reads `filterCapacityUnit` via `unit_fn` ('Hour' -> 'h', 'Minute' -> 'min', 'Second' -> 's'; static 'h' fallback when absent), instead of a hardcoded unit.

**4. `state_class`.** `air_filter_usage_hours` is `total_increasing` (lifetime counter, resets on filter replacement), not `measurement`.

**5. `exists_fn`.** INFO and `air_filter_threshold` now gate via exists_fn (AIR_QUALITY already did). This also resolves the caww_tp2 golden nuance -- `filterDesiredUsage` is absent there, so no `air_filter_threshold` key, matching what HA would create.

**6. Air-quality.** `clean_level` is corroborated as numeric by the top-level `x.com.samsung.da.cleanLevel` scalar (tp1x_da_ac_rac_01011 reports both as '1'), so it's an int `measurement`; odor/dust/fine_dust stay string diagnostics. Reinstated the 2-element-array ambiguity note (Dust/FineDust/SuperFineDust report ['0','0']; v[1] unconfirmed, v[0] taken as the reading). `SuperFineDust` is now modeled for consistency with Dust/FineDust (same shape) rather than skipped without reason.

**7. `beep` write.** `_beep_write` restores the last non-Mute `Volume_` level on 'On' (read via `_option_token(rep, 'Volume')`) instead of forcing `Volume_100`, falling back to 100 when no prior level is known or the prior token is Mute. Every fixture only shows `Volume_Mute`/`Volume_100`, so this is insurance for an intermediate cloud-set level.

**Smaller.** `test_air_quality` now asserts the tp1x_da_ac_rac_01011 `clean_level == 1` non-zero reading (the one value_fn-regression catch in the corpus); renamed `by_type`'s `_AC_IGNORED` -> `_IGNORED_LESS_INFO` to disambiguate from the capabilities module's `_AC_IGNORED` (href strings); golden regenerated for the 8 AC fixtures + dehumidifier.

custom_components/localthings/registry/by_type/airconditioner.py
```diff
@@ -12,16 +12,25 @@
 from ..capabilities import airconditioner, common, dishwasher, ignored
 from ._base import DeviceRegistry, _build
 
+# /information/vs/0 is globally ignored (serial/model identity plumbing), but
+# the AC exposes Software/Firmware version in its items[] -- model those here
+# and drop the no-entity coverage entry so INFO is the sole cap on the href.
+# Renamed from the capabilities module's _AC_IGNORED (a list of href strings)
+# to avoid the two-meaning collision noted in review.
+_IGNORED_LESS_INFO = [c for c in ignored.IGNORED if c.href != '/information/vs/0']
```
mbillow: We either should expose this on all device types or exclude it here. Personally, I excluded it because it felt useless to include. What is the user story for exposing this when we have no control over it? It just seems like exposing data for the sake of exposing it, which just leads to clutter.

blka: Dropped. `/information/vs/0` stays in the global ignore on every registry (reverted the AC-only de-ignore), and the INFO capability + `software_version`/`firmware_version`/`firmware_version_2`/`_3`/`outdoor_unit_version`/`touch_ic_version` entities, helpers, translations and tests are gone. Conforms to the existing 'no version entities' stance rather than expanding the surface for a read-only value with no user story. 625 tests pass (1bbecfa).

mbillow:
Thanks for this, @blka — great write-up and solid hardware verification on the threshold write and beep polarity.

#146 merged into `main` while this was open and created a conflict on the fork branch (both PRs modeled the same `/mode/vs/0` option tokens under different entity names) that we couldn't push a resolution to directly. Continued in #164 with the conflict resolved and a couple of small fixes — closing this one in favor of that.

Continued in #164.


---

Issue #136, opened 2026-07-27
Samsung A/C Windfree (ARTIK051_KRAC_18K)


mindvisionro:
### Appliance model

ARTIK051_KRAC_18K

### What's missing or wrong

Missing AC controls. Only on/off power is available.

### Diagnostics file

[config_entry-localthings-01KYJFZJZ6KNEM4DP68YGRT17MFF.json](https://github.com/user-attachments/files/30431149/config_entry-localthings-01KYJFZJZ6KNEM4DP68YGRT17MFF.json)

perseus177:
I have the same unit (ARTIK051_KRAC_18K, same `10193441` assay, build `200825`) and opened #146 for it.

Your dump was enough to find why it reports only power: `for_device_by_model`'s `_RAC_` check can't match `_KRAC_`, so the device falls back to `unknown` and 8 of the 19 resources stay unbound. Beyond detection, this board keeps fan speed and vane direction in `/airflow/vs/0` (no `/wind/*` at all) and its convenient-mode preset in a `Comode_*` options token (no `/mode/convenient/vs/0`), so those need fallbacks too.

With that in place the unit gets a full climate entity - HVAC modes, target/current temperature, fan `auto...turbo`, swing, and presets including WindFree (`Comode_Nano`) - plus SPI, auto clean, air monitoring, beep volume, Good Sleep, outdoor temperature (your `OutdoorTemp_81` = 26 C; the token is offset by 55) and filter time (your `FilterTime_5460` = 546 h; tenths of an hour). Everything except Good Sleep was confirmed by read-back on my unit.

Note your dump has no `Spi` token where mine does, so that switch simply won't appear for you - the entities are gated on the tokens actually present.

mindvisionro:
Hello! I've tested your modified files, and 90% of the functions are working. What's strange is that I have two Samsung AC devices: one is a WindFree model(AR09NXWXCWKNEU) and the other is a standard one (without the WindFree feature, model AR12RXWSAURNEU). Oddly enough, the WindFree device doesn't work with the WindFree preset, but the non-WindFree one does! :) Also, turning the beep sound on/off when changing settings isn't working.

I don't know if it helps, but I've attached the JSON data for both devices from the SmartThings developer platform.

[AC.JSON](https://github.com/user-attachments/files/30456260/AC.JSON) - 
[AC-WINDFREE.JSON](https://github.com/user-attachments/files/30456261/AC-WINDFREE.JSON)

[...]

PR #146, opened 2026-07-28
Support ARTIK051_KRAC_18K air conditioners (issue #136)

perseus177:
Fixes the ARTIK051_KRAC_18K room air conditioner from #136, which resolved to
`device_type: unknown` and exposed nothing but a power switch: 8 of its 19
resources were unbound. It now binds all 19 and gets a full climate entity plus
seven extra entities.

The `_RAC_` token check in `for_device_by_model` can't see `_KRAC_` — the `K`
sits between the underscore and `RAC` — and there's no `oneUiVersion` on this
board to fall back on. Fixing detection alone isn't enough, though: this
generation predates every AC dump the registry was built from and differs in
three ways.

### What's different about this board

* **No `/wind/*` resources at all.** Fan speed and vane direction share one
  `/airflow/vs/0`, whose `speedLevel` uses the same 0-4 scale as
  `_DEVICE_TO_FAN` and whose `direction` uses the same codes as
  `_DEVICE_TO_SWING`, so the existing maps are reused rather than duplicated.
  The resource reports no `supportedModes`, so the full scale is offered.
* **No `/mode/convenient/vs/0`.** The convenient-mode preset is a `Comode_*`
  token in `/mode/vs/0`'s options, synthesised into a convenient-shaped rep so
  `_preset_to_ha`'s dynamic resolution keeps working untouched. `Nano` lands on
  the existing `nano` preset, already labelled WindFree; `2Step` had no catalog
  entry in either language, so this adds one.
* **Settings as options tokens.** SPI, auto clean, air monitoring, beep volume,
  Good Sleep, outdoor temperature and filter time are `options` tokens here,
  where newer boards have dedicated resources. Writes reuse `option_write`'s
  single-token merge — the same mechanism the display light already uses on this
  href.

### Gating

Newer families carry some of the same tokens *alongside* dedicated resources for
those settings, so all the token entities are gated on this generation's
resource shape (`/airflow/vs/0` present, `/wind/strength/vs/0` absent — the same
test `climate.py`'s fan/swing fallback uses, so the two can't disagree). The
regression harness caught this: ungated, the tokens duplicated auto clean on
TP1X/TP2X boards and applied this board's outdoor-temperature calibration to
theirs. Every existing golden is unchanged.

### Calibrations, from hardware rather than from the token names

* **`OutdoorTemp` is offset by 55.** Token 75 against a 20.3 °C outdoor
  thermometer in the same install, token 74 against a 19.4 °C forecast.
  Fahrenheit fits far worse (74 °F = 23.3 °C); #136's 81 gives 26 °C in a
  warmer climate, which is also plausible.
* **`FilterTime` is tenths of an hour.** Token 1710 while the official Samsung
  app displayed "171 hours 0 minutes" for the same unit's filter, and the app's
  "0 minutes" pins the scale.
* **Humidity** now falls back to the plain `x.com.samsung.da.humidity` where
  `fivepercentHumidity` is absent, still as one entity, and 0 reads as "not
  measuring" rather than 0%. On this board the field carries a real reading
  (51%, matching the same unit's cloud integration at that moment) only while
  Air monitoring is on — which the unit switches back off by itself after about
  a minute, which is why it looks permanently stuck at 0 in most dumps.

### Verified

Read *and* write confirmed by read-back on my own unit (same model as #136):
HVAC modes, target/current temperature, fan `auto…turbo`, swing
`off/vertical/horizontal/both`, all six presets, SPI, and beep volume.

`pytest tests/ -q` — 654 passed, hassfest green
([run](https://github.com/perseus177/localthings/actions/runs/30317915463)). The
HACS job fails on my fork only because a fork has no topics and issues disabled.

### Deliberately not claimed

* **Filter time direction.** It was seen *rising* (171.0 → 171.5) while the unit
  ran, which contradicts the app's "remaining" wording, so the entity is named
  neutrally and no `remaining`/`elapsed` semantics are asserted.
* **Good Sleep's upper bound** (0-12 h) is a guess; only `Sleep_0` has been
  observed, so a write above 0 is unverified.
* **`/airflow/0`** — the OCF-standard mirror, identical values — is ignored
  rather than modelled. `air_purifier.py` found the *opposite* reliability
  ordering between these two hrefs on its own family, so neither form seemed
  safe sight-unseen; I verified writes on the vendor one only.
* `FilterAlarmTime`, `CoolCapa`, `WarmCapa`, `OptionCode`, `ExtendOptionCode`
  and `RacInfo` stay unmodelled per the "don't guess" rule.

### Notes for review

* The fixture is a scrubbed diagnostics dump from my unit rather than #136's,
  since that's the one the writes and calibrations were verified on. #136's unit
  is the same model with a slightly different token set (no `Spi`,
  `FilterTime_5460`, `OutdoorTemp_81`), which the presence gating handles the
  same way.
* If you'd rather the token entities were available to newer families too, the
  gate is one helper (`_is_legacy_board`) — but that needs a dump plus an
  in-app cross-check from one of those boards first, at least for the
  temperature offset.
* The `nl.json` strings are from a non-native speaker; happy to drop them if
  you'd rather a native speaker filled them in.

Refs #136

### DIFF
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -50,6 +50,7 @@ from .registry.capabilities.airconditioner import (
     HREF_WIND_DIRECTION as WIND_DIRECTION_HREF,
     HREF_WIND_OSCILLATION as WIND_OSCILLATION_HREF,
     HREF_CONVENIENT as CONVENIENT_HREF,
+    HREF_AIRFLOW as AIRFLOW_HREF,
 )
 from .registry.capabilities.common import normalize_temp_unit
 
@@ -228,8 +229,45 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
 
     # -- resource helpers ---------------------------------------------------
 
+    # Preset codes on legacy ARTIK051 boards, learned by driving the same unit
+    # through its cloud integration and reading the local token back each time:
+    # Nano=windFree, Quiet, Comfort, 2Step, Speed=Fast Turbo, Off=none.
+    _LEGACY_PRESET_CODES = ('Off', 'Nano', 'Quiet', 'Comfort', '2Step', 'Speed')
+
+    def _legacy_convenient(self) -> dict:
+        """A /mode/convenient/vs/0-shaped rep built from the Comode_* token in
+        /mode/vs/0's options, for boards that have no convenient resource."""
+        options = (self.coordinator.resource(MODE_HREF) or {}).get(
+            'x.com.samsung.da.options') or []
+        for option in options:
+            if isinstance(option, str) and option.startswith('Comode_'):
+                return {_MODES_FIELD: [option.split('_', 1)[1]],
+                        _SUPPORTED_FIELD: list(self._LEGACY_PRESET_CODES)}
+        return {}
+
+    def _legacy_airflow(self) -> dict:
+        """The /airflow/vs/0 rep, but only when it is the fan/swing channel to
+        use -- i.e. this board has no /wind/strength/vs/0."""
+        if self.coordinator.resource(WIND_STRENGTH_HREF):
+            return {}
+        return self.coordinator.resource(AIRFLOW_HREF) or {}
+
+    def _legacy_preset(self) -> bool:
+        """Whether presets come from the Comode_* token rather than a resource.
+
+        Gated on the same board test as _legacy_airflow, not on the convenient
+        rep being empty alone: newer boards carry Comode tokens too, so a
+        momentarily empty /mode/convenient/vs/0 there must not silently switch
+        the preset read (and write) over to the token path.
+        """
+        return (not self.coordinator.resource(CONVENIENT_HREF)
+                and bool(self._legacy_airflow()))
+
     def _rep(self, href: str) -> dict:
-        return self.coordinator.resource(href) or {}
+        rep = self.coordinator.resource(href) or {}
+        if not rep and href == CONVENIENT_HREF and self._legacy_airflow():
+            return self._legacy_convenient()
+        return rep
 
     def _is_on(self) -> bool:
         # Prefer the vendor /power/vs/0 (present on every observed board and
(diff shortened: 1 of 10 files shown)


mbillow:
Thanks for the PR.

> OutdoorTemp is offset by 55. Token 75 against a 20.3 °C outdoor thermometer in the same install, token 74 against a 19.4 °C forecast. Fahrenheit fits far worse (74 °F = 23.3 °C)

I’m not sold, this offset seems completely arbitrary. It could also be explained by the outdoor sensor being next to a warm compressor or being in the sun. 

> FilterTime is tenths of an hour. Token 1710 while the official Samsung
app displayed "171 hours 0 minutes" for the same unit's filter, and the app's
"0 minutes" pins the scale.

Did you confirm this in the app? It says 171 hours? A single digit minute block doesn’t make a whole lot of sense.

--- 

I’ll take a closer look at the actual implementation in a bit.

perseus177:
Screenshot_20260728_081233_SmartThings

From app

perseus177:
About outdoor sensor from AC
At midnight it was 20.3, different outdoor sensor show 20.5
At 8Am it was 25, different outdoor sensor show 24

mbillow:
Thanks for confirming.

perseus177:
### Correction: beep volume writes are accepted but not retained

Someone running this branch reported that pulling the volume to zero doesn't stop the appliance beeping, so I measured it properly. My "confirmed by read-back" claim above was too strong for this one entity, and I've pushed a commit correcting the descriptor comment and the test docstring instead of leaving it as-is.

What actually happens on my unit — write, then poll `/mode/vs/0` every 20 s:

| written | reads back | reverts to `Volume_100` after |
|---|---|---|
| 0 | `Volume_0` | ~100 s |
| 30 | `Volume_30` | ~80 s |
| 50 | `Volume_50` | ~60 s |

Every value lands and then the appliance re-asserts 100 on its own, so in HA the slider springs back and the beep never goes quiet.

Two things worth having on record, because they narrow it down:

* **It isn't a general property of the options blob.** SPI, written the identical way on the same href, held `Spi_On` for a full three-minute observation. So the single-token merge is fine; something specifically re-asserts `Volume`.
* **The cause is not established and I'm not claiming one.** The appliance's own SmartThings link is a candidate — its cloud-side `audioVolume` reads a stuck `100` and also silently ignores writes — but I haven't proven that, and I don't know whether the #136 unit or other boards retain the value.

I left the write in rather than making the entity read-only, since it does reach the device and this may be install-specific. Happy to drop the write, or gate the entity to read-only, if you'd rather ship the conservative version — your call.

Also worth noting for the reviewer: while checking this I verified the display-light inversion in `_display_light_on` behaves exactly as its comment describes on this board (`Light_Off` with the panel lit), so nothing to change there.

Issue #127, opened 2026-07-27
[Device support] ARTIK051_DONGLE_REF - RS68N8321S9, PRESTIGE COLLECTION, Side by Side, Fridge and Freezer


GSzabados:
### Appliance model

RS68N8321S9

### What's missing or wrong

I think most is covered by this issue https://github.com/mbillow/localthings/issues/83, but this is a combo unit.

Door sensors are missing, temperature sensors and controls are missing. If vacation mode supported to control (not the case for the cloud), then that is missing as well. Sabbath mode is not exposed on the physical device, so I am not sure if that should be supported, but it shows up in the integration. It shouldn't expose energy usage either, as far as I know the device doesn't support it, as I have not seen any of it in the SmartThings app.

Thanks for your hard work! This is a so great relief after the last cloud API announcements. This should have been made ages ago for all appliances. Thanks again for developing it!

### Diagnostics file

},
  "integration_manifest": {
    "domain": "localthings",
    "name": "LocalThings",
    "codeowners": [
      "mbillow"
    ],
    "config_flow": true,
    "dependencies": [],
    "documentation": "https://github.com/mbillow/localthings",
    "iot_class": "local_push",
    "issue_tracker": "https://github.com/mbillow/localthings/issues",
    "requirements": [
      "cbor2>=5.4.6",
      "pyOpenSSL>=23.0",
      "smartthings-local>=0.1.0"
    ],
    "version": "0.11.1",
    "is_built_in": false,
    "overwrites_built_in": false
  },
  "setup_times": {},
  "data": {
    "device_type": "unknown",
    "one_ui_version": "",
    "unbound_hrefs": [
      "/diagnosis/vs/0",
      "/door/cooler/0",
      "/door/freezer/0",
      "/temperature/current/cooler/0",
      "/temperature/current/freezer/0",
      "/temperature/desired/cooler/0",
      "/temperature/desired/freezer/0"
    ],
    "resources": {
      "/alarms/vs/0": {},
      "/temperatures/vs/0": {
        "x.com.samsung.da.items": [
          {
            "x.com.samsung.da.id": "0",
            "x.com.samsung.da.description": "Freezer",
            "x.com.samsung.da.desired": "-23",
            "x.com.samsung.da.current": "-23",
            "x.com.samsung.da.maximum": "-15",
            "x.com.samsung.da.minimum": "-23",
            "x.com.samsung.da.unit": "Celsius"
          },
          {
            "x.com.samsung.da.id": "1",
[...]

PR #149, opened 2026-07-28
Distinguish not-yet-fetched stub reps from confirmed-empty ones (issue #127)

mbillow:
## Summary

- `parse_device0_batch` used to collapse `/device/0`'s `{"href": "..."}` "no data fetched yet" marker into a plain `{}`, making it indistinguishable from a resource the device actually polled and confirmed empty.
- Every `exists_fn` using the "not rep or ..." stub carve-out (plus `entity._is_included`'s default field-presence gate) treated both cases the same way, so a resource a model simply doesn't support (e.g. this fridge's `/energy/consumption/vs/0`) spawned a phantom entity permanently stuck at "unknown" — the behavior reported in #127.
- Added `is_stub_rep()`, which recognizes only the literal `{"href": ...}` echo as a stub. A genuine `{}` is now treated as the device's real (if empty) answer and gates the entity off like any other missing field.
- Updated the 9 call sites replicating the stub carve-out by hand (energy meter, self-check error, cooktop burner state, range-hood auto-operation) to use the new helper.

While triaging #127 I also verified the door/temperature per-zone bindings the issue's last comment claimed were still missing — they're not; that comment was wrong (confirmed via source inspection and by replaying `discover()` against the issue's exact dump, which shows zero unbound hrefs). No code change was needed for that part.

## Test plan

- `pytest` needs Python 3.12+; this environment only had 3.11, so I built a throwaway 3.13 venv, installed `requirements-dev.txt`, and ran the real suite: all 674 tests pass.
- Replayed the golden-regression comparison (`discover()` + `flatten()`) across all 40 device fixtures against their golden files. Only 3 changed: `air_purifier`, `refrigerator_artik051_dongle_ref`, `refrigerator_artik051_dongle_ref_cooler` — each has a genuinely empty `/energy/consumption/vs/0` in its real captured dump, confirming those golden files had baked the phantom-entity bug in as "expected." Updated them to drop the now-correctly-excluded energy/power keys.
- Added `tests/test_batch.py` for `is_stub_rep`/`parse_device0_batch`.
- Updated the now-stale stub-assumption tests in `tests/test_common_capabilities.py` to cover both the true-stub and genuinely-empty cases.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QbWKT8SGG4mg1keRk46LJm)_


mbillow:
Ran an independent Opus review against this diff. It caught a real problem: the `entity.py:38` change (swapping the default field-presence gate's `not rep` for `is_stub_rep(rep)`) was much broader than the 9 hand-audited `exists_fn` call sites and would have silently excluded entities on any resource whose *normal* state is `{}` — `/alarms/vs/0` is documented in `fridge.py`'s `_active_alarm_codes` as validly empty when there's no active alarm, not a "this hardware doesn't support alarms" signal. Simulating the change against all 40 fixtures showed real entities disappearing on first-poll timing: `alarm_code` (16 fixtures), `job_beginning_status`, `diagnosis_status` (with an asymmetry against its sibling button), `sabbath_mode`, `defrost_delay`, `ice_maker_enabled` — the opposite of this PR's goal.

It also flagged that hiding `sabbath_mode` directly contradicted what I'd said on #127 itself ("I'd rather leave that alone rather than guess at hiding it").

Pushed 6239706 to fix: the default gate is back to including on either a stub or a genuinely-empty rep (verified byte-identical to the pre-#149 baseline across all 40 fixtures, apart from the 9 deliberately-audited `exists_fn` sites, which are untouched and still fix #127's phantom energy sensors). Added `tests/test_entity.py` covering `_is_included` directly, since it had zero prior test coverage — the golden-fixture harness can't see this class of bug at all (`adapter.flatten()` has no default field-presence gate, only golden fixtures which use `exists_fn`).

---

PR #212, opened 2026-07-30
Parallel DTLS probe + cached leaf (config-flow add ~14s -> ~1.4s)

### LINKED ISSUE #211: Config-flow add takes 30-40s — sequential DTLS handshake per false-positive port
## Summary

Adding an appliance through the config flow (enter IP → wait) takes **30-40 s on the happy path**, even when the device is reachable and answers correctly. The dominant cost is the **sequential per-port DTLS handshake** in `_probe_and_validate`, where every false-positive liveness candidate eats the full 12 s `HANDSHAKE_TIMEOUT_S` before the next port is tried.

## Repro / evidence

A fridge (`TP2X_REF_20K`, real DTLS port 49155) on a healthy LAN. The liveness sweep reports `[49154, 49155, 49153]` as live (silence = candidate — the 1-byte UDP probe can't distinguish a real DTLS server from a port that just doesn't return ICMP port-unreachable). The flow then handshakes them sequentially:

```
config_flow: Live DTLS port candidates on 192.168.x.x: [49154, 49155, 49153]
config_flow: port 49154 failed: DTLS handshake timeout to 192.168.x.x:49154   # ~12s
config_flow: port 49155 failed: DTLS handshake timeout to 192.168.x.x:49155   # (blacklist case) ~12s
config_flow: port 49153 failed: DTLS handshake timeout to 192.168.x.x:49153   # ~12s
```

Happy path (real port reached after the false-positive): ~12 s (49154 timeout) + ~1.4 s (49155 handshake) + GET ≈ 14-16 s. When the device's DTLS stack has throttled the source IP after repeated attempts (a real failure mode I hit), all three time out → ~36-39 s.

Same device, same cert, from a different source IP: `DTLS handshake OK in 1.42s`, `GET /oic/sec/acl -> 2.05`. So the handshake itself is fast; the flow's wall-clock is dominated by **N × 12 s sequential timeouts on false-positive ports**.

## Root cause

`custom_components/localthings/config_flow.py`, `_probe_and_validate` — the `for port in candidates` loop calls `DtlsCoapSession.connect()` (12 s timeout) one port at a time. The liveness probe (`_find_live_ports`, 1.5 s) is cheap but produces false-positives (any silent UDP port = "live"), and each false-positive costs a full 12 s handshake attempt before the next candidate.

## Proposed fix

**Race the handshake across all liveness candidates in parallel**, returning on the first port that completes a valid `GET /device/0`:

```python
def _race_handshake(host, candidates, worker):
    """worker(port) -> info dict | raises _CertRejected/TimeoutError/Exception."""
    cert_error_count = 0
    ex = ThreadPoolExecutor(max_workers=max(1, len(candidates)))
    try:
        futs = [ex.submit(worker, p) for p in candidates]
        for fut in as_completed(futs, timeout=RACE_OVERALL_S):
            try:
                return fut.result(), False            # winner
            except _CertRejected:
                cert_error_count += 1
            except Exception:
                pass
    finally:
        ex.shutdown(wait=False)                        # don't block on losers
    return None, cert_error_count == len(candidates)
```

Wall-clock becomes the real port's time (~1.4 s handshake + GET) instead of N × 12 s.

### Gotcha worth calling out explicitly

The obvious `with ThreadPoolExecutor(...) as ex:` form does **not** work here: `__exit__` calls `shutdown(wait=True)`, which blocks until every still-running losing worker returns. With real DTLS workers, that's ~12 s per loser, so the function returns at `max(winner, slowest_loser)` — i.e. the same wall-clock as the sequential loop, defeating the purpose. `shutdown(wait=False)` is required (the worker threads are daemons on Python 3.9+, so losers finish in the background and clean their sockets in `finally: sess.close()`).

### Optional companion: reuse the leaf cert for 2nd+ devices

The minted leaf cert (CA `AC14K_M` + the UUID from Samsung's cloud cert) is identical for every appliance, and is already stored in each config entry's data — but `_probe_and_validate` re-fetches the UUID from `connect-v2.samsungiotcloud.com:443` and re-mints on every add. Reusing the existing entry's leaf for the `has_creds` path skips the cloud fetch + mint (small time win, but a real **reliability** win: adding a 2nd+ device no longer depends on Samsung-cloud reachability). A fallback re-mint on all-cert-rejected (UUID rotation) keeps it self-correcting.

## Scope / alignment

- No `localthings` library change — the race reuses `DtlsCoapSession` as-is.
- The race races **callables** with an injectable `session_factory`, so it's transport-agnostic and slots into the transport-seam discussion in #168 (8888/HTTPS family) and #163 (TCP AC) if/when those land.
- No UI / error-mapping / `manifest.json` change.

## Ask

This is a real UX papercut (every new appliance add stalls for half a minute). Is a parallel-race PR welcome here, and is `shutdown(wait=False)` + daemon-worker background cleanup acceptable to the maintainer, or would you prefer a different shape (e.g. a per-worker `handshake_timeout` added upstream in `localthings.DtlsCoapSession` so losers give up faster)? Happy to open the PR either way — wanted to check the approach first since it touches the config-flow probe path.

blka:
Resolves #211

The observe grace-period early-exit (the other first-refresh latency win from this work) was split to #227 — it touches `observe.py`, not the config-flow probe path, so it's independent of the race and can land regardless of how the probe latency is approached (see the discussion on #211 around a library-level DTLS ClientHello liveness probe).

## What

Cuts localthings device-add latency on the happy path: races DTLS handshakes across the liveness candidates in parallel instead of sequencing through false-positive ports, and reuses the cached leaf cert for 2nd+ devices. Two commits here (the orphan-cooldown doc rides along — it edits a docstring the race adds, so it can't travel separately).

## Why / measured

Production trace adding a fridge (`TP2X_REF_20K`), config-flow probe (the part #211 is about):

| phase | before | after |
|---|---|---|
| config-flow probe (sequential false-positive DTLS timeouts) | ~14s | ~1.4s |
| coordinator first-connect orphan cooldown | ~8.4s | ~8.4s (documented — see §2) |

The probe went from ~14s to ~1.4s. (The ~15s observe grace wait and the ~1.8s discovery that made up the rest of the first refresh are separate; the grace wait is in #227.)

### 1. Parallel DTLS probe + cached leaf  (`feat(config-flow)`)
`_probe_and_validate` races DTLS handshakes across the liveness candidates in parallel (`ThreadPoolExecutor`; first valid `/device/0` wins; losers time out in the background via `shutdown(wait=False)` — the `with`-form's `shutdown(wait=True)` would block on losers and re-introduce the sequential wall-clock). 2nd+ device adds reuse the cached leaf cert from an existing entry (skip the Samsung-cloud UUID fetch + mint); on all-cert-rejected, a one-shot UUID-rotation re-mint fallback fires and the refreshed leaf is persisted to the existing entry. Race layer races callables with an injectable `session_factory`, so it's transport-agnostic for future TCP/8888 families. No `smartthings_local` library change.

### 2. Document post-probe DTLS orphan-cooldown gotcha  (`docs`)
After the race, the coordinator's first connect can be rejected for ~8s then succeed on retry. Root cause: the race's winning session closes on an **ephemeral** source port; the coordinator binds a **fixed** local port (`_local_source_port`), a different 5-tuple, so per RFC 6347 §4.2.8 it cannot evict the probe's just-closed session at handshake time and waits for age-out. One-time per device add, auto-recovers on retry. Documented rather than fixed: a session hand-off would defer the same ~8s to the first reconnect (the handed-off session is on an ephemeral port, which breaks the fixed-port eviction the coordinator relies on across reconnects), and racing on the fixed port breaks the parallel race (workers can't share a local UDP port — `recvfrom` mis-demux). This is the last visible latency chunk now that the probe races.

## Tests
Full suite: 1005 passed (with the grace-sleep commit; this branch's subset passes too — the race's tests are in `tests/localthings/test_config_flow.py`). No `smartthings_local` library change; no `manifest.json` change; no UI/error-mapping change.

## Note on overlap with library-level work
A DTLS ClientHello-based liveness gate is being prototyped in `smartthings-local` (see @QuiteYellow's comment on #211): it would identify the real DTLS port in ~1 RTT via the HelloVerifyRequest and pre-filter dead ports before any full handshake. If that lands in the library, it supersedes the integration-level race here (false-positives eliminated at the liveness stage → one candidate → no parallel race needed). The orphan doc (§2) is orthogonal to the liveness approach and stands either way; the race (§1) is the integration-level stopgap until/unless the library probe ships.

### DIFF
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -1,6 +1,7 @@
 """Config flow for Local Things integration."""
 from __future__ import annotations
 
+import asyncio
 import datetime
 import json
 import logging
@@ -9,12 +10,17 @@ import selectors
 import socket
 import ssl
 import time
-from typing import Any
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import TimeoutError as FuturesTimeoutError
+from typing import TYPE_CHECKING, Any, Callable
 
 import voluptuous as vol
 from homeassistant import config_entries
 from homeassistant.core import callback
 from homeassistant.data_entry_flow import FlowResult
+
+if TYPE_CHECKING:
+    from homeassistant.core import HomeAssistant
 from homeassistant.helpers.selector import (
     ObjectSelector,
     SelectSelector,
@@ -33,6 +39,7 @@ from .const import (
     CONF_BYPASS_REMOTE_CONTROL,
     PROBE_PORT_RANGE, PREFERRED_PROBE_PORTS, LIVENESS_PROBE_TIMEOUT_S,
     PROBE_GET_TIMEOUT_S,
+    RACE_OVERALL_S,
 )
 
 _TEXT = TextSelector(TextSelectorConfig(type=TextSelectorType.TEXT))
@@ -49,6 +56,36 @@ class CannotConnect(Exception):
 
 class InvalidCA(Exception):
     pass
+
+
+class _CertRejected(Exception):
+    """Raised by a probe worker when the device rejects the client cert
+    (DTLS alert: handshake_failure / bad_certificate / unknown_ca / ...).
+    Distinguished from a port TimeoutError so the probe can decide whether a
+    full-cert-rejection warrants a UUID-rotation re-mint fallback."""
+
+
+# DTLS alert phrases that signal the device rejected the client cert. The
+# first four are the spec/design's cert-reject alerts. The extra two widen
+# the net: `access denied` (alert 49) is the device's ACL rejecting the peer
+# UUID — a genuine cert-rejection signal in OCF; `decrypt error` (alert 51)
+# is broader but included so an all-ports crypto failure still triggers the
+# re-mint attempt. The fallback is bounded: it only fires when ALL
+# candidates match, so a spurious match on one port is harmless.
+_CERT_ALERT_MARKERS = (
+    'handshake failure', 'bad certificate', 'unknown ca',
+    'certificate unknown', 'access denied', 'decrypt error',
+)
+
+
+def _is_cert_alert(exc: Exception) -> bool:
+    """True iff a DtlsCoapSession.connect() failure looks like a DTLS cert-reject
+    alert. DtlsCoapSession wraps SSL.Error as ConnectionError("DTLS handshake
+    error: <SSL.Error str>"); we match on the alert phrase in that string.
+    Conservative: anything unrecognized is treated as a generic handshake error
+    (NOT a cert reject), so the UUID-rotation fallback never fires on a guess."""
+    msg = str(exc).lower()
+    return any(m in msg for m in _CERT_ALERT_MARKERS)
 
 
 def _fetch_samsung_uuid() -> str:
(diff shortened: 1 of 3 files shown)


mbillow:
Closing in favor of #273.

PR #218, opened 2026-07-30
feat(airconditioner): filter dust alarm interval on legacy ARTIK051 boards

perseus177:
**Written with Claude (AI).** Flagging that up front — the analysis, code, comments and tests here are AI-authored on my own hardware, and I've reviewed them before sending. Everything quoted below is a measurement off the device rather than an inference; where something is unproven I've said so explicitly.

Two things for the legacy ARTIK051 boards (the `ARTIK051_KRAC_18K` family from #136): a Select for the filter alarm interval, and a settled answer to what `filter_time` on those boards actually measures. Plus a documented dead end on resetting that counter, so the next attempt doesn't repeat my ground.

## What this adds

`FilterAlarmTime_` in `/mode/vs/0`'s `options[]` is the interval the official app offers next to the filter reminder, as a four-way radio: **180 / 300 / 500 / 700 hours**, default 500. This board generation has no `/filter/*` resource at all, so that token is the entire filter-threshold surface.

How it was established:

* Stepping through **all four** radio positions in the app while diffing **all 19 resources (73 fields)** moved that one token per step and nothing else — so the value is the hour count verbatim, not an index or a scaled code.
* A **local write** of `500` to a unit sitting on `700` was accepted and kept, and survived a Home Assistant restart. Not an optimistic read-back: the same snapshot showed `OutdoorTemp` and `AutocleanProgress` moving, so it came from a genuine poll.
* The cloud exposes none of this, so locally is the only place it's available.

It's gated with the counter it belongs to, so boards carrying a real `/filter/airdustfilter/vs/0` keep using `air_filter_threshold`. The gating test **injects** the token into a newer board's dump on purpose: no non-legacy fixture carries one, so asserting on an unmodified dump would pass for the wrong reason (absent token rather than the board-generation gate).

The `options` tuple is static rather than `options_field`, which I know cuts against the usual rule. `options[]` tokens advertise no supported-values list anywhere in this board's dump, so there is nothing to read one from; the four values are the app's own radio, and the descriptor comment says exactly that plus what would need revisiting if a board offers different steps.

## `filter_time` counts up (my previous comment left this open)

The descriptor used to say the direction was unestablished. It counts **up** — running time accumulated since the last filter reset, not time remaining. Three independent things agree:

1. The token rises while the unit runs (observed 191.5 → 192.5 h over one session).
2. `FilterAlarmTime_` in the same blob is the threshold it's measured against.
3. Across two units on one site, `/alarms/vs/0`'s filter entry tracks the counter in the right direction: a live unsuffixed `FilterAlarm` in state `Created` at `FilterTime_5595`, versus the `FilterAlarm_OFF`/`Deleted` placeholder at `FilterTime_1915`.

And the causal half rather than just correlation: when the counter was reset, the alarm **cleared itself** in the same event, so the board recomputes it from counter vs threshold. That also decodes `FilterAlarmTime_`, which was previously listed as unknown.

The key stays `filter_time` — renaming it to something like `filter_time_elapsed` would change every existing unit's `entity_id` and `unique_id` for a wording improvement.

## Filter reset — I could not work out how to do it

**No reset entity is added.** I tried to build one and failed, and I've recorded the attempts in the descriptor rather than quietly leaving a gap. To be clear about what this is: I did not find the mechanism. That is not the same as proving there isn't one, and nothing below rules out something I've simply missed.

The reframing that explains every failure: in Samsung's own model the reset is a **command**, not a value write — capability `custom.dustFilter`, command `resetDustFilter`, no arguments (implemented in a few SmartThings HA forks, not in core). So nothing that writes the counter can work; the board zeroes it itself on receiving a command.

What I tried, all against live hardware, all unsuccessful:

| attempt | result |
|---|---|
| `FilterTime_0` via the single-token `options` merge that works for every other setting on this href | accepted without error, then discarded. Two units in opposite power states to rule out the obvious confound: `5595` → back to `5595` after 69 s (powered off, alarm active), `1925` → back to `1925` after 65 s (actively cooling) |
| full `options[]` read-modify-write with `FilterTime_0` substituted | zero fields changed anywhere |
| write to `/consumable/vs/0`, the board's own filter resource (`items[{name: FilterProgress, state: N}]`) | discarded — and `/oic/res` declares it `oic.if.s`, i.e. read-only, which fits |
| `/actions/vs/0` (`x.com.samsung.da.actions`, `oic.if.a`) — the obvious local command channel | publishes no schema: `GET` returns `{}` on baseline *and* `oic.if.a`, and five POSTs probing the **shape** (empty map, empty string, empty array, invalid value, items shape) all returned `4.00` with an empty body — no echo of accepted field names, unlike the laundry firmware's `Control fail, <...>` |
| `/hass/state/vs/0`, `/hass/command/vs/0` (advertised in `/oic/res`, and `/opt/data/hass.db` is listed by `/file/list/vs/0`) | `4.04` on every interface — unimplemented scaffolding on this firmware |
| `/file/transfer/vs/0`, to read the action vocabulary out of the firmware | serves only `/mnt/usage.db`; selecting another path returns `4.05`/`4.00` |
| `/rm/micomdata/vs/0` (channel toward the MICOM board the physical panel talks to) | stays empty even after successfully enabling remote management |

I deliberately did **not** enumerate guessed action names against `/actions/vs/0` on a live appliance. With no schema and no hint from the error bodies, that's blind enumeration on a channel called "actions", which could hold a factory reset next to the one I want.

What the failures are **not**: a transport, permission or certificate problem. A control write of `rmState` on `/rm/state/vs/0` was accepted (`2.04 Changed`, value held, restored afterwards), and `FilterAlarmTime_` in this very PR goes through the same `options` merge and is kept. So local writes work fine — this one value just isn't driven that way.

Where I'd look next, if anyone wants to pick it up: the action vocabulary for `/actions/vs/0` from an independent source (a firmware image, or a capture of what the cloud sends the device), or the IR route — the physical remote has a filter reset (Options → Filter Reset → SET), and `IRremoteESP8266` decodes this AC family, though [issue #1277](https://github.com/crankyoldgit/IRremoteESP8266/issues/1277)'s dump doesn't include that button.

## Testing

* Fork CI on this branch: **hassfest ✅, pytest ✅**. The HACS job fails 2/9 purely on fork metadata (no repository topics, issues disabled) — not on code; `description` and `brands` checks pass.
* Goldens for both KRAC fixtures regenerated (one key each, `filter_alarm_time`); zero unbound hrefs on both.
* Three new tests: threshold read + single-token write shape, the injected-token gating check, and the alarm/counter/threshold relationship asserted off fixture data.
* Hardware: two `ARTIK051_KRAC_18K` indoor units on one site, firmware `ARTIK051_KRAC_18K_11200825`.

### DIFF
--- a/custom_components/localthings/registry/capabilities/airconditioner.py
+++ b/custom_components/localthings/registry/capabilities/airconditioner.py
@@ -573,15 +573,109 @@ CLIMATE = Capability(
                    unit='°C', icon='mdi:home-thermometer-outline'),
         # Filter time in tenths of an hour: the token read 1710 while the official
         # Samsung app displayed "171 hours 0 minutes" for the filter on the same
-        # unit, and the .0 matching the app's "0 minutes" pins the scale. Whether
-        # it counts up or down is NOT established -- it was seen rising (171.0 ->
-        # 171.5) while the unit ran, which contradicts the app's "remaining"
-        # wording, so the entity is deliberately named neutrally.
+        # unit, and the .0 matching the app's "0 minutes" pins the scale.
+        #
+        # It counts UP -- running time accumulated since the last filter reset,
+        # not time remaining. An earlier revision of this comment called the
+        # direction unestablished; three independent things now settle it. It was
+        # seen rising while the unit ran (171.0 -> 171.5). FilterAlarmTime_ in the
+        # same options[] blob is the threshold it is measured against (500 on
+        # every unit on record). And across two units on one site, the alarm
+        # tracks the counter in the right direction: at FilterTime_5595 the
+        # /alarms/vs/0 filter entry is live (unsuffixed code 'FilterAlarm', state
+        # 'Created'), while at FilterTime_1915 it is still the
+        # 'FilterAlarm_OFF'/'Deleted' placeholder. That also matches what the app
+        # does at 500 hours -- ask the user to clean the filter and reset it.
+        #
+        # The key stays 'filter_time' rather than becoming something like
+        # filter_time_elapsed: renaming it would change every existing unit's
+        # entity_id and unique_id for a wording improvement.
+        #
+        # (The alarm clearing on its own is the causal proof of the direction
+        # above: the board recomputes it the moment the counter crosses the
+        # threshold, rather than the app clearing counter and alarm separately.)
+        #
+        # ---------------------------------------------------------------------
+        # RESETTING THIS COUNTER: NOT SOLVED YET -- no reset entity here, and
+        # the notes below are so the next attempt starts from the evidence
+        # rather than from scratch. I could not work out how to drive the reset
+        # locally; that is not the same as it being impossible, and none of the
+        # results below rule out a mechanism I simply haven't found.
+        #
+        # What the reset actually is: a *command*, not a value write. Samsung
+        # models it as capability `custom.dustFilter`, command
+        # `resetDustFilter`, no arguments (implemented in several SmartThings HA
+        # forks; not in the core integration). That reframes every failed
+        # attempt below -- nothing changes the counter by writing to it,
+        # because the board zeroes it itself on receiving a command.
+        #
+        # Tried, all against a live unit, all failed:
+        #   * FilterTime_0 via the single-token options merge that works for
+        #     every other setting on this href -- accepted with no error, then
+        #     discarded. Two units, opposite power states, to rule out the
+        #     obvious confound: 5595 -> back to 5595 after 69 s (powered off,
+        #     alarm active) and 1925 -> back to 1925 after 65 s (actively
+        #     cooling).
+        #   * A full options[] read-modify-write with FilterTime_0 substituted,
+        #     instead of the single-token merge -- zero fields changed anywhere.
+        #   * A write to /consumable/vs/0, the board's own filter resource
+        #     (items[{name: FilterProgress, state: N}]) -- discarded. /oic/res
+        #     declares that resource oic.if.s, i.e. read-only, which fits.
+        #   * /actions/vs/0 (x.com.samsung.da.actions, oic.if.a) is the obvious
+        #     local command channel, but publishes no schema: GET returns {} on
+        #     baseline and on oic.if.a, and five POSTs probing the *shape*
+        #     (empty map, empty string, empty array, invalid value, items shape)
+        #     all returned 4.00 with an empty body -- no echo of accepted field
+        #     names, unlike the laundry firmware's "Control fail, <...>". I
+        #     deliberately did not enumerate guessed action names against a live
+        #     appliance: an unknown vocabulary on a channel called "actions" can
+        #     hold a factory reset next to the one we want.
+        #   * /hass/state/vs/0 and /hass/command/vs/0 (advertised in /oic/res,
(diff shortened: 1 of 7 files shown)


perseus177:
One more attempt worth recording, since it closes a door I had left open in the description above.

I sent the known command name `resetDustFilter` straight at `/actions/vs/0` — not an enumeration of guessed names, but the one name from Samsung's own `custom.dustFilter` capability — in four envelopes:

```
{"x.com.samsung.da.actions": "resetDustFilter"}                      -> 4.00
{"x.com.samsung.da.actions": ["resetDustFilter"]}                    -> 4.00
{"x.com.samsung.da.actions": [{"x.com.samsung.da.name": "reset..."}]} -> 4.00
{"x.com.samsung.da.action":  "resetDustFilter"}                      -> 4.00
```

`FilterTime_5` untouched throughout, verified from a fresh session afterwards (two further envelopes killed the DTLS session, so those two are inconclusive rather than negative).

The useful part is *which* error: `4.00` is the **same code an empty map gets**. So the rejection is about the payload **structure**, not the name — meaning trying more action names would be aiming at the wrong thing until the shape is known. Whatever `/actions/vs/0` expects, it isn't any of the obvious `x.com.samsung.da.*` envelopes.

So the open question for anyone picking this up is narrower than I framed it: not "which action name resets the filter", but "what shape does `/actions/vs/0` accept at all". A firmware image or a capture of a cloud→device command would answer that; guessing names won't.

perseus177:
Independent corroboration for the "no OCF token for the reset" half of this, which is stronger evidence than my own negative results.

[`bloodtick/Hubitat`'s `replicaSamsungOcfAirCon.groovy`](https://github.com/bloodtick/Hubitat/blob/main/hubiThingsReplica/devices/replicaSamsungOcfAirCon.groovy) is a driver for exactly this AC family, and it clearly knows the OCF escape hatch — it uses `execute` with an href for the display light and the beep, i.e. the same single-token `options` merge this integration uses:

```groovy
def liteOff() { execute("mode/vs/0", ["x.com.samsung.da.options": ["Light_On"]]) }
def beepOff() { execute("mode/vs/0", ["x.com.samsung.da.options": ["Volume_Mute"]]) }
```

Its `resetDustFilter()` does **not** go that way — it calls the cloud capability command. So an author who reaches for `execute`+href for everything outside the standard capabilities did not use one for the filter reset. That reads as "no OCF token exists for it" rather than "I failed to find it", which is a better-founded conclusion than anything my own probing could establish.

(Incidentally that driver also independently confirms the inverted display-light semantics already documented here — `Light_On` gives lite *off*.)

mbillow:
Thanks for all the research. Can you amend these commits to just have you as an author, no agentic coauthor/committers. Trying to keep the log clean.

perseus177:
Done — force-pushed, the `Co-Authored-By` trailer is gone and both author and committer are just me. Tree is bit-identical, only the message changed.

One thing I left in deliberately, so you can decide rather than find it later: the message body still has a single line noting the work was done with AI assistance on my own hardware. That's not an author/committer trailer, but if you'd rather the log carried none of it at all, say so and I'll strip that too — the same note is in the PR description either way, so nothing gets hidden by removing it.

Issue #210, opened 2026-07-30
ASM-KR-TP1-22-ACMB1M / Samsung Air Monitor Plus


jjyn0215:
### Appliance model

ASM-KR-TP1-22-ACMB1M

### What's missing or wrong

Samsung has few air quality sensors - one of them supports COAP-DTLS which is the plus model.
And however, according to the diagnostics, temperature data is missing. I'm not sure why.
Please add this type of device. Thank you!

### Diagnostics file

[config_entry-localthings-01KYREEJWF1Z0H7KKH3NFCCV1B.json](https://github.com/user-attachments/files/30535296/config_entry-localthings-01KYREEJWF1Z0H7KKH3NFCCV1B.json)

mbillow:
Fixed in `v0.18.0`, please upgrade. Let me know if you have an issues. Thanks for the issue, closing this out.

Issue #209, opened 2026-07-30
Incomplete capa for Air conditioner


hwajin-me:
### Appliance model

TP1X_DA-AC-FAC-02011_0000

### What's missing or wrong

The air conditioner has two fans—an upper one and a lower one—but there is no function to activate or deactivate them.

### Diagnostics file

[config_entry-localthings-01KYRV2NAK0P08R24AJ457YM1V.json](https://github.com/user-attachments/files/30534301/config_entry-localthings-01KYRV2NAK0P08R24AJ457YM1V.json)

mbillow:
Dug into your dump. Your board (`TP1X_DA-AC-FAC-02011`, `ModelInfo_16K_BORA_VENT2` in its options) does report `multidevice.numofsubdevice: 2` — real evidence it has a second logical unit — but neither of the two ways this integration knows how to reach a second subdevice found anything: `/device/1` and `/device/2` both came back empty, and `/subdevices/vs/0` is empty too (no UUID list for a second unit to enumerate). `/wind/strength/vs/0` also only exposes one fan-speed control, with nothing that looks like a per-zone field.

That combination (device says "I have 2" but neither known discovery path finds the second one) is a case this integration doesn't have a working answer for yet — it needs a third mechanism I don't have evidence for from this dump alone. Two things that would help:
1. Does the SmartThings app show independent controls for the upper and lower fan? Screenshots would help a lot.
2. Settings → Devices & Services → LocalThings → your AC → gear icon → Debug: write a resource — browsing the href dropdown there might surface something fan-zone-specific that wasn't in the diagnostics dump.

Not something I can build blind — happy to pick this back up once there's more to go on.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #208, opened 2026-07-30
Incomplete capability for Air Dresser


hwajin-me:
### Appliance model

DA_DF_TP1_21_COMMON

### What's missing or wrong

All cycles are represented by code. 

### Diagnostics file

[config_entry-localthings-01KYRTZCKB2FNR4N26GGJR7XRG.json](https://github.com/user-attachments/files/30534310/config_entry-localthings-01KYRTZCKB2FNR4N26GGJR7XRG.json)

mbillow:
Would you be willing to help provide cycle labels? For each cycle, select it in the SmartThings app and then document which code is selected in Home Assistant. If you provide that mapping, we can fix the labels for you!

Issue #207, opened 2026-07-30
`localthings_subpoll` task delays HA startup/shutdown


Sungha81:
## This document was created by Claude.

# `localthings_subpoll` task delays HA startup/shutdown

[logs.txt](https://github.com/user-attachments/files/30531625/logs.txt)

## Summary

`coordinator.py`'s `_run_subpolls` task is registered with `hass.async_create_task`, which ties it into Home Assistant's startup/shutdown task tracking. Since this task is an intentionally long-running background loop (up to ~27s per cycle, per `_SUBPOLL_STEP_S` × 9 slots), HA ends up waiting on it during both shutdown and the next startup, adding real delay to both.

## Evidence

Two log lines from the same root cause, on two different HA lifecycle events:

**On shutdown** (previous run, right before a restart):
```
Task <Task pending name='localthings_subpoll' coro=<LocalThingsCoordinator._run_subpolls() running at /config/custom_components/localthings/coordinator.py:470> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[set.remove()]> was still running after final writes shutdown stage;
Integrations should cancel non-critical tasks when receiving the stop event to prevent delaying shutdown
```

**On the next startup**:
```
Logger: homeassistant.bootstrap
Source: bootstrap.py:1002

Setup timed out for bootstrap waiting on {<Task pending name='localthings_subpoll' ...>, <Task pending name='localthings_subpoll' ...>} - moving forward
```

(Two entries because this instance runs 3 appliances — 2 of them had a subpoll cycle in flight at the timeout point.)

## Root cause

```python
# coordinator.py
self._subpoll_task = self.hass.async_create_task(
    self._run_subpolls(force=sweep_mismatch), name="localthings_subpoll"
)
```

`async_create_task` marks the task as one HA's bootstrap/shutdown sequencing should wait on. `_run_subpolls` is explicitly *not* that kind of task — it's a self-limiting background loop between summary polls, cancelled and recreated every coordinator refresh cycle (see the `if self._subpoll_task is not None: self._subpoll_task.cancel()` at the top of `_async_update_data`). Nothing about it needs to block startup completion or delay shutdown.

## Fix

Swap to `async_create_background_task`, which is HA's supported API for exactly this case (a task the integration owns and manages the lifecycle of, that shouldn't be tracked by HA's own startup/shutdown sequencing):

```python
self._subpoll_task = self.hass.async_create_background_task(
[...]

PR #224, opened 2026-07-31
Issue triage batch: #207, #189, #208, #181/#183, #196, #210 + contributing docs

mbillow:
## Summary

Batch triage across 8 open issues, one commit per fix (see individual commits for full reasoning/evidence on each):

- **#207** — `coordinator.py`'s subpoll task used `hass.async_create_task`, which ties it into HA's startup/shutdown sequencing even though it's a self-limiting loop the coordinator already owns and cancels every refresh cycle. Switched to `hass.async_create_background_task`.
- **#189** — "Device is already configured" adding a second device. Root cause: `_is_placeholder_serial` only caught the `'Nothing(SVC)'` sentinel; the `DA_WM_A51_20_COMMON` laundry board family reports a different one (`'FFFFFFFFFFFFFFF'`, confirmed from the reporter's logs on two different physical units — a washer and a dryer — colliding on the config-entry unique_id). Widened the check to catch any all-same-repeated-hex-digit placeholder.
- **#208** — Air Dresser (`DA_DF_TP1_21_COMMON`) reports `/buzzersound/vs/0`, previously unbound on this board. Bound via the existing shared `laundry.BUZZER_SOUND` capability. (The reporter's actual complaint — course cycles showing as raw codes — is a labelling gap with no code->name mapping anywhere in the dump, not a coverage bug; needs a reporter to identify the codes.)
- **#181 / #183** — Both reported a broken "child lock" toggle. Root cause: `common.KIDS_LOCK_VS_FALLBACK`'s write contract (`'Enable'`) was never once confirmed against any dump in the fixture corpus — every dump reports either `'Ready'` or `'Run'`. #181's reporter confirmed directly: writing the *correct* value (`'Run'`) still 4.05s on real hardware, and the SmartThings app itself has no control for it. Converted to a read-only `binary_sensor` (semantic note: `binary_sensor`'s `lock` device class is inverted from the old switch's polarity — On now means open/unlocked). `KIDS_LOCK_GENERIC` (the OCF-standard `/kidslock/0` boolean) is untouched.
- **#196** — Water purifier's "Favorite hot water temperature" select read its options from `favorite.supportedList` (four fixed presets), but the SmartThings app also lets a user add one custom value to their own display list (`favorite.showList`), and a default set to that custom value rendered as HA's "unknown". Switched the options source to `showList`.
- **#210** — New device family: Samsung Air Monitor Plus, a standalone battery-powered air-quality sensor puck. Routes via both `/oic/d` (`x.com.st.d.airqualitysensor`, confirmed against the real dump) and the `ASM` modelNum board token. Reuses `air_purifier.py`'s existing `/sensors/vs/0` decode for dust/fine_dust/super_fine_dust/odor/clean_level (shared catalog entries), adds CO2. Particulate sensors deliberately get no `pm10`/`pm25`/`pm1` device_class — the three-tier naming doesn't confirm where it maps to the two-tier PM standard, and mislabeling a read-side unit is a standing, not one-shot, kind of wrong. `/dnd/vs/0`'s write contract is an explicitly-flagged educated guess (no idle-vs-active dump exists for it) mirroring the read-side format.
- **#209** — Investigated but not fixed: the AC reports `multidevice.numofsubdevice: 2`, but neither known subdevice-discovery mechanism (`/device/1`/`/device/2`, `/subdevices/vs/0`) finds a second unit. Replied on the issue asking for more data rather than guessing at a third mechanism.

Also:
- Added `CONTRIBUTING.md` / `AGENTS.md` codifying the commit-attribution policy (author/committer = the accountable human, no AI co-author trailers) — generalized from this session's own instructions, not naming anyone specific.
- Relaxed `.claude/skills/adding-device-support/SKILL.md`'s "don't guess" rule: educated guesses at a write contract are fine when backed by real evidence (a supported-values/range field, an idle-vs-active dump diff, a confirmed sibling-board pattern) as long as they're flagged in code and the reporter is asked to confirm on real hardware — a CoAP write against an invalid value is rejected, not acted on, so the worst case is a no-op. Still forbids inventing a write/entity from nothing, and calls out that this doesn't cover read-side device_class/unit guesses (those fail silently, not with a rejection).

## Test plan
- [x] `pytest tests/ -q` — full suite green (1048 tests) as of the last commit.
- [x] New fixtures/goldens/capability tests added for #208 and #210.
- [ ] Independent Opus review pass in progress, focused on the #181/#183 semantic-polarity flip, the #189 serial heuristic, and #210's new registry — findings to follow as a PR comment.
- [ ] Reporters on #181, #183, #209 asked to confirm specific next steps on real hardware (comments posted on each issue).

### DIFF
--- a/.claude/skills/adding-device-support/SKILL.md
+++ b/.claude/skills/adding-device-support/SKILL.md
@@ -270,23 +270,49 @@ sub-polled between summary polls. Pick descriptor types from `entities.py`
 (`SensorDesc`, `SelectDesc`, `SwitchDesc`, `NumberDesc`, `BinarySensorDesc`,
 `TimeDesc`, `ButtonDesc`) — the class selects the HA platform.
 
-**Don't guess.** If a field's meaning or write contract is unclear from the dump
-(opaque encoded blobs, no supported-values list), leave it unbound so it surfaces
-as a gap for a human, or ignore it with a documented reason — never invent an
-entity on a hunch (`ignored.py`'s rule).
+**Educated guesses are fine — flag them, don't hide them.** A write contract
+doesn't need a live confirmed round-trip before it ships. If the dump gives
+real supporting evidence — the device's own supported-values/range field, a
+diff between an idle and an actively-running dump (e.g. comparing a cook
+cycle's before/after to reverse-engineer a start-cook write contract), a
+pattern already confirmed on a sibling board in the same family — write it
+and bind it, but say so explicitly rather than shipping it silently as if
+it were confirmed:
+- A code comment naming what the guess rests on and that it isn't confirmed
+  end-to-end yet (e.g. "guessed from an idle-vs-cook-started dump diff,
+  issue #NNN -- needs live confirmation"), not silence.
+- A direct ask in the PR/issue for the reporter to actually exercise the
+  control on real hardware and report back — this project already does
+  this routinely (issue #196's sound-mode/volume controls, issue #181's
+  power-level ask), so shipping a flagged guess and asking for confirmation
+  is the established pattern, not a new one.
 
-This is a rule about **writes and entities**, not about reading. A speculative
-`GET` of an href a dump doesn't contain is fine and the codebase already relies
-on it: `read_identity` reads `/oic/p`, `/oic/d` and `/oic/res`, and
-`subdevices.enumerate_subdevices` probes `/device/<n>`, `/<uuid>/device/0` and
-`/multidevice/vs/0` on every device — and, when a prefixed candidate's own
-`/<uuid>/device/0` doesn't answer (issue #205: not guaranteed even on the
-board this pattern was built against), every href the master itself
-answered this cycle, individually under that UUID's prefix (see §11). A
-RETRIEVE is non-mutating and a 4.04 is tolerated everywhere in that path, so
-the cost of a wrong guess is one wasted round trip. Guessing a *write*
-against live hardware is the thing this rule forbids — as is materializing
-an entity from a field you can't explain.
+Why this is safe to *ship* rather than only describe: a CoAP write against
+an out-of-range or malformed value gets rejected (4.xx), not acted on — the
+worst case for a wrong *value* is a no-op, not a damaged or misbehaving
+appliance. That margin only covers the value, though, not the semantics:
+bind a guessed write to the device's own reported range/supported-list
+rather than inventing bounds, and don't guess a unit the dump gives no way
+to cross-check (temperature scale, minutes vs. seconds) — being
+syntactically valid but semantically backwards is exactly the case a
+rejection won't catch.
+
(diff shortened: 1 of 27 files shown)


mbillow:
1. common.py:326 — correctness — CONFIRMED
child_lock polarity now inverted between the two kids-lock capabilities. KIDS_LOCK_GENERIC (/kidslock/0) is untouched — still on=locked. KIDS_LOCK_VS_FALLBACK (/kidslock/vs/0) is now on=unlocked. 10 boards (washers, dryers, air dressers) bind the generic cap; 11 (dishwasher, ranges/oven, microwaves, air purifier) bind the fallback. Same entity name, opposite meaning depending on which a device has. Also: device_class='lock' was never valid on SwitchDesc (HA’s switch device classes are only {outlet, switch}), so the “inverted from the old switch convention” premise in the commit/comment is itself resting on a device_class HA never honored there.

2. test_water_purifier_capabilities.py:372 — test-coverage — CONFIRMED
The “case that exercises the #196 fix” asserts something that passes identically against the old field too (the fixture’s current default, '85', is in both lists). Nothing in the suite actually reproduces the reported bug. Also: showList could plausibly be a subset of supportedList (if a user deselects a fixed preset), which would reintroduce the same “unknown” bug from the other direction — reading the union of both fields would be strictly safer.

3. air_monitor.py:55 — correctness — CONFIRMED
dust/fine_dust/super_fine_dust get no state_class, so they’re excluded from HA long-term statistics on a device whose whole purpose is trending air quality — while co2 in the same capability does get state_class='measurement'. The stated reason (“matching air_purifier’s existing descriptors”) is consistency with a prior omission, not a reason of its own.

4. by_type/air_dresser.py:1 — commit-message-accuracy — CONFIRMED
The #208 commit message says the resource shape is {setBuzzerSound, supportedFinishSound} — the real field is supportedBuzzerSound; supportedFinishSound is absent (the branch’s own test asserts this). The module docstring added in the same commit gets it right; only the git commit message is wrong.

5. SKILL.md:295 — documentation — CONFIRMED
The relaxed “educated guesses” section deleted the old read/write distinction and never states a rule about read-side device_class/unit guesses — yet air_monitor.py‘s docstring justifies its biggest judgment call by citing “the skill’s own carve-out” for exactly that, which doesn’t actually exist in the text.

6. by_type/__init__.py:117 — style — CONFIRMED
Trailing whitespace on the new 'ASM' line (git diff --check flags it). Also the new dnd catalog entry breaks alphabetical order in both en.json/nl.json’s switch section.

7. air_monitor.py:40 — simplification — CONFIRMED
_AIR_QUALITY_SENSORS is copy-pasted byte-for-byte from air_purifier.py despite the docstring claiming it’s “reused.” Should be imported from one place so the two don’t silently drift. common.sensor_item_value‘s own docstring also wasn’t updated to mention this third caller.

Issue #172, opened 2026-07-28
Incomplete capability coverage for Samsung Microwave (TP2X_DA-KS-MICROWAVE)


firstof9:
### What's missing or wrong

Missing Light high/low/off
Missing Fan speed high/med/low/off

### Diagnostics file

[config_entry-localthings-01KYN1XD61MW50T0YXE8DBWBST.json](https://github.com/user-attachments/files/30476413/config_entry-localthings-01KYN1XD61MW50T0YXE8DBWBST.json)

PR #225, opened 2026-07-31
Route Samsung Bespoke Qooker to microwave registry

atc722:
## What changed

- Resolve distinctive local-resource signatures before OIC type and model metadata, reusing the existing `/oven/vs/0` + `MicroWave*` microwave matcher.
- Reject malformed scalar `supportedModes` values so resource-first routing cannot turn substring membership into a false positive.
- Add a scrubbed real-device fixture for Samsung Bespoke Qooker MW7500A (`TP2X_DA-KS-OVEN-01011`), a golden state-key regression, and routing/capability coverage.

## Why

Qooker reports `oic.d.oven` and an `OVEN` board token, so the old metadata-first resolver selected the oven registry before reaching its already-supported microwave resource signature. Its verified local API is microwave-shaped: `/mode/vs/0` includes `MicroWave`, `/oven/vs/0` exposes the microwave cavity surface, and every exposed href binds cleanly to the existing microwave registry.

The live capability surface is the registry's most relevant signal. Existing resource matchers already require multiple independent details; when none matches, OIC type still takes precedence over model-string parsing.

## User impact

Qooker is represented with microwave semantics (`cooking_mode`, `power_level`, and the 40–200 °C setpoint) instead of misleading oven entities such as `oven_mode` and `oven_state`.

## Safety and regression coverage

- The fixture corpus locks resource/model disagreements to the one verified Qooker case (`microwave` vs `oven`), guarding against future greedy matchers.
- Negative tests cover microwave mode without an oven cavity and malformed scalar `supportedModes`.
- Existing microwaves that omit `/information/vs/0` continue to use the same resource signature.

## Validation

- GET-only real-device verification: `/device/0` and `/oic/d` both returned `2.05 Content`.
- Modified resolver against that live response: `microwave`, zero unbound hrefs, expected microwave entity keys, no `oven_mode`.
- `pytest tests/ -q`: 1076 passed.
- `compileall`, JSON parsing, and `git diff --check` passed.
- The fixture was scrubbed and checked for UUID, MAC, IPv4, email, and PEM material; serial/device identifiers remain redacted.

## Related work

- #140 split microwave semantics from the oven registry.
- #174 and #172 established the existing resource-signature fallback for microwaves without usable model metadata.

## Scope

This PR adds no capability or write function. Device writes, cooking start, and mode changes were not exercised or claimed; the live verification was read-only.

### DIFF
--- a/custom_components/localthings/registry/by_type/__init__.py
+++ b/custom_components/localthings/registry/by_type/__init__.py
@@ -296,13 +296,12 @@ def for_device_by_model(model_num: str, description: str) -> Optional[DeviceRegi
 def for_device_by_resources(resources: dict[str, dict]) -> Optional[DeviceRegistry]:
     """Detect a device family from a distinctive local-resource signature.
 
-    This is the highest-confidence routing signal: the live capability surface
-    wins over OIC/model metadata when they disagree (Qooker's ``oic.d.oven``
-    declaration is the first verified case). It also types boards that ship no
-    ``/information/vs/0`` at all, leaving `for_device_by_model` nothing to
-    read. Some newer cooktops are the original case: their mode resource still
-    identifies them, carrying a DeviceType option and multiple per-burner
-    OperationState options.
+    This runs first as an override path for non-standard devices, not because
+    resource signatures are inherently more trustworthy than OIC/model
+    metadata. It also types boards that ship no ``/information/vs/0`` at all,
+    leaving `for_device_by_model` nothing to read. Some newer cooktops are the
+    original case: their mode resource still identifies them, carrying a
+    DeviceType option and multiple per-burner OperationState options.
 
     Require two independent shapes for every signature here, never one, so
     putting this ahead of OIC/model metadata cannot let a common resource
--- a/custom_components/localthings/registry/by_type/__init__.py
+++ b/custom_components/localthings/registry/by_type/__init__.py
@@ -296,13 +296,17 @@ def for_device_by_model(model_num: str, description: str) -> Optional[DeviceRegi
 def for_device_by_resources(resources: dict[str, dict]) -> Optional[DeviceRegistry]:
     """Detect a device family from a distinctive local-resource signature.
 
-    For boards that ship no ``/information/vs/0`` at all, leaving
-    `for_device_by_model` nothing to read. Some newer cooktops are the
-    original case: their mode resource still identifies them, carrying a
-    DeviceType option and multiple per-burner OperationState options.
+    This is the highest-confidence routing signal: the live capability surface
+    wins over OIC/model metadata when they disagree (Qooker's ``oic.d.oven``
+    declaration is the first verified case). It also types boards that ship no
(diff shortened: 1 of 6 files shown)


custom_components/localthings/registry/by_type/__init__.py
```diff
@@ -325,6 +325,27 @@ def for_device_by_resources(resources: dict[str, dict]) -> Optional[DeviceRegist
...
+    fallback in ``for_device_by_resources``.
+    """
+    mode = resources.get('/mode/vs/0', {})
+    cavity = resources.get('/oven/vs/0', {})
+    supported_modes = mode.get('x.com.samsung.da.supportedModes') or ()
+    return (
+        isinstance(supported_modes, (list, tuple))
+        and any(
+            value in supported_modes
+            for value in ('MicroWave', 'MicroWaveGrill', 'MicroWaveConvection')
```
mbillow: This logic already exists almost exactly: https://github.com/mbillow/localthings/blob/e55c55c55e97fdf07fdfd9cf853d022afc0feed8/custom_components/localthings/registry/by_type/__init__.py#L314-L320

custom_components/localthings/registry/by_type/__init__.py
```diff
@@ -350,11 +379,25 @@ def resolve(
...
     """
     info = resources.get('/information/vs/0', {})
-    return (
+    metadata_registry = (
         for_device_by_oic_type(device_types)
         or for_device_by_model(
             info.get('x.com.samsung.da.modelNum', ''),
             info.get('x.com.samsung.da.description', ''),
         )
-        or for_device_by_resources(resources)
```
mbillow, line 374: We should just check `for_device_by_resources` -> `for_device_by_oic_type` -> `for_device_by_model`. Just need to make sure that the overrides in `for_device_by_resources` aren't too greedy otherwise they will be quite disruptive.

custom_components/localthings/registry/by_type/__init__.py
```diff
@@ -296,13 +296,17 @@ def for_device_by_model(model_num: str, description: str) -> Optional[DeviceRegi
 def for_device_by_resources(resources: dict[str, dict]) -> Optional[DeviceRegistry]:
     """Detect a device family from a distinctive local-resource signature.
 
-    For boards that ship no ``/information/vs/0`` at all, leaving
-    `for_device_by_model` nothing to read. Some newer cooktops are the
-    original case: their mode resource still identifies them, carrying a
-    DeviceType option and multiple per-burner OperationState options.
+    This is the highest-confidence routing signal: the live capability surface
```
mbillow: I don’t love this comment. It’s kind of misleading. If anything, it’s the lowest confidence and most obtuse signal. We allow it to be first because it’s meant to support the most non-standard devices that need overrides.

atc722: Updated the docstring to describe resource matching as an override path for non-standard devices, rather than a higher-confidence signal. Thanks for clarifying.

PR #250, opened 2026-08-02
Fix washer course enum display

galaxysj:
## Summary

- add the confirmed Samsung Table_02 washer course mappings, including `69` as AI Wash
- decode device-provided personal course names from their byte-counted UTF-8 payload
- show unrecognized washer enum bytes as `Unknown (0xNN)` without guessing course names
- normalize the current state and option list through the same display path while preserving raw values on write

## Root cause

The device reports the selected course as a string token such as `Course_69` and its selectable courses as a packed string such as `EditCourseList_696F...`. The integration correctly split those into one-byte enum codes, but its Table_02 translation catalog did not contain this washer's newer enum family. Untranslated values then reached the generic camel-case formatter, which rendered values such as `6F` as `6 F`.

The diagnostics do not provide names for standard course codes, so only owner-confirmed mappings are added. Personal courses are handled separately using the names carried by the device itself.

## Validation

- `pytest -q`: 1077 passed
- `git diff --check`


galaxysj:
Superseded by #251, which targets galaxysj:main and includes the dishwasher course labels.

PR #251, opened 2026-08-02
Fix appliance course labels and AC setup timeouts

galaxysj:
## Summary

- add confirmed Samsung Table_02 washer course mappings, including `69` as AI Wash
- add confirmed dishwasher course mappings for `82`, `8A`, `A7`, `A8`, `8C`, and `88`
- localize the new washer and dishwasher course labels in the shipped English, Czech, and Dutch catalogs
- decode device-provided personal washer course names from byte-counted UTF-8 payloads
- show unrecognized washer enum bytes as `Unknown (0xNN)` without guessing course names
- normalize current state and option lists through the same display path while preserving raw values on write
- bound first-setup subdevice enumeration so silent AC endpoints cannot stall config-entry setup for minutes

## Root cause

Samsung reports selected courses as string tokens such as `Course_69` and selectable courses as packed strings such as `EditCourseList_696F...`. The integration split these into one-byte enum codes, but its translation catalog lacked this device family enum set. Untranslated values reached the generic formatter and rendered values such as `6F` as `6 F`.

The diagnostics do not provide names for standard course codes, so only owner-confirmed mappings are catalogued. Personal courses use names carried by the device itself. Course labels remain in Home Assistant translation catalogs, not Python mappings, so the UI resolves them using the active language.

Some UUID-prefixed air conditioners do not expose a `/<uuid>/device/0` Collection and silently drop unknown prefixed Property requests instead of returning 4.04. The old fallback scanned roughly 30 hrefs sequentially with a 10-second timeout for each one, allowing first setup to take several minutes before Home Assistant cancelled it. Enumeration now has one 15-second network-wait budget, short Property timeouts, registry-driven primary-resource ordering, and preserves useful responses received before the deadline.

## Validation

- `.venv/bin/python -m pytest -q`: 1082 passed
- `git diff --check`


custom_components/localthings/registry/capabilities/laundry.py
```diff
@@ -276,12 +277,56 @@ def cycle_write(p, rep, href=None):
...
+
+
+def washer_cycle_fallback(value, resources):
+    """Label an untranslated washer course without guessing its meaning."""
+    if not isinstance(value, str):
+        return None
+    if label := personal_course_labels(resources).get(value.upper()):
+        return label
+    if len(value) == 2 and all(char in string.hexdigits for char in value):
+        return f'Unknown (0x{value.upper()})'
```
mbillow, line 320: I’m kind of partial to just leaving the I know hex codes in the list as is. This change adds hard coded, untranslatable English into the component that we’ve put a lot of effort to make translatable.

custom_components/localthings/translations/cs.json
```diff
@@ -280,7 +280,13 @@
           "07": "Předoplach",
           "8d": "Hrnce a pánve",
           "8e": "Plast",
-          "8f": "Dětské potřeby"
+          "8f": "Dětské potřeby",
+          "82": "Auto",
+          "8a": "Normal",
+          "a7": "Heavy",
```
mbillow: Let’s not bake English into other languages translation files. Most (if not all) of these are general terms, not brand-specific titles. There should at least be an attempt at translation.

galaxysj:
Understood. Making changes soon.

mbillow:
Closing in favor of #333, which squash-merges this together with #275 and #312 (the same washer/dishwasher course-code catalog was touched by all three, and #312's German translations needed those codes to stay complete). Your commits and course-code confirmations are carried over — thanks for the detailed writeup!

---

PR #247, opened 2026-08-01
Add Samsung EHS (Eco Heating System) heat pump support

pookey:
## Summary

- Adds a `by_type/ehs.py` registry + `capabilities/ehs.py` for the `TP1X_DA_AC_EHS` board family (Samsung EHS air-to-water heat pumps), routed via a new `EHS` board token in `_BOARD_TOKEN_TO_KEY`.
- An EHS unit runs two independent loops off one outdoor unit: `zone1` space heating/cooling (`/mode/vs/0`, `/power/vs/0`, `/temperatures/indoor/vs/0`) and `dhw` domestic hot water (`/mode/dhw/vs/0`, `/power/dhw/vs/0`, `/temperatures/dhw/vs/0`). Each loop gets a power switch, mode select, and current/target temperature sensor+number — the same switch/select/number/sensor split `dehumidifier.py` uses for its single loop, since this integration has no `water_heater` platform and the AC family's composite `climate` entity is hardcoded to a different resource shape (`airconditioner.py`'s `HREF_MODE`/`HREF_TEMP*` OCF pattern, not this family's `/mode/*/vs/0` / `/temperatures/*/vs/0` shapes).
- Also adds an `away_mode` switch (`/option/outgoing/vs/0`) and reuses `airconditioner.MUTE_ONCE` (`/option/muteonce/vs/0` — identical field shape on this family).
- Remaining resources (opaque hex-encoded cycle/factory-setting/schedule blobs, static installation info, a cosmetic front-panel display toggle) are declared as inert coverage capabilities per the "don't guess" convention, rather than modeled or reverse-engineered.

## Verification

Verified against a real `TP1X_DA_AC_EHS_01001_0000` diagnostics dump (firmware `AEH-WW-TP1-22-AE6000_17260402`, TizenRT 3.1 / DAWIT 2.0), and against the live device after installing this branch — all new entities (`zone_power`, `zone_mode`, `zone_temperature`, `zone_target_temperature`, `dhw_power`, `dhw_mode`, `dhw_temperature`, `dhw_target_temperature`, `away_mode`) came up reporting correct live values, with zero unbound hrefs.

- `tests/fixtures/ehs_device.json` + `tests/fixtures/golden/ehs.json`: golden-regression fixture built from the same (redacted) diagnostics dump.
- `tests/test_ehs_capabilities.py`: registry resolution, no-unbound-hrefs, state reads, and write contracts for every new entity.
[...]

PR #252, opened 2026-08-02
Add Samsung EHS (Eco Heating System) heat pump support

pookey:
Supersedes #247, which I closed and reopened as this PR — same commits, no other changes. I'm restructuring my fork so `main` stays free for local HA testing while I develop several follow-up features in parallel branches, and a fork PR can't be retargeted to a different branch once opened, so a fresh PR from a dedicated branch was the only way to free it up.

## Diagnostics dump

[config_entry-localthings-01KYZKTCG33MCPCCD4JMM4AFYT.json](https://github.com/user-attachments/files/30628493/config_entry-localthings-01KYZKTCG33MCPCCD4JMM4AFYT.json)

## Summary

- Adds a `by_type/ehs.py` registry + `capabilities/ehs.py` for the `TP1X_DA_AC_EHS` board family (Samsung EHS air-to-water heat pumps), routed via a new `EHS` board token in `_BOARD_TOKEN_TO_KEY`.
- An EHS unit runs two independent loops off one outdoor unit: `zone1` space heating/cooling (`/mode/vs/0`, `/power/vs/0`, `/temperatures/indoor/vs/0`) and `dhw` domestic hot water (`/mode/dhw/vs/0`, `/power/dhw/vs/0`, `/temperatures/dhw/vs/0`). Each loop gets a power switch, mode select, and current/target temperature sensor+number — the same switch/select/number/sensor split `dehumidifier.py` uses for its single loop, since this integration has no `water_heater` platform and the AC family's composite `climate` entity is hardcoded to a different resource shape (`airconditioner.py`'s `HREF_MODE`/`HREF_TEMP*` OCF pattern, not this family's `/mode/*/vs/0` / `/temperatures/*/vs/0` shapes).
- Also adds an `away_mode` switch (`/option/outgoing/vs/0`) and reuses `airconditioner.MUTE_ONCE` (`/option/muteonce/vs/0` — identical field shape on this family).
- Remaining resources (opaque hex-encoded cycle/factory-setting/schedule blobs, static installation info, a cosmetic front-panel display toggle) are declared as inert coverage capabilities per the "don't guess" convention, rather than modeled or reverse-engineered.

## Verification

Verified against a real `TP1X_DA_AC_EHS_01001_0000` diagnostics dump (firmware `AEH-WW-TP1-22-AE6000_17260402`, TizenRT 3.1 / DAWIT 2.0), and against the live device after installing this branch — all new entities (`zone_power`, `zone_mode`, `zone_temperature`, `zone_target_temperature`, `dhw_power`, `dhw_mode`, `dhw_temperature`, `dhw_target_temperature`, `away_mode`) came up reporting correct live values, with zero unbound hrefs.

- `tests/fixtures/ehs_device.json` + `tests/fixtures/golden/ehs.json`: golden-regression fixture built from the same (redacted) diagnostics dump.
- `tests/test_ehs_capabilities.py`: registry resolution, no-unbound-hrefs, state reads, and write contracts for every new entity.
- `tests/test_by_type.py`: board-token resolution test (`EHS` -> `ehs`), following the existing `test_dehumidifier_via_dhm_token` pattern.
- `translations/en.json` + `cs.json` + `nl.json`: catalog entries for every new entity key, all translated.

## Test plan

- [x] `pytest tests/` passes locally
- [x] Installed on a live HA instance against a real EHS unit; all new entities report correct values and accept writes


mbillow:
We’ll tackle this in #253. Closing.

Issue #177, opened 2026-07-28
Samsung 2-in-1 Air Conditioner - Only Master Unit Is Discovered, Sub Device (Wall-mounted Unit) Is Missing


jhkwon19:
### Appliance model

AF17A6478WZN, AR06A9170HNQ

### What's missing or wrong

I upgraded to LocalThings v0.15.0, and the original issue has been fixed. Thank you for the update.

The floor-standing air conditioner is now correctly recognized as a climate entity and can be controlled without any problems.

However, I have another issue regarding Samsung's 2-in-1 air conditioner system.

This system consists of:

1 Floor-standing indoor unit (Master)
1 Wall-mounted indoor unit (Sub)

When I registered the system in the official SmartThings app, I only registered the floor-standing unit. SmartThings automatically detected the wall-mounted indoor unit as well, and both indoor units appear as separate devices in SmartThings, even though they share a single IP address.

In LocalThings, however, only the master (floor-standing) unit is discovered when adding the device by IP address.

The wall-mounted indoor unit is not discovered and no separate climate entity is created for it.

From the diagnostics, I noticed that the device exposes a resource, so it appears that the air conditioner reports sub-device information./subdevices/vs/0

Could LocalThings also enumerate these sub-devices and create a separate config entry or climate entity for the wall-mounted indoor unit?

If this requires a different implementation for Samsung 2-in-1 systems, I'd be happy to provide additional diagnostics or test builds.

Thank you!

### Diagnostics file

[config_entry-localthings-01KYNG1HANTMCYQHFX5NVVGPTB.json](https://github.com/user-attachments/files/30482914/config_entry-localthings-01KYNG1HANTMCYQHFX5NVVGPTB.json)

mbillow:
I have a hypothesis here that some devices expose multiple devices on different ports. I think #19 might be another case of this. I just don't have one to poke at. Can you send me the output of the `nmap` command in the README?

jhkwon19:
> 여기서 제가 가설을 세운 것은 일부 기기들이 서로 다른 포트에 여러 기기를 노출시키고 있다는 것입니다. #[19](https://github.com/mbillow/localthings/issues/19)도 이런 사례일 수 있다고 생각해요. 그냥 건드릴 만한 게 없을 뿐이에요. README에 있는 명령어 출력을 보내주실 수 있나요?`nmap`

[...]

Issue #160, opened 2026-07-28
AC humidity sensor: fivepercentHumidity of 0 now reads as unknown instead of 0%


mbillow:
### Where

`custom_components/localthings/registry/capabilities/airconditioner.py`, `_humidity()` (introduced in #146):

```python
def _humidity(rep):
    for field in ('x.com.samsung.da.fivepercentHumidity',
                  'x.com.samsung.da.humidity'):
        if field in rep:
            value = _num(rep[field])
            return value if value else None
    return None
```

### What's wrong

`value if value else None` collapses a genuine `0` reading to `None` ("not measuring") for **both** fields it checks. That collapse was added because ARTIK051 boards' plain `x.com.samsung.da.humidity` field only populates while "Air monitoring" is briefly on and zeroes out afterward — a real hardware quirk specific to that field.

But the same collapse now also applies to `x.com.samsung.da.fivepercentHumidity`, which every other AC model in the registry already relies on and which has never been documented as getting stuck at zero. Before #146, this field was read with a plain `value_fn=_num`, so `fivepercentHumidity: "0"` rendered as 0% humidity. After #146, the same input renders as `unknown`.

### Impact

Any existing (non-ARTIK051-legacy) AC board that reports a genuine `fivepercentHumidity` of `0` will silently switch from showing "0%" to showing "unknown". Low real-world frequency (near-zero indoor RH is rare), but it's a regression for previously-working boards, not just new coverage for the KRAC board this PR added.

### Suggested fix

Only apply the zero-as-"not measuring" carve-out to the `x.com.samsung.da.humidity` fallback field, and let `fivepercentHumidity` pass `0` through as a real reading, e.g.:

```python
def _humidity(rep):
    if 'x.com.samsung.da.fivepercentHumidity' in rep:
        return _num(rep['x.com.samsung.da.fivepercentHumidity'])
    if 'x.com.samsung.da.humidity' in rep:
        value = _num(rep['x.com.samsung.da.humidity'])
        return value if value else None
    return None
```

(Or confirm via a dump that `fivepercentHumidity` has the same stuck-at-zero behavior before keeping the collapse for both fields.)

Found during review of #146.

mbillow:
Fixed in **v0.15.0** — `fivepercentHumidity` now passes a genuine `0` through as 0% instead of collapsing it to unknown; the zero-as-"not measuring" carve-out only applies to the `x.com.samsung.da.humidity` fallback field now, as suggested.

---
[...]

Issue #155, opened 2026-07-28
Air conditioner fan mode only shows "Auto" (AR07C9150HZN)


leekh406-zeppelin:
### Appliance model

AR07C9150HZN(TP1X_DA-AC-RAC-01001_0000)

### What's missing or wrong

Symptom: In the HA UI, the fan mode option only shows "Auto" — no other options are available.
Other functions (power, temperature, mode Cool/Off) are confirmed working normally.

Model: AR07C9150HZN (TP1X_DA-AC-RAC-01001_0000)

Looking at the diagnostics, the device's /wind/strength/vs/0 resource actually reports 6 supported modes:

supportedModes: ["0","31","32","33","34","35"]
modesName: ["Auto","1","2","3","4","MAX"]

It seems the fan_mode mapping for airconditioner.py isn't reading these into HA's fan mode options yet.

Diagnostics file attached.

[config_entry-localthings-01KYK4MEXP346N993QDDM4J1F6.json](https://github.com/user-attachments/files/30457221/config_entry-localthings-01KYK4MEXP346N993QDDM4J1F6.json)

mbillow:
Thank you for this report — you nailed the root cause yourself. The fan-speed mapping wasn't reading the device's own `modesName`/`supportedModes` and was silently falling back to a fixed map that didn't cover this board's non-standard codes. Fixed in **v0.15.0**: fan speed now reads the device's own labels live, so all six of your unit's speeds (Auto, 1-4, MAX) should show up correctly.

Please upgrade and confirm — thanks again for pointing straight at the resource!

---
_Generated by [Claude Code](https://claude.ai/code)_

leekh406-zeppelin:
Just updated to v0.15.0 and confirmed the fan speed mapping now works perfectly — all 6 modes (Auto, 1~4, Max) show up correctly. Thanks so much for the quick fix!

<img width="589" height="884" alt="Image" src="https://github.com/user-attachments/assets/6159f37e-9b48-478e-86b0-e4bde6c81a33" />

Issue #93, opened 2026-07-27
Air conditioner: several device-reported modes dropped by climate.py maps (Nano/LongWind/NanoSleep, Fan/AIComfort, Left_And_Right)


strow82:
### Appliance model

A-CAWW-TP2-20-COMMON (system air conditioner, mnid 0AJT / setupid 023) Software 02274A260512, Firmware 02419A25051901 Detected via the '-CAWW-' branch in for_device_by_model() (issue #52)

### What's missing or wrong

The `airconditioner` registry binds correctly and the climate entity works well — thanks for
adding the `-CAWW-` routing in #52. But several values the device advertises in
`supportedModes` never reach the entity, because `climate.py`'s device↔HA maps don't contain
them and `_read_modes()` silently drops unmapped codes.

Comparing the local entity against the official SmartThings cloud integration for the same
physical unit:

**Preset (`/mode/convenient/vs/0`)**

Device reports: `["Off", "Nano", "LongWind", "Speed", "Sleep", "NanoSleep"]`

| Device code | In `_DEVICE_TO_PRESET`? | SmartThings cloud exposes |
|---|---|---|
| Off | yes → `none` | `none` |
| Nano | **missing** | `wind_free` |
| LongWind | **missing** | `long_wind` |
| Speed | yes → `speed` | `boost` |
| Sleep | yes → `sleep` | `sleep` |
| NanoSleep | **missing** | `wind_free_sleep` |

`Nano` is Samsung's WindFree — arguably the single most-used feature on these units, so its
absence is the most visible gap. Note also that `Speed` currently maps to a custom `speed`
preset while the cloud integration calls the same device code `boost`; worth aligning so
automations port between the two integrations unchanged.

**HVAC mode (`/mode/vs/0`)**

Device reports: `["Auto", "Cool", "Dry", "Fan", "AIComfort"]`

`_DEVICE_TO_HVAC` has `Wind` → `FAN_ONLY`, but this firmware sends `Fan`, not `Wind`, so
fan-only is unreachable. `AIComfort` is unmapped entirely (cloud shows it; `HVACMode.AUTO`
is free since `Auto` already takes `HEAT_COOL`).

**Swing (`/wind/direction/vs/0`)**

Device reports: `["Up_And_Low", "Fix", "Left_And_Right", "All"]`

`Left_And_Right` is unmapped; `horizontal` is the obvious HA constant and is what the cloud
integration uses.

**Net effect**

[...]

PR #253, opened 2026-08-02
Add a water_heater platform for the EHS DHW loop

pookey:
Follow-up to #252 (review feedback from mbillow on the original #247: "Having water heaters automatically leverage the right platform would be pretty cool!"). Stacked on #252 — this PR's diff includes its 3 commits until that one merges, then it'll shrink to just the commit below.

## Summary

Replaces the DHW loop's switch/select/number/sensor split with a proper composite `water_heater` entity, following the same primary-resource-plus-sibling-reads pattern `climate.py` already uses for the AC:

- `custom_components/localthings/water_heater.py` (new): `LocalThingsWaterHeater`, binding the primary `/mode/dhw/vs/0` capability and reading sibling `/power/dhw/vs/0` / `/temperatures/dhw/vs/0` resources off the coordinator snapshot.
- Operation modes (Eco/Std/Force/Power) map onto HA's own standard water_heater states (`STATE_ECO`/`STATE_HEAT_PUMP`/`STATE_HIGH_DEMAND`/`STATE_PERFORMANCE`) — the same mapping HA core's own `smartthings` integration uses for this exact Samsung capability over the cloud API, so no custom state translations were needed.
- `registry/entities.py`: new `WaterHeaterDesc` (mirrors `ClimateDesc`/`FanDesc`).
- `registry/capabilities/ehs.py`: replaced `DHW_POWER`/`DHW_MODE`/`DHW_TEMPERATURE` with a composite `DHW` capability plus coverage-only caps for the sibling hrefs.
- Translations: removed the now-obsolete `dhw_power`/`dhw_mode`/`dhw_temperature`/`dhw_target_temperature` catalog entries from en/cs/nl.

`zone1` (space heating/cooling) intentionally stays switch/select/number/sensor — it's a leaving-water-temperature setpoint, not a thermostat, so neither `climate` nor `water_heater` fits it better.

## Test plan

- [x] `pytest tests/` passes locally (1100 passed)
- [ ] Installed on a live HA instance against a real EHS unit

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ custom_components/localthings/
   diagnostics.py           Redacted diagnostics download (device state + coverage metadata)
   const.py                 Domain, config keys, probe ports
   entity.py                Base entity wiring capability registry -> HA entity
-  sensor.py / binary_sensor.py / switch.py / number.py / select.py / button.py / time.py / fan.py / climate.py
+  sensor.py / binary_sensor.py / switch.py / number.py / select.py / button.py / time.py / fan.py / climate.py / water_heater.py
                             One module per HA platform
   catalog.py               Reads the shipped translation catalog (which keys/states exist)
   translations/            Config-flow copy + entity name/state translations, one file per
--- a/custom_components/localthings/const.py
+++ b/custom_components/localthings/const.py
@@ -2,7 +2,7 @@ DOMAIN = "localthings"
 
 PLATFORMS = [
     "sensor", "binary_sensor", "switch", "number", "select", "button",
-    "time", "climate", "fan",
+    "time", "climate", "fan", "water_heater",
 ]
 
 CONF_HOST         = "host"
--- a/custom_components/localthings/registry/by_type/__init__.py
+++ b/custom_components/localthings/registry/by_type/__init__.py
@@ -5,7 +5,7 @@ from typing import Optional, Sequence
 from ._base import DeviceRegistry
 from . import (
     air_dresser, air_monitor, air_purifier, airconditioner, cooktop,
-    dehumidifier, dishwasher, dryer, induction_cooktop, microwave, oven,
+    dehumidifier, dishwasher, dryer, ehs, induction_cooktop, microwave, oven,
     range as _range, range_hood, refrigerator, vacuum_station, washer,
     water_purifier,
 )
@@ -27,6 +27,7 @@ _REGISTRY_BY_KEY: dict[str, DeviceRegistry] = {
     'dehumidifier': dehumidifier.REGISTRY,
     'dishwasher': dishwasher.REGISTRY,
     'dryer': dryer.REGISTRY,
+    'ehs': ehs.REGISTRY,
     'induction_cooktop': induction_cooktop.REGISTRY,
     'microwave': microwave.REGISTRY,
     'oven': oven.REGISTRY,
@@ -87,6 +88,10 @@ _BOARD_TOKEN_TO_KEY: dict[str, str] = {
     'CAC': 'airconditioner',        # issue #191 -- TP1X_DA-AC-CAC-01001_0000
     'ARA': 'airconditioner',
     'DHM': 'dehumidifier',          # issue #88 -- target humidity, no climate
+    'EHS': 'ehs',                   # Eco Heating System air-to-water heat pump --
+                                     # zone1 space heating/cooling + dhw domestic
+                                     # hot water, its own /mode/*/vs/0 and
+                                     # /temperatures/*/vs/0 resource shapes
     'TVTL': 'air_purifier',         # issue #56 (ARTIK051)
     'VTWW': 'air_purifier',         # issue #151 (BESPOKE Cube Air)
     'AVT': 'air_purifier',          # issue #190 -- AVT-WW-TP1-23-AXX500, a
--- a/custom_components/localthings/registry/by_type/ehs.py
+++ b/custom_components/localthings/registry/by_type/ehs.py
@@ -1,0 +1,29 @@
+"""EHS (Eco Heating System) air-to-water heat pump device registry
+(Samsung TP1X_DA_AC_EHS-class).
+
+Shares the DA_AC_ board prefix with the room-AC family in
+airconditioner.py, but its /mode/*/vs/0 and /temperatures/*/vs/0 resources
+are its own shape (two independent loops: zone1 space heating/cooling and
+dhw domestic hot water), not airconditioner.py's HREF_MODE/HREF_TEMP* OCF
+pattern -- so nothing from that module is reused here except MUTE_ONCE,
+whose /option/muteonce/vs/0 field shape (`muteonce`) is identical on this
+family's dump.
+"""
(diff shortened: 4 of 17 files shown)


mbillow:
We can just close #252, I’m fine with this being one PR. The scope of the change isn’t that crazy. I’ll close the other one. Can you amend all commits to get rid of agentic coauthor/committer info. Trying to keep the log clean.

I’ll close the other PR and take a look at this later. Thanks!

mbillow:
## Overall

The shape is right. It genuinely does carry most of the climate lessons: DHW_CONSUMED caps pinned to poll_tier='warm' (the issue #17 lag fix), _rep() translating through subdevice.to_actual (#177), the deduped _warn_unmapped (#93), hrefs declared once in ehs.py and imported so a sibling read can’t drift from its coverage entry, power-on-before-mode-write, reading the same resource writes target, _first() for the single-element modes list, plus golden + tests. The zone1 rationale is well argued and correct.

Three real gaps, though.

### 1. async_set_temperature drops operation_mode — same bug as a9f2198 (#128)

water_heater.set_temperature accepts an optional operation_mode, exactly like climate.set_temperature accepts hvac_mode. Confirmed in HA core:

```python
SET_TEMPERATURE_SCHEMA: VolDictType = {
    vol.Required(ATTR_TEMPERATURE, "temperature"): vol.Coerce(float),
    vol.Optional(ATTR_OPERATION_MODE): cv.string,
}
```

…and async_service_temperature_set forwards every kwarg straight to entity.async_set_temperature(**kwargs).

water_heater.py:async_set_temperature only reads temperature. That’s the identical failure “Honour hvac_mode in the AC climate set_temperature” fixed for climate: a dashboard “boost to 55” button sets the setpoint but never changes mode or powers on. Fix mirrors climate.py exactly — handle operation_mode first, return early if it’s STATE_OFF.

### 2. Away mode — the cited precedent implements it, this doesn’t

The PR leans on HA core’s smartthings water_heater for the Eco/Std/Force/Power map, and that’s accurate. But the same file also does:

```python
Capability.CUSTOM_OUTING_MODE / Attribute.OUTING_MODE == "on"   # -> WaterHeaterEntityFeature.AWAY_MODE
```

That is EHS’s /option/outgoing/vs/0 → x.com.samsung.da.away, which this PR leaves as a standalone config switch. Given the PR’s whole thesis is “use the right platform affordances,” taking the mode map from that integration and not the away map needs resolving one way or the other.

There is a good reason to keep it a switch — /option/outgoing/vs/0 is device-wide, so it affects zone1 too, and hanging it off the DHW card would misrepresent scope. But that reasoning appears nowhere. Either fold it in with AWAY_MODE, or put a comment on ehs.AWAY_MODE saying why it deliberately stays a switch.

### 3. min_temp/max_temp fall back independently

```python
return _num(self._rep(TEMPERATURE_HREF).get('x.com.samsung.da.minimum')) or super().min_temp
```

Two problems, both previously learned here:

	•	Independent fallback. climate._range() deliberately requires lo and hi together. Here, a board reporting minimum but not maximum gives min=40 (device) / max=60.0 (HA’s 140 °F default) — plausible-looking and silently wrong, when the real unit allows 62.
	•	or on a numeric. Collapses a genuine 0.0 to the default — the same shape as c859753 (#160, “stop collapsing a genuine fivepercentHumidity=0”). Same applies to target_temperature_step ... or 0.5, and to ZONE_TEMPERATURE’s or 5.0 / or 30.0 inherited from the earlier commit.

### Smaller

Naming. _attr_name = None makes this the device’s primary entity. That’s right for the AC (the climate card is the AC); it’s shakier for EHS, where DHW is one of two loops and the siblings are named “Zone Mode”, “Zone Target Temperature”. The user gets a water-heater card labelled just “EHS”. Consequences: translation_key='dhw' on the desc is currently dead code (nothing reads a catalog name), and the module docstring claims water_heater.dhw is in UNNAMED_DESCRIPTORS when the actual entry is ("water_heater", "water_heater") — it’s keyed on desc.key. The comment in test_translations.py also conflates two independent things: needing no name entry vs. needing no state entries. The states genuinely need no catalog entry; that says nothing about the name. I’d give it a name (“Hot water”), drop the UNNAMED_DESCRIPTORS entry, and let translation_key='dhw' do its job.

Static reverse map. _STATE_TO_DEVICE vs climate’s dynamic resolution from the unit’s own supportedModes (#155). Low risk in practice — the map is bijective, and HA validates set_operation_mode against operation_list before it reaches you — so only the read side breaks on a board spelling eco/ECO, and _warn_unmapped catches that. A case-insensitive normalize is a cheap hedge, not a blocker.

README. The module tree at line 140 lists every platform module and still omits water_heater.py.

pookey:
Thanks — all three, plus the smaller ones. Force-pushed as one amended commit (keeping the single-commit shape you asked for).

**1. `async_set_temperature` dropped `operation_mode`.** Fixed, mirroring `climate.async_set_temperature` exactly: handle `operation_mode` first (which also powers the loop on via `async_set_operation_mode`), return early on `STATE_OFF`, then write the setpoint. Two tests: one asserting the full `power -> mode -> temperature` sequence from a powered-off loop, one asserting `STATE_OFF` writes only the power-off.

**2. Away mode.** Kept as a switch, with the reasoning written down — you'd identified it correctly. `/option/outgoing/vs/0` carries a single device-wide `x.com.samsung.da.away` flag with no dhw-scoped sibling href, unlike every other resource in the loop, so it covers zone1 too; presenting it on the DHW card would misstate its scope. That's now a comment on `ehs.AWAY_MODE` naming the SmartThings precedent explicitly and saying it stays a switch until a board turns up with a per-loop away resource.

**3. `min_temp`/`max_temp`.** Both fixed:
- Added `water_heater._range()`, same contract as `climate._range()` — both ends together or neither, so a board reporting `minimum` without `maximum` falls back whole instead of pairing 40 with HA's 140 °F default.
- `target_temperature_step` now tests `is None` rather than `or`.
- Same treatment for `ZONE_TEMPERATURE`, which had the same shape inherited from the earlier commit: `_bounds(rep, lo, hi)` resolves the pair in one place (following `oven._setpoint_bounds`) and `_step(rep, default)` uses `is None`. Tests for the half-range fallback and for a genuine `0` increment on both the water_heater and the zone number.

**Smaller:**
- **Naming** — you're right, and the mismatch you spotted between the docstring and the actual `UNNAMED_DESCRIPTORS` entry was the tell that it hadn't been thought through. Dropped `_attr_name = None`, dropped the `("water_heater", "water_heater")` entry, and `translation_key='dhw'` now does its job: `entity.water_heater.dhw.name` = "Hot water" (cs/nl too). The `test_translations.py` comment is rewritten to separate the two claims — states genuinely need no catalog entry; the name is an independent decision, and it goes the other way here.
- **Static reverse map** — took the cheap hedge: a case-folded read-side lookup (`_to_state`), with a comment on why the write side can stay on `_STATE_TO_DEVICE` (the map is bijective, unlike `Wind`/`Fan`).
- **README** — `water_heater.py` added to the module tree.

1114 tests green.

PR #255, opened 2026-08-02
Report whether the auto-clean cycle is running, and how far through

moKorean:
`/option/autoclean/vs/0` carries three fields and only `settingStatus` is read today. That one says the feature is *enabled* — which it is whether or not the unit is drying right now — so nothing reports an actual cycle.

```
settingStatus: On      <- the existing auto_clean switch
status:        Stop    supportedStatus: [Start, Stop]
progress:      0
```

This adds a binary sensor for `status` and a percentage sensor for `progress`.

### Measured

On a `TP1X_DA-AC-CAC-01001`, sampling the resource every eight seconds across a cycle:

| | `status` | `progress` |
|---|---|---|
| mid-cycle | `Start` | `98` |
| the moment it finished | `Stop` | `0` |

The percentage matches the figure the appliance shows on its own display — checked against 55% part-way through, reported by the owner watching both at once.

### Notes

- Golden state keys updated for the 18 fixtures that bind `AUTO_CLEAN`. Additive only; no key was removed from any of them.
- Translations added for `en`, `nl` and `cs`. `cs.json` is CRLF in this repo, and the patch preserves that — worth knowing, since a JSON round-trip silently rewrites the whole file.
- **Not** included: `/selfcheck/vs/0` has the same `status`/`progress` shape and `SELF_CHECK` already maps `status`, `result` and `error` but not `progress`. I have only ever seen that field at `0` on this hardware — no self-check has run — so I have not mapped it here. Happy to add it if you have a dump with one in progress.

Full suite passes: **1070 tests**, Python 3.13 via `requirements-dev.txt`.

Context: same hardware as #230. I maintain a Homey port of this integration ([moKorean/com.lomohome.localthings](https://github.com/moKorean/com.lomohome.localthings)).

### DIFF
--- a/custom_components/localthings/registry/capabilities/airconditioner.py
+++ b/custom_components/localthings/registry/capabilities/airconditioner.py
@@ -742,6 +742,25 @@ AUTO_CLEAN = Capability(
                    write_fn=lambda p, rep, href=None: (
                        ['option', 'autoclean', 'vs', '0'],
                        {'x.com.samsung.da.settingStatus': 'On' if p == 'On' else 'Off'})),
+        # The cycle, as opposed to the switch above -- `settingStatus` says the
+        # feature is enabled, which it is whether or not the unit is drying right
+        # now. `status` is the run state, and the resource advertises exactly
+        # Start/Stop in `supportedStatus`. Observed on a TP1X_DA-AC-CAC-01001:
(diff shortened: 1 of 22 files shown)


jayrage887:
This only updates about once per minute, is this intended behaviour?

PR #256, opened 2026-08-02
Partial Italian translation

g1za:
I translated the custom integration and the labels and text for washing machine (my model).
It can be used as a base for future Italian translation implementation.
Not sure if there is any additional file that needs updating to reference this one.
Cheers!


mbillow:
Thanks @g1za for kicking off the Italian translation — the config/options/issues/exceptions strings and the washing-machine entity labels were exactly right.

This branch's base had drifted from `main`, so rather than resolve conflicts in place, I squashed the commit onto current `main` and filled in the rest of `it.json` (every other appliance family) in #262. Closing this one in favor of that.

---

Issue #266, opened 2026-08-03
Air-quality sensors carry no state_class, so they record no long-term statistics


kkqq9320:
### What's wrong

`dust`, `fine_dust` and `super_fine_dust` show live values fine, but Home Assistant keeps no long-term statistics for them. Once recorder's `purge_keep_days` window passes (10 days by default) the history is gone, so they can't back a long-range air-quality graph.

### Why

HA only writes long-term statistics for sensors that declare a `state_class`, and `air_purifier.AIR_QUALITY`'s descriptors set none — the entities come up carrying just an icon:

```
sensor.<device>_dust
  state: "5"
  attributes: {icon: "mdi:blur", friendly_name: "... Dust"}
  # no state_class, no unit_of_measurement, no device_class
```

The values aren't the obstacle — `common.sensor_item_value` already returns `int`. And three sensors in the same module (`filter_progress`, `fan_speed_level`, `hepa_filter_usage`) already declare one, so this looks like an oversight rather than a decision.

### The part that needs your call

`air_monitor.SENSORS` reads the same `/sensors/vs/0` items through the same helper and **already stamps `state_class="measurement"` on all five** — including `odor` and `clean_level`. Its module docstring describes this as *"matching air_purifier.AIR_QUALITY's existing precedent"*, but that precedent isn't actually set here. So the two modules disagree today.

Two ways to settle it:

1. **All five**, matching `air_monitor`. Simplest, and makes the two consistent.
2. **Particulates only.** They fall monotonically with particle size across three independent board families, which is concentration behaviour where a time average means something:

   | fixture | Dust | FineDust | SuperFineDust |
   |---|---|---|---|
   | `air_purifier` (ARTIK051_TVTL, #56) | 11 | 9 | 5 |
   | `air_purifier_avt_ww` (AVT-WW-TP1, #190) | 10 | 9 | 6 |
   | `range_hood` | 18 | 14 | 9 |
   | `air_purifier_tp1x_da_ac_air` (#130) | 5 | 5 | 5 |
   | `air_purifier_vtww` (#151) | 5 | 5 | 5 |

   `odor` and `clean_level` read 0–2 on every fixture — graded indices, where the mean of a grade is harder to justify.

`range_hood.AIR_QUALITY` has the same gap and would presumably follow whichever way this goes.

### Implementation note

[...]

PR #267, opened 2026-08-03
fix(air_purifier): record long-term statistics for the particulate sensors

kkqq9320:
Addresses #266.

`dust` / `fine_dust` / `super_fine_dust` carried no `state_class`, so Home Assistant kept no long-term statistics for them and their history vanished with the next recorder purge. This stamps `state_class="measurement"` on the three particulate readings.

That's option 2 from the issue — particulates only. The reasoning is in the code comment: they fall monotonically with particle size on three independent board families (11/9/5 on ARTIK051_TVTL, 10/9/6 on AVT-WW-TP1, 18/14/9 on the range hood), which is concentration behaviour where a time average means something, while `odor` and `clean_level` read 0-2 everywhere and look like graded indices.

**If you'd rather have option 1** — all five, matching what `air_monitor.SENSORS` already does — it's a one-line change to the fourth column. I went with the narrower one because I couldn't justify averaging a grade, but `air_monitor`'s docstring already claims to be following this module's precedent, so consistency is a fair argument the other way. Happy to switch.

### Notes

- The `state_class` is the fourth column on `_AIR_QUALITY_SENSORS`. `air_monitor.py` imports the same rows and unpacks four, but discards the fourth: that board has stamped all five readings as `measurement` since it was added, and consuming the column would silently drop long-term statistics for `odor`/`clean_level` there. Two of the new tests guard that coupling and that deliberate divergence.
- No `device_class`/`unit`: pm1/pm25/pm10 with µg/m³ would assert the reading is a mass concentration, which no dump states.
- Metadata only — no key, name, value or unit changes. No entity changes identity and every golden is untouched. Statistics accumulate from the upgrade onward; existing short-term history is unaffected.
- `range_hood.AIR_QUALITY` has the same gap; left for whichever way #266 is decided.

### Verification

`ruff format --check`, `ruff check` and `ty check` all clean. Test suite at 1131 passed / 0 failed.

### DIFF
--- a/custom_components/localthings/registry/capabilities/air_purifier.py
+++ b/custom_components/localthings/registry/capabilities/air_purifier.py
@@ -107,36 +107,12 @@ def _has_top_level_modes(rep, resources):
 # reading is a µg/m³ concentration, and the dumps never say so. That's a
 # separate call from making the series recordable at all.
 _AIR_QUALITY_SENSORS = (
-    ("dust", "mdi:blur", "Dust"),
-    ("fine_dust", "mdi:blur", "FineDust"),
-    ("super_fine_dust", "mdi:blur", "SuperFineDust"),
-    ("odor", "mdi:scent", "Odor"),
-    ("clean_level", "mdi:air-filter", "CleanLevel"),
+    ("dust", "mdi:blur", "Dust", "measurement"),
+    ("fine_dust", "mdi:blur", "FineDust", "measurement"),
+    ("super_fine_dust", "mdi:blur", "SuperFineDust", "measurement"),
+    ("odor", "mdi:scent", "Odor", None),
+    ("clean_level", "mdi:air-filter", "CleanLevel", None),
 )
-
-# Sensors that get a state_class, which is what makes Home Assistant keep
-# long-term statistics -- without one a reading only lives in the short-term
-# recorder history and is dropped at the next purge (10 days by default), so it
-# can't back a long-range air-quality graph. Kept as a key set rather than a
-# fourth column because air_monitor.py imports the tuple above and unpacks it
-# as a triple.
-#
-# Only the three particulate readings are listed. They fall monotonically with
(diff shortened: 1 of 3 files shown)


custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -87,6 +87,25 @@ def _has_top_level_modes(rep, resources):
     return isinstance(rep.get("x.com.samsung.da.supportedModes"), (list, tuple))
 
 
+# The fourth column is state_class, which is what makes Home Assistant keep
```
mbillow, line 90: This comment implies you’re about to do this the right way (add a fourth column to this set)… then you don’t touch the set and add a new data structure and check it in the list comprehension below. Please fix this to just use a fourth column.

kkqq9320: Fixed — `_AIR_QUALITY_SENSORS` now carries `state_class` as a fourth column and the comprehension unpacks it; `_RECORDED_AIR_QUALITY` and its duplicated rationale block are gone. You were right: the comment promised one shape and the code did another, and the duplicated block was the tell. Exactly one block left, and it describes a column that now exists.

One thing worth flagging since it isn't obvious from the diff: `air_monitor.py` unpacks the fourth column but deliberately discards it (`for key, icon, sensor_type, _ in ...`). That board (#210) has stamped all five readings as `measurement` since it was added, whereas this module leaves Odor/CleanLevel unstamped because they read as graded indices on the purifier. Consuming the column there would silently drop long-term statistics for two sensors on shipped devices — a behaviour change this branch has no evidence to justify — so the shared rows supply only key/icon/type to that module. There's a comment at the call site saying so, and a test that guards it.

`test_shared_sensor_tuple_keeps_its_three_column_shape` is replaced rather than renumbered, since its name and docstring asserted the premise you overruled. PR description updated too — it described the old key-set approach. `pytest` 1131 passed; ruff format/check and `ty` clean.

Issue #190, opened 2026-07-29
Samsung Air Purifier AX053B810HGD (AVT-WW-TP1-23-AXX500) is not recognized


gm2945:
### Appliance model

AX053B810HGD

### What's missing or wrong

The appliance is discovered successfully, but it is not recognized as an Air Purifier.

The diagnostics show that one_ui_version is empty, so LocalThings detects the device type as "unknown". As a result, only the common entities are created instead of the Air Purifier entities.

This model supports HEPA filter information, dust sensors, fan modes (SMART/MAX/WINDFREE/Sleep), humidity, and other air purifier resources.

I have attached the diagnostics JSON generated by Home Assistant.

### Diagnostics file

[config_entry-localthings-01KYPMW1QK0ZHR0X6K78YTWAEW.json](https://github.com/user-attachments/files/30502718/config_entry-localthings-01KYPMW1QK0ZHR0X6K78YTWAEW.json)

gm2945:
Additional information:

This is the SmartThings information screen for the appliance.

Consumer model: AX053B810HGD
SmartThings app version: AP.5.1.6

The attached screenshot confirms that this device is a Samsung Air Purifier.

mbillow:
Fixed and released in v0.17.0 — added `AVT` as its own board token, so `AVT-WW-TP1-23-AXX500` now routes to the same air_purifier registry as the existing BESPOKE Cube Air board, with zero unmodeled resources on your dump. You should see full air purifier support (fan modes, HEPA filter, air quality sensors, etc.) once you update. Thanks for the extra SmartThings screenshot confirming the model — let me know if anything looks off once you're on the new version.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #151, opened 2026-07-28
Samsung BESPOKE Cube Air


jhkwon19:
### Appliance model

AX53A9379WGD

### What's missing or wrong

I have a Samsung Bespoke Air Purifier that is detected as **"A-VTWW-TP2-21-COMMON"** in Home Assistant.
Although the device can be discovered, it is not recognized correctly and cannot be controlled.
Could you please add support for this model?
If needed, I can also provide diagnostic logs or any additional information required to help implement support.
Thank you.

### Diagnostics file

[config_entry-localthings-01KYKFFWQG11TEXSAEBB1T9PAS.json](https://github.com/user-attachments/files/30443318/config_entry-localthings-01KYKFFWQG11TEXSAEBB1T9PAS.json)

mbillow:
Thanks for the diagnostics — the BESPOKE Cube Air (A-VTWW-TP2-21-COMMON) is now recognized and supported as of **v0.15.0**. Please upgrade and let us know if anything looks off. Thanks for reporting this!

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #130, opened 2026-07-27
[Device support] Samsung Air Purifier (TP1X_DA-AC-AIR-01031_0000)


taki-769:
### Appliance model

Samsung Air Purifier (TP1X_DA-AC-AIR-01031_0000)

### Device Model
- Device Type: Samsung Air Purifier
- Firmware/Model ID: `TP1X_DA-AC-AIR-01031_0000`

### Description
The device is successfully added via LocalThings. However, only power and child lock switches are created. The `fan` domain entity for controlling fan speeds and modes is missing.

### Diagnostics Data
I have attached the redacted diagnostics JSON file below for adding resource mapping.

### Diagnostics file

[config_entry-localthings-01KYHVB1JCWQKWPATGYV6H8ZAZ.json](https://github.com/user-attachments/files/30419136/config_entry-localthings-01KYHVB1JCWQKWPATGYV6H8ZAZ.json)

mbillow:
Thanks for the diagnostics. Root cause: this newer board (`TP1X_DA-AC-AIR-01031_0000`) reports fan modes (Smart/Max/Mid/WindFree/Sleep) directly on `/mode/vs/0`'s top-level `modes`/`supportedModes` fields, but the registry only knew how to read the *older* air-purifier family's `/mode/vs/0` shape (an `options[]`-packed scheme with no real fan-speed selector at all — that family's board never had one). Both generations share the same href, so I added a second capability discriminated by whether `supportedModes` is present, rather than a new device type.

Your unit now gets a real `fan` entity (`PRESET_MODE` — Smart/Max/Mid/WindFree/Sleep are named behaviors, not an ordered speed, so no percentage slider). While in there, I also noticed your diagnostics still had 16 other unbound hrefs beyond the fan gap (display, HEPA filter, panel status, pet-filter mode, sound settings) — the "incomplete capability coverage" repair was firing on all of those too, so I bound or ignored (with reasoning) each one. Your dump now binds with zero unbound hrefs.

Fixed in 4e1eb7d, on the `claude/triage-issues-111-127-on5l50` branch — not yet merged/released.

---
_Generated by [Claude Code](https://claude.ai/code)_

mbillow:
Added fan-mode control for `TP1X_DA-AC-AIR` air purifiers, plus display, HEPA filter, pet-filter mode, and sound-setting controls that were also missing on this board. A follow-up pass also fixed the fan power write and a silent preset-mode rejection found during review.

[...]

PR #268, opened 2026-08-03
feat(air_purifier): expose the AI Purify sensing engine on /airlevelcheck/vs/0

kkqq9320:
Part of #84 (see the scope note below — this doesn't close it).

`/airlevelcheck/vs/0` has been covered as *"periodic air-quality sensing scheduler plumbing"* since the registry gained a coverage stub for it. Two AVT-WW-TP1-23-AXX500 dumps (#84 and #190) show it isn't plumbing: it drives the feature the SmartThings app calls **AI Purify** — the unit wakes on a timer, samples the air, and optionally acts on the result. Every field is named, none are opaque, and two of them are already user-set on the reported units.

This binds it and adds nine entities: the mode select, the master sensing switch, the interval, the sensing-skip window (switch + two times), and three read-only diagnostics.

### Naming

The select carries the device's own action list rather than a typed-in tuple — its options come straight off `supportedAutoExeState` (`Off` / `Airpurify` / `Alarm`), so a board advertising a fourth action is accepted on both the options list and the write path. The catalog supplies the labels, transcribed from the SmartThings app: **Sensing only** (sample, take no action), **Auto clean** (purify while the air reads bad, stop once it improves) and **Get notified** (raise a SmartThings notification). Taken from the Korean app (감지만 하기 / 자동 청정 / 알림받기) and rendered in English; the auto-stop half of "Auto clean" is the app's own description and isn't otherwise visible in the dump, which reports only the selected `autoExeState`.

Note the switch and the select both have an `off`, and they are not the same thing: the switch's stops the unit sampling at all, the select's is the advertised `autoExeState: "Off"` where it keeps sampling and simply doesn't act on the reading.

Everything else is named after its raw field rather than inventing a concept — the skip window is "sensing skip", after `periodicSensingSkipStatus` / `periodicSensingSkipTime`.

### Which boards this touches

Three of this registry's four board families report the resource with the same field names — TP1X_DA-AC-AIR (#130), A-VTWW-TP2 (#151) and AVT-WW-TP1 (#84, #190). Only ARTIK051_TVTL (#56) has no such href, and its golden is unchanged.

Bound unconditionally rather than behind a `match_fn`; the one field that genuinely varies (`periodicSensingInterval`, absent on the #130 board) is gated per-entity, so that board gets eight entities instead of nine rather than a broken one.

### Why range_hood.AIR_LEVEL_CHECK isn't reused

It already models this same href, and its read-only keys are reused verbatim here so both families share one catalog entry. It's deliberately **not** imported: the hood exposes `periodic_air_sensing` as a read-only `BinarySensorDesc` and this board needs a writable `SwitchDesc` on that key, so reusing the hood's capability would migrate every hood user's entity to a different platform.

### Hardware verification

Every write was exercised on AVT-WW-TP1-23-AXX500. This board returns 2.04 for writes it silently discards (see `HEPA_FILTER`'s filter-reset note), so an echo proves nothing — each was judged by whether the value survived a reconnect, which forces a new DTLS session, fresh discovery and a fresh observe of the href, leaving no cached state to read back.

- `sensing_mode`'s combined two-field PUT lands both fields, both ways: `sensing_only` → `auto_purify` raises `autoExeState` with activation still On, and back again lowers it.
- The sensing-skip switch holds Off → On and back.
- The half-preserving time writes hold: from 13:00–23:00, writing `start=07:30` then `end=22:00` left the device on `07302200` — each write kept the half it wasn't given.
- `periodic_air_sensing` and `sensing_interval`: writing 60 s drove an observed ~60 s sensing cycle.
- The read side of the skip window is separately cross-confirmed on two units: #84's sits at the inert `00000000`, #190's carries a real `03002300` (03:00–23:00), which is what pins the HHMMHHMM split.

The other two families get the writes on field-shape grounds — the same basis on which they already share `MODE`, `HEPA_FILTER` and the air-quality sensors.

### Scope

Deliberately narrow. The same two dumps also support HEPA filter derivations (usage in hours, life remaining, rated life) and a `Pollution_*` air-quality indicator switch; both are left out to keep this reviewable, so **#84 stays open** for them.

`range_hood._timestamp` moves to `common.epoch_to_utc` so both callers share it, matching how `filter_usage_percent` was shared. No behaviour change — that's the only file outside the air-purifier path.

Every existing entity is untouched: the three golden updates are purely additive, no renames, no unit or `device_class` changes.

### Verification

`ruff format --check`, `ruff check` and `ty check` all clean, and the full suite passes — 1140 on the reviewer's run. An earlier revision of this description quoted "468 passed / 1 failed"; that came from an HA-free harness of mine that produces a false negative on `test_every_ac_convenient_mode_code_has_a_preset_label`, and the failure isn't real.

Goldens are untouched by the review changes: the key set never moved, only `sensing_mode`'s value shifted from the folded slug to the raw `autoExeState`.


custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -507,6 +510,246 @@ def _wind_strength_fan_write(payload, rep, href=None):
...
+
+
+AIR_LEVEL_CHECK = Capability(
+    href="/airlevelcheck/vs/0",
+    poll_tier="warm",
+    entities=(
+        SelectDesc(
+            key="sensing_mode",
+            rep_fn=_sensing_mode,
+            options=("off", "sensing_only", "auto_purify", "st_alarm"),
```
mbillow: **This is the one thing I'd want changed before merge.** The device advertises this list and we're not reading it.

All three fixtures carry:

```json
"x.com.samsung.da.supportedAutoExeState": ["Off", "Airpurify", "Alarm"]
```

That field is read nowhere in the PR. Both this tuple and `_SENSING_MODE_BODIES` are typed-in tables, which is the `oven._OVEN_MODES` / #138 shape the skill's §6 calls out: a board generation that adds a fourth `autoExeState` gets it rejected by both the HA options list *and* `_sensing_mode_write`, even though the device itself advertises support for it.

Worth noting this module already argues the other way — `SOUND_MODE`'s comment ~200 lines up says those descriptors read "the live supported values instead of a hardcoded table," and `SOUND_VOLUME` pulls its bounds off the rep via `native_min_fn`/`native_max_fn`.

The folded-select shape isn't an obstacle: `select.py`'s `options=<callable>` form receives the full resource snapshot, so

```python
_AUTO_EXE_SLUG = {"Off": "sensing_only", "Airpurify": "auto_purify", "Alarm": "st_alarm"}

def _sensing_mode_options(resources):
    rep = resources.get("/airlevelcheck/vs/0") or {}
    supported = rep.get("x.com.samsung.da.supportedAutoExeState") or _AUTO_EXE_SLUG
    return ["off"] + [_AUTO_EXE_SLUG.get(v, v.lower()) for v in supported]
```

with `_sensing_mode_write` inverting the same map instead of indexing a fixed dict. I ran this against all three fixtures and it returns exactly the current static tuple on each — so it's behaviour-preserving today and doesn't move the goldens.

---

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -507,6 +510,246 @@ def _wind_strength_fan_write(payload, rep, href=None):
...
+            translation_key="sensing_mode",
+            icon="mdi:radar",
+            entity_category="config",
+            write_fn=_sensing_mode_write,
+        ),
+        SwitchDesc(
+            key="periodic_air_sensing",
```
mbillow, line 683: This switch and the `sensing_mode` select above both drive `periodicSensingActivationState`, and the select already exposes `off` — so this is a second config entity for a state the select fully covers.

The justification in the module comment (the hood models this key read-only, this board needs it writable) argues for the *descriptor type*, not for shipping both controls. Could you add a sentence on what a user does with the switch that the select doesn't already do? If the answer is "nothing," dropping it would leave one control per concept.

---

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -507,6 +510,246 @@ def _wind_strength_fan_write(payload, rep, href=None):
...
+# read back. The other two families get the writes on field-shape grounds, the
+# same basis on which they already share MODE, HEPA_FILTER and the air-quality
+# sensors.
+# ---------------------------------------------------------------------------
+
+
+def _interval_minutes(seconds):
+    """Device stores the interval in seconds; the entity is in minutes."""
+    secs = int_or_none(seconds)
+    return round(secs / 60) if secs else None
```
mbillow: Nit: `if secs` is falsy for a reported `0`, so an interval of zero reads as unknown rather than zero — `if secs is not None` keeps them distinct.

Also worth a thought: anything under 30 s rounds to `0`, which is below the entity's own `native_min=1`.

---

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -507,6 +510,246 @@ def _wind_strength_fan_write(payload, rep, href=None):
...
+                return None
+        return None
+
+    return _read
+
+
+def _skip_time_write(part):
+    def _write(value, rep, href=None):
+        cur = (str(rep.get("x.com.samsung.da.periodicSensingSkipTime", "") or "") + "00000000")[:8]
+        hhmm = f"{value.hour:02d}{value.minute:02d}"
```
mbillow, line 670: Nit: the padding handles a short value, but a malformed one gets carried through verbatim — writing `start` over a junk `periodicSensingSkipTime` sends `"0730" + junk` back to the device. Since `_skip_time_read` already knows how to reject a half it can't parse, normalizing an unparseable half to `0000` here would keep the pair's round-trip honest in the same cases.

---

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -507,6 +510,246 @@ def _wind_strength_fan_write(payload, rep, href=None):
...
+            value_fn=_skip_time_read("end"),
+            write_fn=_skip_time_write("end"),
+        ),
+        # Read-only status, same keys as range_hood.AIR_LEVEL_CHECK.
+        SensorDesc(
+            key="air_sensing_state",
+            field="x.com.samsung.da.sensingState",
+            icon="mdi:radar",
+            entity_category="diagnostic",
+            enabled_default=False,
```
mbillow: `enabled_default=False` here and on `last_air_sensing_level` diverges from `range_hood.AIR_LEVEL_CHECK`, which leaves both enabled for the same href and the same fields — while `last_air_sensing_time` just below, the timestamp of that same sensing event, stays enabled here.

Not wrong for two families to differ, but the PR asserts key parity with the hood while quietly diverging on visibility, and the three keys disagree among themselves. Worth picking one posture and saying why.

---

tests/test_air_purifier_airlevelcheck.py
```diff
@@ -0,0 +1,224 @@
...
+def test_last_sensing_time_reads_as_utc():
+    state = _state("air_purifier_avt_ww")
+    assert state["last_air_sensing_time"].tzinfo is not None
+    assert state["last_air_sensing_time"].year >= 2020
+
+
+def test_read_only_keys_match_the_range_hood_capability():
+    """Same href, same fields -- the keys are shared deliberately so both
+    families read from one translation catalog entry. If either side renames
+    one, this catches the drift."""
```
mbillow: The "one translation catalog entry" claim doesn't hold for `periodic_air_sensing`, which is one of the four keys this assertion covers. It's a `BinarySensorDesc` on the hood and a `SwitchDesc` here — different platforms, so different catalog entries, and they're worded differently today:

```
binary_sensor.periodic_air_sensing -> "Periodic air sensing"
switch.periodic_air_sensing        -> "AI Purify sensing"
```

True for the three sensor keys. Since the next test asserts precisely that platform split, it'd read better to drop `periodic_air_sensing` from this set and let the docstring's claim stay accurate.

---

tests/test_air_purifier_airlevelcheck.py
```diff
@@ -0,0 +1,224 @@
...
+        assert reg is not None, fixture
+        unbound = []
+        discover(resources, reg.capabilities, reg.pattern_capabilities, log=unbound.append)
+        assert unbound == [], f"{fixture}: {unbound}"
+
+
+def test_sensing_mode_folds_toggle_and_action():
+    """Off beats any pending auto-action; On with no action is 'sensing only'."""
+    assert (
+        air_purifier._sensing_mode(
```
mbillow: Style: these assertions reach into module-private helpers (`_sensing_mode`, `_sensing_mode_write`, `_interval_minutes`, `_skip_time_read`/`_write`, `_SENSING_MODE_BODIES`), where the rest of the corpus goes through the descriptor — e.g. `tests/test_air_monitor_capabilities.py:106` does `dnd.write_fn("On", {})`.

This file already has `_desc(key)`; routing through `_desc("sensing_mode").rep_fn(...)` / `.write_fn(...)` covers the wiring as well as the helper, and won't go stale if a descriptor is later pointed at a different function.

---

mbillow (commented):
Reviewed against the `adding-device-support` skill. Nice piece of work — the hardware verification notes are genuinely useful, and the reasoning for not importing `range_hood.AIR_LEVEL_CHECK` is right. One thing I'd want changed, a couple of things worth a reply, and some nits inline.

## Verification

Checked out `d30cad5` on Python 3.13:

- **Tests: 1140 passed** (full suite). The PR body's "468 passed / 1 failed" looks stale — `test_every_ac_convenient_mode_code_has_a_preset_label` passes here on your head commit.
- **`ruff check` / `ruff format --check`**: clean apart from two pre-existing complaints unrelated to this diff (`pyproject.toml` missing `name`, and the fenced Python in `SKILL.md`).
- **`ty check`: clean.**
- No CI check runs have reported on the PR yet — probably fork-PR workflow approval rather than anything in the diff.

## Against the skill

Correct on the things that usually go wrong: the coverage stub is removed from `COVERAGE` with a regression test guarding its return (§8), all five catalogs are updated in lockstep with alphabetical placement and labelled select states (§7), `epoch_to_utc` is hoisted to `common.py` rather than cross-imported (§9), `exists_fn` gates `periodicSensingInterval` instead of a `match_fn` on the whole capability, goldens are purely additive with TVTL explicitly asserted untouched, and the interval's unit conversion is hardware-confirmed rather than guessed (§5). No reporter names in the prose (§10).

## Main finding

**§6 — the `sensing_mode` options are hardcoded while the device advertises the list.** Details inline; short version is that `x.com.samsung.da.supportedAutoExeState` is present on all three fixtures and read nowhere in the PR, and both the options tuple and `_SENSING_MODE_BODIES` are static tables. I've suggested a `options=<callable>` form and confirmed it reproduces the current tuple on all three fixtures, so it's behaviour-preserving and doesn't touch the goldens.

## Worth a reply

**`startSensingOnce` is bound to nothing.** It's on all three dumps (`On`/`Off`) and is the obvious "sense now" one-shot — a `ButtonDesc`. The module docstring says "Every field is named, none are opaque," and the Scope section carefully defers the HEPA derivations and `Pollution_*` but doesn't mention this one. Since you have the unit on the bench, either bind it or add it to the deferred list so the next person doesn't have to re-derive that it was a deliberate omission.

**The select and the switch both drive `periodicSensingActivationState`** — inline.

Everything else is nits. Happy to re-review once the options question is settled.

---

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Correcting myself — I tested this on the unit, and the answer to your question isn't "nothing."

The two fields are independent knobs, and folding them into one select lost two things:

- With `periodicSensingActivationState: Off`, the select read `off` no matter what `autoExeState` held, so a configured action was invisible while the feature was off.
- No select option could write activation without also writing `autoExeState` — every non-`off` body forced one. Toggling the feature while preserving the configured action is the thing only the switch can do.

I'd earlier reasoned that "resume the pending action" wasn't worth an entity because the pending value isn't surfaced anywhere. That was circular: it wasn't surfaced *because* of the fold.

The appliance settles it. Its own UI exposes an on/off for AI Purify separately from the three mode choices — so the device models this as two controls, and the fold was my invention rather than the device's shape.

So rather than dropping the switch, I've unfolded it in `a5484f7`: the switch owns `periodicSensingActivationState` alone, and the select owns `autoExeState` alone.

The two `off`s mean opposite things, so the catalog keeps them distinguishable:

| control | value | meaning |
|---|---|---|
| switch | `off` | AI Purify is off — the unit doesn't sample at all |
| select | `off` | AI Purify is running and sampling — it just doesn't act on what it measures: no purifying, no notification |

The select's `off` is the advertised `autoExeState: "Off"`, which the app calls "sensing only". It isn't a way to turn the feature off; that's the switch.

This also resolves your options comment at the source rather than working around it — see the reply there.

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Agreed, and the `SOUND_MODE` / `SOUND_VOLUME` precedent two hundred lines up makes it worse — this module already argues for reading live values and I typed in a table anyway. You're also right that the field was read nowhere: `grep supportedAutoExeState custom_components/` came back empty.

Fixed in `a5484f7`, and unfolding the switch (see the comment below) let it be fixed at the source rather than worked around. The select now sits on `autoExeState` alone with

```python
options_field="x.com.samsung.da.supportedAutoExeState",
```

which is the same shape `SOUND_MODE` uses for `supportedModes` — no callable needed, since the advertised list lives on this entity's own rep. `_SENSING_MODE_BODIES` and `_sensing_mode_write` are gone; the write body is now just the raw value `select.py` hands back:

```python
write_fn=lambda p, rep, href=None: (
    ["airlevelcheck", "vs", "0"],
    {"x.com.samsung.da.autoExeState": p},
),
```

A board advertising a fourth `autoExeState` now gets it on both the options list and the write path. If the catalog doesn't know the new token, `select.py`'s `_display` falls back to the device's own form rather than dropping it.

Option slugs are the advertised values lowercased — `off` / `airpurify` / `alarm` — rather than the invented `sensing_only` / `auto_purify` / `st_alarm`. That was only ever needed because the folded select had a synthetic `off` for the activation field to collide with; with the fold gone there's nothing to dodge. The catalog carries the labels, so "sensing only" still reaches the user as text.

Goldens don't move: the key set is unchanged and only `sensing_mode`'s value shifts from the folded slug to the raw `autoExeState`.

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Both correct, and the second one points at the actual bug — the conversion is fine, the bound was wrong. Fixed in `a5484f7`.

`is None` now, so a reported `0` stays `0` rather than reading as unknown. And `native_min` drops to 0: the corpus already converts device time values into friendlier units and pairs that with a zero floor — `oven.cook_time` exposes `operationTime` as `unit="min"` via `_op_minutes` with `native_min=0`, and `operational.delay_start_hours` does the same shape with `unit="h"`. I'd matched the conversion half and then set `native_min=1`, which is exactly what made a sub-30s reading render out of range. No clamping — 20 s rendering as 0 min is honest about the rounding rather than showing a value the device isn't reporting.

While this was open, the Number-rather-than-Select choice was worth stating, since the app does present this as a fixed list. It's now in the write helper's comment:

- The SmartThings app offers exactly three intervals — 10 min, 30 min, 1 hour.
- The resource advertises **no** supported-values or range field for `periodicSensingInterval`. `supportedAutoExeState` sits right beside it on the same href, so this board clearly does advertise constraints where it has them, and it has nothing to say here.
- The board accepts values the app doesn't offer: 60 s is six times finer than the app's smallest choice, and it drove an observed ~60 s sensing cycle.

Modelling it as a three-option Select would mean typing in a list the device never advertised and that hardware shows isn't the real constraint — the same failure mode as the `autoExeState` table, just sourced from the app's UI instead of my head.

The remaining `native_max=60` mirrors the app's largest choice; I haven't tested whether the board accepts longer intervals, so I'd rather say that than imply a verified ceiling. Happy to widen or drop it if you'd prefer it not assert a limit either.

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Agreed — the read side rejects a half it can't parse and the write side happily concatenated it, which is inconsistent in exactly the case that matters. Fixed in `a5484f7`: `_skip_half` reuses `_skip_time_read` to decide whether the untouched half is parseable, and zeroes it if not, so the pair round-trips honestly instead of echoing junk back to the device.

`test_skip_time_write_normalizes_a_half_it_cannot_parse` covers both directions alongside the existing preservation test.

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Fair, and the internal inconsistency was the worse half — I'd disabled two of the three read-only keys and left the timestamp of the same sensing event enabled, with no reason behind the split.

Aligned on the hood in `a5484f7`: `enabled_default=False` is gone from both `air_sensing_state` and `last_air_sensing_level`, so all three now match `range_hood.AIR_LEVEL_CHECK`. The PR claims key parity with that capability, so diverging on visibility while asserting parity was the wrong shape. `test_read_only_diagnostics_match_the_hood_on_visibility` locks it in.

tests/test_air_purifier_airlevelcheck.py
kkqq9320: Checked and you're right — they're separate entries with different text today:

```
binary_sensor.periodic_air_sensing -> "Periodic air sensing"
switch.periodic_air_sensing        -> "AI Purify sensing"
```

So the docstring's "one translation catalog entry" claim was false for that key, and it stays false after unfolding the switch — it's still a `SwitchDesc` here against the hood's `BinarySensorDesc`, so different platform catalogs either way.

`a5484f7` drops `periodic_air_sensing` from the assertion, leaving it over the three read-only sensor keys where the claim holds, and the docstring now says so explicitly instead of leaving it implied — including why that key is excluded, pointing at the next test.

tests/test_air_purifier_airlevelcheck.py
kkqq9320: Agreed. `a5484f7` routes the read/write assertions through `_desc(key).value_fn(...)` / `.write_fn(...)` for the select, the switch, the interval and both skip-time entities, matching `test_air_monitor_capabilities.py`. It covers the descriptor wiring as well as the helper, and won't quietly pass if a descriptor is later pointed somewhere else.

`_SENSING_MODE_BODIES` went away entirely with the unfold, so that assertion went with it — as did `_sensing_mode` and `_sensing_mode_write`.

kkqq9320:
Thanks — this is a genuinely useful review, and the `supportedAutoExeState` catch is a fair hit. Everything is addressed in `a5484f7`; replies inline on each thread.

**On the test counts: you're right and my numbers were wrong.** I ran the suite through an HA-free harness (stubbing `custom_components.localthings` to skip the `homeassistant` import), and `test_every_ac_convenient_mode_code_has_a_preset_label` fails under that harness on an untouched `main` too. I read that as "pre-existing failure, unrelated to me" when the real answer is that my harness produces a false negative. Your 1140-passed run is the correct number — I've dropped the stale claim from the PR body.

**The switch comment turned into a design change rather than a fix.** Testing on the unit showed the switch wasn't the duplicate I'd assumed: the two fields are independent, and folding them hid a configured action while the feature was off. The appliance presents them as two controls too. So it's unfolded — switch on `periodicSensingActivationState`, select on `autoExeState` — which also resolves the options finding at the source: the select now reads `supportedAutoExeState` via `options_field` rather than carrying a typed-in tuple, and `_sensing_mode` / `_sensing_mode_write` / `_SENSING_MODE_BODIES` are gone with the fold.

**On `startSensingOnce`:** you're right that the docstring said "Every field is named, none are opaque" while this one was bound to nothing — that inconsistency was mine. It's now recorded as deferred in the module comment rather than silently absent, so the next person doesn't have to re-derive that it was deliberate.

I'd rather not bind it on this pass. It looks like the obvious one-shot trigger, but this board is the one that hands out 2.04 for writes it discards, and I haven't confirmed the side effect — only that the field exists and reads `On`/`Off`. Binding a button on that basis is exactly the mistake the filter-reset note exists to prevent. Happy to do it as a follow-up once someone can watch `lastSensingTime` move in response to the write.

Goldens are untouched throughout — the key set never changed, only `sensing_mode`'s value moved from the folded slug to the raw `autoExeState`.

custom_components/localthings/registry/capabilities/air_purifier.py
kkqq9320: Follow-up on my own reply — dropping `native_min` to 0 fixed the read range and quietly opened a write I didn't intend. `native_min` governs what the user can enter, not just what renders, so 0 became enterable and would have gone out as `periodicSensingInterval: "0"`.

Nothing establishes what that does to this board. Both fixtures report 600, the app's smallest choice is 10 min, and 60 s is the lowest value I've confirmed accepted. And the two precedents I leaned on differ in exactly the way that matters: `oven.cook_time` and `operational.delay_start_hours` both sit at a zero floor under a value where 0 is a real setting ("no timer", "no delay"). Here it means nothing known.

There's also a positive reason for one minute rather than zero, which I'd missed: **that's the resolution this board reports results at.** `lastSensingTime` lands on an exact minute on every sample from the AVT-WW-TP1 and A-VTWW-TP2 boards — both fixtures (`% 60 == 0`) plus eleven consecutive live readings — where the TP1X, AC and hood boards report arbitrary seconds. So a sub-minute interval is unobservable here whether or not the board honours it.

`96d0636`:

- `native_min` is 1, so the UI never offers a value that would be refused.
- The read rounds **up** rather than to nearest, so a hypothetical sub-minute reading renders as 1 instead of 0 — which is what your original point was about. A reported literal 0 still reads as 0 rather than unknown, per the `is None` fix.
- The write refuses anything under a minute anyway, since `native_min` only guards the UI path and not a service call. `None` return, the same silent no-op `range_hood._lamp_level_write` uses.

`test_interval_floor_is_one_minute` covers it.

PR #276, opened 2026-08-04
Fix AC temperature step quantization

moridew:
### Summary
- Fixes an issue where Samsung SmartThings local AC temperature writes did not preserve `0.5` degree steps.
- Previously, temperature values were always rounded to integers, causing half-degree setpoints to be lost.

### Changes
- `custom_components/localthings/registry/capabilities/airconditioner.py`
  - Read the device-reported temperature increment (`increment`, `x.com.samsung.da.increment`)
  - Quantize temperature writes according to that step size
  - Preserve integer vs fractional formatting correctly when serializing values
- `tests/test_airconditioner_capabilities.py`
  - Add regression coverage to verify AC climate writes preserve half-degree temperature steps

### Test
```bash
python -m pytest -q tests/test_airconditioner_capabilities.py -k climate_write
```

### Notes
- Branch: `fix-ac-temperature-step`
- Remote: `myfork/fix-ac-temperature-step`

### DIFF
--- a/custom_components/localthings/registry/capabilities/airconditioner.py
+++ b/custom_components/localthings/registry/capabilities/airconditioner.py
@@ -449,7 +449,48 @@ def _humidity(rep):
     return None
 
 
-def _climate_write(payload, rep, href=None):
+def _quantize_temperature(value, step):
+    """Quantize a temperature to the device's advertised step size."""
+    try:
+        numeric = float(value)
+    except (TypeError, ValueError):
+        return None
+    if step is None:
+        return numeric
+    try:
+        step_value = float(step)
+    except (TypeError, ValueError):
+        return numeric
+    if step_value <= 0:
+        return numeric
+    quantized = round(numeric / step_value) * step_value
+    if quantized.is_integer():
+        return int(quantized)
+    return quantized
+
+
+def _temperature_step(rep, resources=None):
+    """Return the device-reported temperature increment from the coordinator snapshot."""
+    candidates = []
+    if resources is not None:
+        candidates.extend(
+            (
+                resources.get(HREF_TEMP_CONTROL),
+                resources.get(HREF_TEMPS_VS),
+            )
+        )
+    candidates.append(rep)
+    for resource in candidates:
+        if not isinstance(resource, dict):
+            continue
+        for key in ("increment", "x.com.samsung.da.increment"):
+            step = _num(resource.get(key))
+            if step is not None:
+                return step
+    return None
+
+
+def _climate_write(payload, rep, href=None, resources=None):
     """Map a (kind, value) command from the climate platform to the
     (path_segs, body) for that one sub-write. `value` is already the raw device
     code (the platform maps HA<->device). async_send_command POSTs to path_segs,
@@ -478,18 +519,25 @@ def _climate_write(payload, rep, href=None):
     if kind == "mode":
         return (["mode", "vs", "0"], {"x.com.samsung.da.modes": [value]})
     if kind == "temperature_ocf":
-        return (["temperature", "desired", "0"], {"temperature": round(float(value))})
+        step = _temperature_step(rep, resources)
+        quantized = _quantize_temperature(value, step)
+        return (["temperature", "desired", "0"], {"temperature": quantized})
     if kind == "temperature":
         # Vendor items[] array; only one item observed on every AC dump, id
         # '0'. See the docstring above for why this doesn't echo current/
         # minimum/maximum/unit back at the unit.
+        step = _temperature_step(rep, resources)
+        quantized = _quantize_temperature(value, step)
+        desired = quantized
+        if isinstance(quantized, float) and quantized.is_integer():
+            desired = int(quantized)
         return (
             ["temperatures", "vs", "0"],
             {
                 "x.com.samsung.da.items": [
                     {
                         "x.com.samsung.da.id": "0",
-                        "x.com.samsung.da.desired": str(round(float(value))),
+                        "x.com.samsung.da.desired": str(desired),
                     }
                 ]
             },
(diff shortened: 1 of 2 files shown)


jayrage887:
My Aircon only does full 1 degree steps, does that mean this will break it?

moridew:
> My Aircon only does full 1 degree steps, does that mean this will break it?

No — this should not break a 1-degree-only unit.

How it works:
- climate.py exposes `target_temperature_step` as `1.0` when the device does not advertise an increment.
- airconditioner.py only applies step-aware rounding when the device explicitly reports `increment` / `x.com.samsung.da.increment`.
- If that field is missing, `_quantize_temperature()` returns the requested temperature unchanged.

So for a unit that only supports full degrees:
- the UI will fall back to 1°C steps,
- the write path will not force 0.5°C behavior,
- and only devices that advertise `0.5` will preserve half-degree writes.

In short: this patch preserves 0.5°C support where advertised, but does not break 1°C-only aircons.

In my local test, the patch works for both:
- living room AC that advertises `0.5` support
- bedroom AC that only supports `1.0`

How it behaves:
- If the device reports `increment` / `x.com.samsung.da.increment`, the patch preserves half-degree writes.
- If the device does not report it, the existing fallback behavior remains, and the write path does not force 0.5°C.
- climate.py already falls back to `1.0` for `target_temperature_step` when no increment is advertised.

So the patch is additive: it enables 0.5°C support where advertised, without breaking a 1°C-only unit.

Attach the Diagnostics file from my device.
[config_entry-localthings-01KZ3ZDG2MQ74JES7E12FHTP4X.json](https://github.com/user-attachments/files/30728388/config_entry-localthings-01KZ3ZDG2MQ74JES7E12FHTP4X.json)

mbillow (commented):
Checked out the branch, ran the suite (1192 pass), and probed the new helpers against the real AC fixtures. The problem is real and the direction is right: `/temperature/control/vs/0` genuinely advertises `0.5` on CAC and TP1X FAC boards, and on those the fix works — verified `23.6 → "23.5"` on the vendor path. Two issues below would need fixing before merge, plus the branch currently conflicts with main.

## 1. Blocker — the `/temperatures/vs/0` fallback never matches

`airconditioner.py:479` reads the increment off the resource rep top-level:

```python
resources.get(HREF_TEMPS_VS),
...
step = _num(resource.get(key))
```

But on that resource the increment lives inside the `items[]` array, in every fixture:

```json
{"href": "/temperatures/vs/0", "rep": {"x.com.samsung.da.items": [
  {"x.com.samsung.da.id": "0", "x.com.samsung.da.increment": "0.5", ...}]}}
```

So this candidate is dead — the fix only ever works via `/temperature/control/vs/0`. The file already has `_temps_vs_item()` at line 176 for exactly this unwrapping (`climate.py:438` uses its own copy the same way).

The new test passes only because it fabricates a flat shape (`{"/temperatures/vs/0": {"x.com.samsung.da.increment": "0.5"}}`) that no device produces.

## 2. Blocker — with no advertised increment, writes are no longer rounded at all

When `_temperature_step` returns `None`, `_quantize_temperature` returns the raw float (lines 458-459). Previously both branches did `round(float(value))`. Verified against the ARTIK051 fixtures, which have no `/temperature/control/vs/0` **and** no increment in their temps_vs item:

```
ARTIK051 step found    : None
ARTIK051 ocf write 23.6: (['temperature','desired','0'], {'temperature': 23.6})
```

That's `23.6` on the wire to a board that used to receive `24`. The edit to `test_climate_write_targets` (`{"temperature": 24}` → `{"temperature": 23.6}`) pins the regression in as expected behavior rather than catching it.

`climate.py:439` already defaults `target_temperature_step` to `1.0` when nothing is advertised — the write path should mirror that (`... or 1.0` in `_temperature_step`) rather than pass values through unrounded.

This is also the case @jayrage887 asked about above. The reply is right that half-degree behavior isn't forced on a 1°-only unit, but it doesn't cover this: on a board that advertises no increment anywhere, the requested value is now sent unrounded.

## 3. Non-numeric values POST `null`

`_quantize_temperature` returns `None` on `TypeError`/`ValueError`, and both branches drop that straight into the body — confirmed `{"temperature": None}` goes out. The coordinator already has a rejection path (`coordinator.py:951` logs and skips when `write_fn` returns `None`); `_climate_write` should return `None` here instead of building a body around it.

## 4. int/float normalization differs between the two branches

The vendor branch re-normalizes integral floats (lines 531-533), the OCF branch doesn't — so with no step, `24.0` goes out as a CBOR float where it used to be int `24`. Worth normalizing once inside `_quantize_temperature` and dropping 531-533. (Mostly moot once #2 is fixed, but the asymmetry is still confusing.)

## Minor

- **Float noise in `str(desired)`**: `round(21.7/0.1)*0.1` → `21.700000000000003` → serialized as `"21.700000000000003"`. No fixture advertises `0.1` today, so this is theoretical — a `round(quantized, 2)` after quantizing is a cheap guard.
- **`rep` as a step candidate is dead** (line 482): `ClimateDesc` has `rep_fn=_first_mode`, so `rep` is the mode resource and never carries an increment.
- **Test should use a real fixture shape** — that's what let #1 through. `airconditioner_cac_device.json` advertises `0.5` on both resources; an ARTIK051-shaped no-increment case would pin the fallback from #2.

## Housekeeping

Branch is behind and conflicts with main — both touched files moved (#296, plus a ruff format pass). Needs a merge/rebase before it can land. Changed files are ruff-clean and the full suite passes on the branch as-is.

---

mbillow:
Closing in favor of #344, which squashes this branch's commit and applies the review feedback above (items[] unwrapping for the `/temperatures/vs/0` fallback, whole-degree fallback when no increment is advertised, rejecting non-numeric writes instead of posting null, and single-point int/float normalization). Thanks for the initial fix, @moridew — the direction here was right, this just closes the gaps the review found.

---

Issue #168, opened 2026-07-28
Proposal: the legacy 8888/HTTPS appliance family behind a transport seam?


perseus177:
Asking before writing any code, since this is an architecture call that's yours to make.

## The situation

I have a **WW6500 washer** (`TP6X_WW6500|FF1BE000|...`, `TP6X_WASHER`, SW `02095A200308`) that is exactly the family the README scopes out: nothing on UDP `49152-49160`, only **TCP 8888** — nginx/1.2.7 over HTTPS with a mandatory client certificate.

That family *is* locally controllable. I have it running: reading the live cycle, machine state, progress percentage and remaining time, water temperature / rinse cycles / spin level, per-cycle energy from the appliance's own usage DB, plus start / pause / cancel and the "laundry out" timer. It's a separate integration ([`perseus177/ha-samsung-washer-local`](https://github.com/perseus177/ha-samsung-washer-local)) written the same shape as this one, because 8888 is explicitly out of scope here.

## Why I think it's worth raising rather than just leaving it separate

The interesting part is that **the two protocols carry the same vocabulary in a different envelope**:

| | localthings (CoAP-DTLS) | 8888 (HTTPS+mTLS) |
|---|---|---|
| cycle | `/mode/vs/0` → `{"x.com.samsung.da.options": ["Course_5C", ...]}` | `/devices/0/mode` → `{"options": ["Course_5C", ...]}` |
| state | `/operational/state/vs/0` | `/devices/0/operation` → `state`, `progress`, `progressPercentage`, `remainingTime` |
| settings | `/washer/vs/0`-class fields | `/devices/0/washer` → `waterTemperature`, `rinseCycles`, `spinLevel` + `supported*` |

Same tokens (`Course_5C`, `LaundryOutTime_0`, `EnergyKW_396`), same single-token `options` merge semantics on writes — just bare field names instead of `x.com.samsung.da.*`, and `/devices/0/<x>` instead of `/<x>/vs/0`.

So a shim that reshapes an 8888 response into a normal rep dict would let **`registry/`, `discovery.py`, `adapter.py` and `by_type/washer.py` work unchanged**. None of the entity mapping would need duplicating — which is the part that's actually expensive, and the part this repo has already solved.

## What would genuinely need building

1. **A transport seam.** Today the coordinator's write path is CoAP-shaped (`path_segs + body` → POST). An 8888 device wants `PUT /devices/0/<resource>` with a bare-field JSON body.
[...]

PR #289, opened 2026-08-04
feat(airconditioner): reset the legacy filter counter locally

perseus177:
The legacy ARTIK051 filter counter can be reset locally after all. The token is `FilterCleanAlarm_Clear`, and it goes through the same single-token options merge as every other setting on `/mode/vs/0`.

*Disclosure, as in #146 and #168: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

## Measured

On an `ARTIK051_KRAC_18K`, powered off, counter at 9 h 30 min:

```
POST /mode/vs/0   {"x.com.samsung.da.options": ["FilterCleanAlarm_Clear"]}
-> 2.04 Changed, body echoed the token
```

| | before | +5 s | +25 s | +70 s | fresh DTLS session | HA |
|---|---|---|---|---|---|---|
| `FilterTime_` | `95` | `0` | `0` | `0` | `0` | `0.0` |

None of the other 17 tokens moved, and both `/alarms/vs/0` entries stayed `Deleted`. The button has since been pressed through HA end to end on the deployed integration, with no error and the counter staying at zero.

## Why the descriptor said this was unsolved

The comment block this PR replaces was written after two rounds of failures, and it concluded the reset had to be cloud-only. That conclusion was wrong, and how it was reached is the part worth keeping:

it came from **diffing every resource the appliance reports, before and after pressing reset in Samsung's own app**. The diff showed the counter zeroing and the alarm clearing, and nothing else — from which I concluded that the cloud writes the same value we do and the board simply accepts it only from Samsung. But a *trigger* token cannot appear in such a diff, because a trigger is never stored. The board acts on it and forgets it; `FilterCleanAlarm_` is not among the tokens this generation reports at all.

Both failures stay in the comment, because they say what this is **not**:

* `FilterTime_0` — writing the *value*. Accepted with no error and discarded: `5595` → `5595` after 69 s, `1925` → `1925` after 65 s, two units in opposite power states.
* The cloud capability's command name (`custom.dustFilter` / `resetDustFilter`) POSTed to `/actions/vs/0` in five envelope shapes — all `4.00` with an empty body. The name is real; it names a cloud command, and this transport does not carry it.

Samsung's own app for this appliance sends the token above, and skips the write entirely when the counter is already zero.

## Shape

- `ButtonDesc(key="filter_time_reset", payload="FilterCleanAlarm_Clear")`, writing `{"x.com.samsung.da.options": [payload]}` to `["mode", "vs", "0"]` — the existing `_option_switch_write` idiom, spelled out because the payload is a fixed trigger rather than a value.
- Gated on `_has_option_token("FilterTime")`, so it appears only where there is a counter to reset. Newer boards report filter usage through their own resource and would need a different mechanism; offering a legacy token there would be a guess.
- `entity_category="config"`, `mdi:restart`, one `en.json` string.
- Two tests in `tests/test_airconditioner_artik051_krac.py`: the write payload, and absence on a board without the counter token.

Branch is cut from current `main` (0.19.0). No behaviour changes anywhere else.


mbillow:
Closing in favor of #293 — same change (all credit to @perseus177 for the find), rebased onto current `main` to resolve the conflict with #292's comment cleanup, with a `Co-Authored-By` trailer removed from one commit per `CONTRIBUTING.md`. See #293 for the full diff and verification.


---

PR #297, opened 2026-08-05
feat(airconditioner): say what an error code means, in the appliance's words

perseus177:
`/alarms/vs/0` reports a fault as `ErrorCode_<code>` and `alarm_code` surfaces it verbatim. That is right for an automation and useless for a person: `E464` says nothing, and the code alone sends you to the manual. This adds an `error_message` sensor beside it, so the same fault reads **"IPM Over Current (INV)"**.

*Disclosure, as in #146, #168, #290 and #296: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

## Where the table comes from, and a call for you to make

The appliance's own app resolves the code against a catalog of 45 entries, reached by taking the part after the underscore and lower-casing it (`ErrorCode_E464` → `racmob_error_message_e464_result`). That table is what this PR adds.

**The English strings are Samsung's own, quoted rather than rewritten.** That is deliberate — Home Assistant then shows what the appliance's app would have shown, and nobody has to trust my paraphrase of what `E472` means — but it does mean 45 vendor strings land in your repository, and that is your call rather than mine. I raised it with the owner and his position is that a PR is exactly the place to decide it, so: if you would rather have wording of our own, say so and I will replace every line with mine (that is what our washer integration does for its own fault table, for the same reason).

They stay in English whatever the user's language, and only the entity's *name* is translated: the strings are the message, not a label, and rewording them per language would be inventing text the appliance never used.

## Behaviour

| the appliance reports | the sensor reads |
|---|---|
| `ErrorCode_E464`, state `Created` | `IPM Over Current (INV)` |
| `ErrorCode_E554` | `Gas shortage error` |
| a code outside the 45 (`ErrorCode_E999`) | `E999` |
| `ErrorCode_OFF`, or any `Deleted` row | `none` |
| `FilterAlarm` | `none` |

The unknown-code case is on purpose: 45 is what one app build knew, not what the firmware can report, and an unrecognised code is still the thing a service call asks for. Deleted rows and the `_OFF` placeholder are skipped, the same rows `common._active_alarm_codes` already skips. Reminders like `FilterAlarm` are left alone — they have their own entities and are not faults with a code.

## Shape

- `_ERROR_MESSAGES` plus `_error_message()` in `capabilities/airconditioner.py`.
- `common.ALARMS` swapped for `ALARMS_WITH_MESSAGE` in the AC registry, the same pattern `ENERGY_METER` already uses there, so the shared `alarm_code` sensor is untouched for every other family.
- One name string in each of the five catalogs; 22 AC goldens gain the key.
- Five assertions covering the table hit, the unknown code, the deleted row and the reminder.

Cut from current `main`. Independent of #293 and #296.

**One caveat I would rather state than hide:** the 45 strings were transcribed from an extraction run against the app's catalog, and the machine holding it is currently powered down, so I have not re-verified them byte for byte since. I will do that when it is back on and push a correction if anything differs.

### DIFF
--- a/custom_components/localthings/registry/capabilities/airconditioner.py
+++ b/custom_components/localthings/registry/capabilities/airconditioner.py
@@ -266,6 +266,113 @@ def _legacy_cumulative_power_kwh(v):
     return round(n / 100000.0, 2)
 
 
+# ---------------------------------------------------------------------------
+# What an error code means, in the appliance's own words.
+#
+# /alarms/vs/0 carries the code as `ErrorCode_<code>`, and common.ALARMS surfaces
+# it verbatim -- useful for an automation, useless for a person: "E464" says
+# nothing. The appliance's own app resolves it against a catalog of 45 codes
+# (RACMOB_error_message_<code>_result in its language file, reached by taking the
+# part after the underscore and lower-casing it, exactly as below).
+#
+# The English strings here are Samsung's own, quoted from that catalog rather
+# than rewritten, so what Home Assistant shows is what the appliance's app would
+# have shown. They stay in English regardless of the user's language: they are
+# the message, not a label, and translating them here would mean inventing
+# wording the appliance never used.
+#
+# A code outside this table becomes the code itself rather than nothing -- the
+# table is what one app build knew, and an unrecognised code is still the thing a
+# service call will ask for.
+_ERROR_MESSAGES = {
+    "E101": "Indoor unit communication reception error",
+    "E121": "Short circuit or open circuit of the indoor temperature sensor",
+    "E122": "A short or open circuit of the Eva-in sensor of the indoor unit",
+    "E123": "A short or open circuit of the Eva-MID sensor of the indoor unit",
+    "E154": "Indoor fan error",
+    "E162": "EEPROM error",
+    "E163": "EEPROM option setting error",
+    "E201": "Indoor/outdoor unit communication error",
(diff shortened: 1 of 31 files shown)


custom_components/localthings/registry/capabilities/airconditioner.py
```diff
@@ -266,6 +266,113 @@ def _legacy_cumulative_power_kwh(v):
...
+ALARMS_WITH_MESSAGE = replace(
+    common.ALARMS,
+    entities=(
+        *common.ALARMS.entities,
+        SensorDesc(
+            key="error_message",
+            field="x.com.samsung.da.items",
+            icon="mdi:alert-circle-outline",
+            entity_category="diagnostic",
+            value_fn=_error_message,
```
mbillow, line 370: Until something like [`ListEntity`](https://github.com/home-assistant/architecture/discussions/1319) is available, I’m not really onboard with this change. Because error codes are currently surfaced as a comma-separated string, they are already one of the only things we can’t translate. Baking English into the component seems even worse. If Home Assistant had better handling for multi-value entities, I can definitely see revisiting this.

mbillow:
I really appreciate all the work you’re doing to improve support for your devices! I just don’t think this is a great UX improvement with the current inflexible way we’re forced to surface error codes. If others disagree we can revisit, but this seems like a great thing to implement in an automation.

PR #299, opened 2026-08-05
fix(airconditioner): Good Sleep needs the mode token its duration belongs to

perseus177:
`number.*_good_sleep` does not currently set Good Sleep. It writes `Sleep_<n>` on its own, and a legacy board answers `2.04 Changed` and then throws the value away. #296 fixed the *unit* that number is published in; this fixes the write not landing at all.

*Disclosure, as in #146, #168, #290, #296 and #297: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

## Measured on an ARTIK051_KRAC_18K

Four writes, each read back off the appliance — not out of Home Assistant — at +8 s and +45 s, because this board has been seen to accept a token and drop it a poll later:

| written | +8 s and +45 s |
|---|---|
| `["Sleep_4"]` — what this integration sends | `Comode_Off`, **`Sleep_0`** |
| `["Comode_Sleep", "Sleep_4"]` — what the app sends | `Comode_Sleep`, `Sleep_4` |
| `["Comode_Nano"]` — this integration's nano preset, over a live timer | `Comode_NanoSleep`, **`Sleep_16`** |
| `["Comode_NanoSleep", "Sleep_4"]` | `Comode_NanoSleep`, `Sleep_4` |

Row 2 is the control for row 1: the same duration, in the same session, one token apart — one holds and one does not. So `Sleep_<n>` is a parameter of a mode rather than a setting of its own, which is also the only form `getGoodSleepCommand()` ever puts on the wire.

Row 3 is the second bug. Nano wind and Good Sleep share the single `Comode_` slot, so the board upgrades a bare `Comode_Nano` to `Comode_NanoSleep` by itself — and then supplies a duration of its own, turning the user's two hours into eight with nothing in Home Assistant showing it happened. `getNanoWindCommand()` sends the pair for exactly this reason.

## Three changes

1. **`good_sleep` writes its mode token with the duration.** `Comode_Sleep` + `Sleep_<n>`, or `Comode_NanoSleep` + `Sleep_<n>` when nano is on. Zero leaves the mode too — `Comode_Off`, or `Comode_Nano` if nano was running, so switching the timer off does not also switch nano off. That is how the app reads it back.

2. **The nano preset carries a running timer's own duration** instead of letting the board pick one.

3. **`Sleep` and `NanoSleep` became presets.** They have to be: they live in the same slot, so a unit running Good Sleep — set from its *remote*, with no involvement from Home Assistant — reports a code that was not in the list, and a `preset_mode` outside `preset_modes` is not a state HA allows. Verified against a live unit:

   ```
   preset_mode:  "sleep"          # board on Comode_Sleep
   preset_modes: ["none","nano","quiet","comfort","2step","speed"]
   ```

   Gated on the `Sleep_` token, so boards without it are unaffected. Selecting one of them outright has no duration to reuse, so it takes `Sleep_16` — the value the appliance itself falls back to in row 3 above, rather than a number of my choosing.

`nanosleep` is already in all six catalogs (the newer boards expose it as a convenient mode); `sleep` is added to each, with the other five translated from the English.

## Caveat I would rather state

Rows 1 and 2 were each measured twice; rows 3 and 4 once, in one clean run. I have not tested whether a *heat*-mode Good Sleep behaves differently — the app builds a different body for `Heat` (no `"modes": ["Cool"]` fixup) but sends the same option pair, which is all this touches.

Cut from current `main`. Independent of the Smart Saver PR, except that both touch `_LEGACY_PRESET_CODES`, so whichever lands second wants a one-line rebase.

### DIFF
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -267,15 +267,26 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
     # Nano=windFree, Quiet, Comfort, 2Step, Speed=Fast Turbo, Off=none.
     _LEGACY_PRESET_CODES = ("Off", "Nano", "Quiet", "Comfort", "2Step", "Speed")
 
+    # Good Sleep occupies the same Comode_ slot as the presets above, so a unit
+    # running it reports Comode_Sleep -- or Comode_NanoSleep, which the board
+    # will produce by itself when nano wind is asked for while the timer runs.
+    # Neither is in the list above, and a preset_mode outside preset_modes is
+    # not a state HA allows, so they are added for boards that have the Sleep_
+    # token these codes come with.
+    _LEGACY_SLEEP_PRESET_CODES = ("Sleep", "NanoSleep")
+
     def _legacy_convenient(self) -> dict:
         """A /mode/convenient/vs/0-shaped rep built from the Comode_* token in
         /mode/vs/0's options, for boards that have no convenient resource."""
         options = self._rep(MODE_HREF).get("x.com.samsung.da.options") or []
+        codes = list(self._LEGACY_PRESET_CODES)
+        if any(isinstance(option, str) and option.startswith("Sleep_") for option in options):
+            codes += self._LEGACY_SLEEP_PRESET_CODES
         for option in options:
             if isinstance(option, str) and option.startswith("Comode_"):
                 return {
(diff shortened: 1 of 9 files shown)


mbillow:
Closing in favor of #304

PR #303, opened 2026-08-05
Allow non-blocking setup when device is offline (#295)

### LINKED ISSUE #295: Setup uses exponential backoff up to 15min for offline devices — unacceptable UX for local integration
Problem: If a Samsung appliance is powered off during HA restart, when the device comes back online it can take up to 15 minutes before HA reconnects. For a fully local LAN integration this is unacceptable UX — the expectation for a device that was just powered on is sub-30-second recovery.

Root cause: async_setup_entry raises ConfigEntryNotReady when the first refresh fails, and HA's built-in retry uses exponential backoff (30s → 60s → 120s → ... → capped at 15min). This backoff pattern exists to avoid hammering cloud APIs, but is inappropriate for a LAN-local device where the only party affected by frequent probes is the device itself, and a 5-15s DTLS handshake attempt every 30s costs nothing meaningful.

Suggested approach: Follow the pattern used by ESPHome, Shelly local, LIFX, WLED, etc.:

async_setup_entry never fails on device-unreachable — it always succeeds, so entities are created and appear as unavailable
The coordinator runs its own reconnect loop at a fixed short interval (~30s), independent of HA's ConfigEntryNotReady machinery
Entities become available on the first successful poll, no user action required

Additional suggestion: If Samsung firmware broadcasts any mDNS service on boot, wire zeroconf discovery so the integration is notified the instant the device announces itself and can retry immediately — sub-5s recovery instead of waiting for the next tick.

Currently: users must manually reload the config entry or wait an indeterminate amount of time. Not a viable UX for automation-critical devices.

firstof9:
Fixes #295

### Problem
If a Samsung appliance is powered off during Home Assistant restart, `async_setup_entry` raises `ConfigEntryNotReady`. HA's built-in retry uses exponential backoff up to 15 minutes, which is poor UX for a LAN-local integration.

### Solution
- Catch initial connection / refresh errors in `async_setup_entry`, log a warning, and proceed with setup.
- Platforms are forwarded and entities are created in Home Assistant (in an unavailable state until the device comes online).
- The coordinator continues background retry polling.
- Updated unit tests to verify offline setup succeeds cleanly.

### DIFF
--- a/custom_components/localthings/__init__.py
+++ b/custom_components/localthings/__init__.py
@@ -7,7 +7,6 @@ import logging
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import EVENT_HOMEASSISTANT_STOP
 from homeassistant.core import Event, HomeAssistant, callback
-from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers import entity_registry as er
 
@@ -147,14 +146,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     try:
         await coordinator.async_config_entry_first_refresh()
     except Exception as err:
-        # `_poll_once` deliberately leaves the session up on a TimeoutError
-        # (see its docstring), so a refresh failing that way leaves a live,
-        # bound UDP socket nothing would ever close. HA retries setup with a
-        # new coordinator, and the source port is fixed by design
-        # (`_local_source_port`), so an abandoned socket would squat the
-        # exact port the next attempt binds.
-        await coordinator.async_close()
-        raise ConfigEntryNotReady(f"Cannot connect to device: {err}") from err
+        _LOGGER.warning(
+            "Initial connection to device failed (%s); starting offline and retrying in background",
+            err,
+        )
     hass.data[DOMAIN][entry.entry_id] = coordinator
 
     # Send the DTLS close_notify on Core shutdown, not just on unload (issue
--- a/tests/localthings/test_coordinator.py
+++ b/tests/localthings/test_coordinator.py
@@ -67,8 +67,8 @@ async def test_summary_interval(hass: HomeAssistant, mock_entry, mock_coordinato
     assert coordinator.update_interval == timedelta(seconds=SUMMARY_INTERVAL_S)
 
 
-async def test_update_failed_on_persistent_poll_error(hass: HomeAssistant, mock_entry) -> None:
-    """ConfigEntryNotReady raised when poll fails even after reconnect."""
+async def test_offline_initial_setup_succeeds(hass: HomeAssistant, mock_entry) -> None:
+    """ConfigEntry setup succeeds in offline mode when initial poll fails."""
 
     with (
         patch("custom_components.localthings.coordinator.LocalThingsCoordinator._connect_session"),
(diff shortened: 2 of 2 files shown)


firstof9:
@mbillow I still hope we could get this merged.

mbillow:
Sorry, I’m on vacation this week. 🏖️ This one is going to take some thought, so I was planning on getting to it when I get back next week. Sorry for the delay.

firstof9:
No worries, it basically makes the device and entities not do the infinite "initializing" and just loads them while re-connection occurs in the background, similar to how the cloud integration works.

have a good vacation man.

mbillow:
So, at a high level, I have a few concerns with enabling this pattern. 

Currently, all devices use the same setup flow, so this would work. There is active work in the underlying library to support a newer flow that involves interacting with the device during setup. If we ever support the pattern #168, those devices also involve interaction during setup.

If we allow this pattern now, we could end up in a situation where we need a user to interact with their device with no good way to trigger that setup.

Also, if a device isn’t responsive we don’t have any metadata to create a device from. The end result of this is just a config entry that ends up sitting there completely useless. Maybe I am missing something, but I don’t see why that is useful? Can you maybe elaborate on some user stories this would help enable?

mbillow:
Oh, also, thanks for your patience on this. Sorry to leave it lingering for so long!

firstof9:
> Can you maybe elaborate on some user stories this would help enable?

I'd just like the devices and entries already configured to load their previously recorded states so the device/entities are not left in an "unknown"/"unavaliable" state until the appliance is physically used again.

mbillow:
Oh, man, I’m sorry. I was moving too fast and completely misread what your goal was here. That’s my bad. Give me a second to reread this and think about the implications of the current implementation.

mbillow:
Closing in favour of #374, which carries your commit at the base of the branch.

Thanks for the patience on this one, and sorry it sat so long. Re-reading your comment about wanting already-configured devices to load rather than sit unavailable: that's the right goal, and it's what #374 implements. It just needed more underneath it than the setup change alone.

The blocker is that this integration has no static entity list. Every entity comes out of discovery, discovery only runs inside a successful poll, and all ten platforms enumerate `coordinator.bound` exactly once at forward time. Measured on this branch — offline setup, then four summary intervals:

| | |
| --- | --- |
| entry state | `LOADED` |
| `coordinator.bound` | 0 |
| entities in the state machine | 0 |
| coordinator listeners | 0 |
| poll attempts over the next 4 intervals | **0** |

The entry loads empty, and because `DataUpdateCoordinator` only re-arms its timer while it has listeners, it then never polls again — so the device coming back online doesn't recover it. The full suite passed because the two tests guarding that path were the ones the diff rewrote.

The comparison to ESPHome/Shelly/LIFX/WLED in #295 is fair, and it's what pointed at the fix: those integrations can load offline because each has a persisted device description to build entities from. #374 adds the equivalent here — it banks the resource dict each successful first cycle feeds discovery and replays it when the device is unreachable, plus a reconcile-and-reload for when the live set later disagrees, and a permanent coordinator listener so polling continues with zero live entities.

One thing worth flagging since it's the gap between this and what you asked for: restored entities are `unavailable` rather than showing their last recorded values. Giving them values means asserting state we can't verify — a washer unplugged for a week would read "Running". Recorded history is kept by the recorder either way, so graphs are unaffected.

---

Issue #302, opened 2026-08-05
Legacy ARTIK051 AC reports no energy at all (correctly) — and why the capability bit must not be used to gate the sensor


perseus177:
`sensor.*_energy` reads a permanent `0.0` on `ARTIK051_KRAC_18K`. **The zero is correct** — the appliance genuinely does not meter power — so this is not a parsing bug report. It is about whether the integration should be offering the entity at all on such a unit, and about why the obvious way to decide that is wrong. There is a counterexample sitting in this repository's own fixtures.

*Disclosure, as in #146, #168, #290, #296–#299 and #301: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

Two `ARTIK051_KRAC_18K` units, `OptionCode_35882` / `ExtendOptionCode_7`.

## The zero is the appliance's own answer

```
/energy/consumption/vs/0   ->  {"x.com.samsung.da.cumulativePower": "0"}      (3 of 3 reads, both units)
/energy/consumption/0      ->  {}                                            (empty on both)
```

And the appliance's own app never shows kWh for these units. It decides that from a capability bit:

```js
$scope.isRunningTimeGraph = !isPowerConsumptionSupported();   // rac.js
function isPowerConsumptionSupported() { return $scope.racOptionCodeValue[1] === "1"; }
```

`OptionCode_35882` is `1000110000101010`, so index `[1]` is `0` and the app draws a **running-time** graph instead, in hours. What these units record in place of energy, and the format it is in, is #301.

**Positive control, in your fixtures:** `airconditioner_artik051_krac_energy_device.json` is the *same model*, but `OptionCode_56378` → `[1]` is `1`, and it reports `cumulativePower: 117430000` (1174.30 kWh with the existing `/100000` legacy divisor from #193). Same model, one bit apart, and the bit predicts which one meters. `airconditioner_tp1x_rac_odor_controller_device.json` has `OptionCode_35880` → `[1]` = 0 and carries no `cumulativePower` field at all. Three for three.

## Why gating the entity on that bit would be a mistake

That is the change I was about to propose, and your fixtures talked me out of it:

| fixture | model | legacy board | OptionCode | `[1]` | cumulativePower |
|---|---|---|---|---|---|
| `airconditioner_artik051_krac_18k` | ARTIK051_KRAC_18K | yes | 35882 | 0 | `0` |
[...]

Issue #301, opened 2026-08-05
Legacy ARTIK051 AC: /file/transfer/vs/0 holds a decoded runtime-hours history, unreachable through batch discovery


perseus177:
Legacy ARTIK051 air conditioners keep a **runtime-hours history** on the appliance, at a resource this integration cannot currently reach. This is not a feature request with a PR behind it — it is the reverse engineering written down, plus the one architectural question that decides whether it can ever be an entity. I built it, it worked, and then I threw the working version away because it only worked *outside* the integration, which is not something I want to keep running.

*Disclosure, as in #146, #168, #290, #296, #297, #298 and #299: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

Measured on two `ARTIK051_KRAC_18K` units (`192.168.0.121`, `192.168.0.166`), firmware token set `OptionCode_35882` / `ExtendOptionCode_7`.

## The numbers in Samsung's own app are hours, not kWh

The owner read four monthly figures off the app and took them for consumption. They are running time. Four independent reasons, and they all agree:

**1. The app decides this from a capability bit.** In its RAC plugin:

```js
$scope.isRunningTimeGraph = !isPowerConsumptionSupported();   // rac.js, ~4572
function isPowerConsumptionSupported() { return $scope.racOptionCodeValue[1] === "1"; }
```

`OptionCode_35882` is `1000110000101010`, so `[1]` is `0` on these units: the app draws a **running-time** graph, labelled with `WEBMOB_common_hours`. That is also why `sensor.*_energy` reads a permanent `0.0` here — `/energy/consumption/vs/0.cumulativePower` really is `0` on this board generation (a reporter on another family has `117430000` there, i.e. 1174.30 kWh). The zero is correct, not a parsing bug.

**2. The appliance's own history reproduces the app's figures exactly.** See the format below; `raw / 10` gives hours, and month-over-month differences land on the app's numbers (bedroom May `0.0`, bedroom August `8.5`, living room May `67.5`, all exact).

**3. Where they disagree, they disagree by equal and opposite amounts, and the sums match.** Bedroom June + July: the app says 282.0, the file says 282.0, but split differently between the two months. The app does not bucket by calendar month — it counts from a meter date (`$scope.currentMeterDate`), which is a user setting.

[...]

PR #304, opened 2026-08-05
feat(climate): derive legacy AC presets from the unit's own capability bits, per HVAC mode

perseus177:
> **Stacked on #299.** The first commit here is that PR's; only the second one (`feat(climate): offer legacy presets per HVAC mode…`) belongs to this review. It needs #299's `_LEGACY_SLEEP_PRESET_CODES`, so I did not try to untangle them — say the word once #299 lands and I will rebase this onto `main` so the diff is just the one commit.

This is the answer to your review on #298: **the list can come from the appliance after all** — not as names, which it never publishes, but as capabilities, which it does. So `_LEGACY_PRESET_CODES` stops being a list every legacy board gets and becomes a fallback, and what a given unit offers is derived from that unit's own bits plus the HVAC mode it is in.

*Disclosure, as in #146, #168, #290, #296–#299, #301 and #302: this account is @perseus177's — he owns the appliances and reads everything before it goes out — but the measuring and the writing here are Claude Code's.*

## First, what these boards do not publish

I checked rather than assumed, on an `ARTIK051_KRAC_18K`:

```
/oic/res                     -> the only mode hrefs are /mode/0 and /mode/vs/0
GET /mode/convenient/vs/0    -> empty payload; the resource does not exist here
/mode/vs/0 supportedModes    -> ['Cool','Dry','Wind','Auto','Heat','HOMECARE_WIZARD_V2']
                                HVAC modes only, no Comode_* entries
/mode/vs/0 options           -> Comode_Off ... one token, the active one, never a list
/configuration/vs/0          -> {"x.com.samsung.da.region": "0000000000"}, no airconOptionList
```

That absence is why `_legacy_convenient()` fabricates a convenient-shaped rep at all. A name list is not obtainable: the mapping from a capability to its token spelling exists only in the vendor's app.

## What they do publish

Two bit maps, in that same options array: `OptionCode` (16 wide) and `ExtendOptionCode` (32). The app reads them by padding the number to binary and indexing the string, and gates each comfort mode on a bit *plus* the current mode, in `updateOptionsList()`. This PR is that function transcribed, with the bit each rule reads named in a comment beside it:

| item | the app's rule | this unit |
|---|---|---|
| WindFree | show on `eoc[31]`; disable in Heat, in AIComfort, and in Auto unless 18K (`eoc[30]`) | Cool/Dry/Fan/Auto ✓, Heat ✗ |
| Fast Turbo | Cool/AIComfort always; Heat only on `oc[12]` | both ✓ |
| 2-Step | Cool/AIComfort only | ✓ |
| Comfort | Cool/Heat/AIComfort, hidden in Heat when `!oc[12]` | both ✓ |
| d'light Cool | `Cool && oc[2]` | never — `oc[2]` is 0 |
| Single User | `oc[3] \|\| (oc[11] && Heat)` | never — both 0 |
| Quiet | `oc[10]` | ✓ |
| Good Sleep | Cool/Heat/AIComfort only | ✓ |

`option_bit()` / `option_code_bit()` / `extend_option_code_bit()` land in `capabilities/airconditioner.py`, using the app's own indexing so the two can be compared line by line. They return `None` when the token is absent — "the bit is 0" and "this board publishes no map" are different facts, and the second one must not silently mean "supports nothing".

## Does it match reality

The owner read the available comfort modes off both his remote and the app, per HVAC mode, before I wrote any of this. What the code derives, printed from the code itself:

```
Cool   -> none, nano, speed, 2step, quiet, comfort, smart, sleep, nanosleep
          owner: 2-Step, Fast Turbo, Comfort, Quiet, WindFree
Heat   -> none, speed, quiet, comfort, sleep, nanosleep
          owner: Fast Turbo, Comfort, Quiet — WindFree cannot be turned on
Dry    -> none, nano          owner: nothing else, and WindFree keeps the mode
Fan    -> none, nano          owner: nothing else, and WindFree keeps the mode
Auto   -> none, nano          owner: nothing else; the app reaches WindFree by switching to Cool
```

All five agree. And the appliance backs the two interesting cases directly: `["Comode_Nano"]` written while the unit is in Auto is answered `2.04 Changed` and **dropped** (still `Comode_Off` at +8s and +45s), while the same token in Cool holds.

## The write ordering, which is measured and not obvious

Putting `modes: ["Cool"]` and the token in **one** POST does not work — the mode moves and the token is still dropped, so the board judges the option against the mode it was in. Sent as its own write first, it holds. How much of a gap that needs was measured rather than guessed:

| gap between the two writes | result at +8s and +38s |
|---|---|
| 0 s (back to back) | `Comode_Off` — dropped |
| 2 s | `Comode_Nano` — holds |
| 5 s | `Comode_Nano` — holds |

Hence `_NANO_AFTER_MODE_DELAY = 3`, and it only ever runs for this one preset in this one mode.

## Three deliberate safety rails

1. **No bit map → the old unconditional tuple.** Boards that publish no `OptionCode` keep exactly today's behaviour. This is the reason `_LEGACY_PRESET_CODES` survives.
2. **Whatever the unit is running is always listed**, even where the rules deny it. A remote can put a board in a state this table would not offer — measured: `Comode_2Step` was accepted and held in Auto, where neither the remote nor the app offers it — and a `preset_mode` outside `preset_modes` is not a state HA allows.
3. **Smart Saver cannot be derived, and I would rather say so than bury it.** It has no bit of its own, and the app hides it from every single RAC outright (`showSaverOption = false`) — yet the appliance accepts `Comode_Smart` and does what Samsung documents: `min_temp` 16 → 24 and a 22 target lifted to 24, both fields the *device* reports, both reverting on `Comode_Off`. So it is added in Cool on evidence, not on a bit. **This is the one judgement call in the PR** — if you would rather not carry a preset that no vendor UI offers, delete those three lines and everything else stands.

**#298 becomes redundant either way** and I am happy for you to close it: Smart arrives through this path instead, and the tuple it edited is left untouched here.

## One place I trust the hardware over the vendor's code

`Quiet` is gated on `oc[10]` alone in the JS, with no mode condition, which would put it in Dry/Fan/Auto too. The owner sees it in Cool and Heat only, on the remote as well as in the app. I kept the mode gate — flagging it because it is the one rule that is an observation rather than a transcription. (Their `// Nothing` fall-through for the other modes looks like a bug, which may be the whole explanation.)

## Shape

- `option_bit` + two wrappers in `capabilities/airconditioner.py`; `_legacy_preset_codes()` and `_legacy_preset_needs_cool()` in `climate.py`.
- Six new tests: the per-mode table, a cleared bit removing its presets, a board with no maps keeping the old list, the active code surviving the rules, and both write paths. 1232 pass, `ruff format --check` and `ruff check` clean.
- No new translation strings: `smart`, `nano` and `nanosleep` are already in all six catalogs from the newer boards.

### DIFF
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -81,6 +81,8 @@ from .registry.capabilities.airconditioner import (
 )
 from .registry.capabilities.airconditioner import (
     extend_option_code_bit,
+    has_extend_option_code,
+    has_option_code,
     is_legacy_board,
     option_code_bit,
 )
@@ -282,6 +284,13 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
     # token these codes come with.
     _LEGACY_SLEEP_PRESET_CODES = ("Sleep", "NanoSleep")
 
+    # HVAC modes _legacy_preset_codes() has a rule for. Anything else is a mode
+    # this transcription has never seen, which is the same "cannot judge" case as
+    # a board that publishes no capability map -- and gets the same fallback.
+    _LEGACY_KNOWN_HVAC = frozenset(
+        {"Cool", "Heat", "HeatClean", "Dry", "Fan", "Wind", "Auto", _AI_COMFORT_MODE}
+    )
+
     # Which comfort modes a legacy board offers in which HVAC mode, and which of
     # them it has at all. Both come from the appliance rather than from a table
     # per model: the unit publishes its capabilities as two bit maps in
@@ -292,41 +301,65 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
     #
     # Confirmed against an ARTIK051_KRAC_18K whose owner read the same lists off
     # the remote and the app: WindFree in Cool/Dry/Fan and (being an 18K model)
-    # Auto but never Heat, Fast Turbo and Comfort in Heat because oc[12] is set,
-    # and no d'light or Single User anywhere because oc[2], oc[3] and oc[11] are
-    # zero -- the appliance refuses both of those locally too.
+    # Auto but never Heat, and Fast Turbo and Comfort in Heat because oc[12] is
+    # set. d'light Cool is gated the same way on oc[2], which is zero here -- and
+    # the appliance refuses the token locally too.
+    #
+    # Single User is deliberately not modelled, though the app does gate it on
+    # oc[3] / oc[11]: it has no token of its own. The app's own Single User
+    # command sends `Comode_Smart` -- the Smart Saver token -- with a hardcoded
+    # 24 desired alongside it, so there is nothing to write that would be
+    # distinguishable from the Smart preset below, and no name to give it that
+    # the appliance would recognise.
 
     def _legacy_preset_codes(self, hvac: str, options: list) -> list[str]:
-        """Comfort-mode codes this unit offers in this HVAC mode."""
+        """Comfort-mode codes this unit offers in this HVAC mode.
+
+        Derived only for boards that publish *both* capability maps. One map on
+        its own is not enough: every eoc-gated rule would then read None, and
+        None means "this board does not publish the map", never "the feature is
+        absent". The FAC/CAC boards on record carry only the older map, with
+        values small enough that RAC bit positions all read as zeros, so
+        requiring both also keeps these rules inside the family they were
+        documented for.
+
+        Within the derived path, a bit that cannot be read (a malformed or
+        over-wide token) is treated as permission rather than denial for the
+        codes the unconditional list already carried -- losing a working preset
+        to a parsing failure is worse than offering one too many. Codes that were
+        never in that list (d'light) still need their bit to be explicitly set.
+        """
         rep = self._rep(MODE_HREF)
-        cool = hvac in ("Cool", _AI_COMFORT_MODE)
-        heat = hvac in ("Heat", "HeatClean")
-        # An absent bit map is not a claim that nothing is supported, so a board
-        # that publishes no OptionCode keeps the older unconditional list.
-        if option_code_bit(rep, 1) is None and extend_option_code_bit(rep, 31) is None:
+        if (
+            not has_option_code(rep)
+            or not has_extend_option_code(rep)
+            or hvac not in self._LEGACY_KNOWN_HVAC
+        ):
             return list(self._LEGACY_PRESET_CODES)
 
+        cool = hvac in ("Cool", _AI_COMFORT_MODE)
+        heat = hvac in ("Heat", "HeatClean")
         codes = ["Off"]
-        # WindFree: shown on eoc[31], disabled in Heat, in AIComfort, and in Auto
-        # unless this is an 18K model (eoc[30]) -- the app switches such a unit to
-        # Cool instead, which is what _legacy_preset_needs_cool does.
+        # WindFree: shown on eoc[31]; disabled in Heat, in AIComfort, and in Auto
+        # unless this is an 18K model (eoc[30]), where the app switches to Cool for
+        # it instead -- which is what _legacy_preset_needs_cool does.
         nano_mode_ok = not heat and hvac != _AI_COMFORT_MODE
         if hvac == "Auto":
-            nano_mode_ok = bool(extend_option_code_bit(rep, 30))
-        if extend_option_code_bit(rep, 31) and nano_mode_ok:
+            nano_mode_ok = extend_option_code_bit(rep, 30) is not False
+        if extend_option_code_bit(rep, 31) is not False and nano_mode_ok:
             codes.append("Nano")
-        if cool or (heat and option_code_bit(rep, 12)):  # Fast Turbo, oc[12] in Heat
+        if cool or (heat and option_code_bit(rep, 12) is not False):  # Fast Turbo
             codes.append("Speed")
         if cool:
             codes.append("2Step")
-        if cool and option_code_bit(rep, 2):  # d'light Cool
+        if cool and option_code_bit(rep, 2):  # d'light Cool -- needs the bit set
             codes.append("DlightCool")
         # Quiet reads oc[10] with no mode condition in the app, but the owner of
         # the unit above sees it in Cool and Heat only, on the remote as well as
         # in the app -- the observation wins over the reading.
-        if option_code_bit(rep, 10) and (cool or heat):
+        if option_code_bit(rep, 10) is not False and (cool or heat):
             codes.append("Quiet")
-        if (cool or heat) and not (heat and not option_code_bit(rep, 12)):  # Comfort
+        if cool or (heat and option_code_bit(rep, 12) is not False):  # Comfort
             codes.append("Comfort")
         # Smart Saver has no bit of its own and the app hides it from every single
         # RAC outright (showSaverOption = false), yet the appliance accepts it and
@@ -354,13 +387,11 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
         if active is None:
             return {}
 
-        codes = self._legacy_preset_codes(
-            str(_first(self._rep(MODE_HREF).get(_MODES_FIELD))), options
-        )
+        codes = self._legacy_preset_codes(_first(self._rep(MODE_HREF).get(_MODES_FIELD)), options)
         # Whatever the unit is actually running has to be listed whether the
-        # table expects it there or not -- a preset_mode outside preset_modes is
+        # rules expect it there or not -- a preset_mode outside preset_modes is
         # not a state HA allows, and the appliance has the last word on what it
-        # is doing (a remote can put it in a mode this table would not offer).
+        # is doing (a remote can put it in a mode these rules would not offer).
         if active not in codes:
             codes.append(active)
         return {_MODES_FIELD: [active], _SUPPORTED_FIELD: codes}
@@ -716,8 +747,13 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
         Comode_Off at +8s and +45s), and putting `modes: Cool` in the *same* POST
         does not help -- the mode moves and the token is still dropped, so the
         board judges the option against the mode it was in. Sent as its own write
-        first, it holds. The appliance's own app does the same thing for the same
-        reason (its nano command carries `modes: Cool` only in Auto/AIComfort).
+        first, it holds. The appliance's own app pairs `modes: Cool` with its nano
+        command for the same reason.
+
+        Auto only. The app's builder also covers AIComfort, but its
+        `updateOptionsList()` disables the WindFree button there outright, so that
+        pairing can never fire -- and `_legacy_preset_codes()` likewise does not
+        offer `Nano` in AIComfort, which would leave such a branch unreachable.
 
         The pause is measured, not padding: back to back (same session, no gap at
         all) the token was dropped again, two seconds apart it held. Three is that
@@ -726,9 +762,13 @@ class LocalThingsClimate(LocalThingsEntity, ClimateEntity):
         """
         if not self._legacy_preset() or code != "Nano":
             return
-        if _first(self._rep(MODE_HREF).get(_MODES_FIELD)) not in ("Auto", _AI_COMFORT_MODE):
+        if _first(self._rep(MODE_HREF).get(_MODES_FIELD)) != "Auto":
             return
-        await self.coordinator.async_send_command(self._bound, ("mode", "Cool"))
+        # Same resolver the rest of the platform uses -- the device code for an HA
+        # mode is read off the unit's own supportedModes rather than assumed.
+        await self.coordinator.async_send_command(
+            self._bound, ("mode", self._device_code_for_hvac(HVACMode.COOL))
+        )
         await asyncio.sleep(_NANO_AFTER_MODE_DELAY)
 
     async def async_set_preset_mode(self, preset_mode: str) -> None:
--- a/custom_components/localthings/climate.py
+++ b/custom_components/localthings/climate.py
@@ -21,6 +21,7 @@ temperature and wind resources alike.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 
 from homeassistant.components.climate import (
@@ -79,7 +80,9 @@ from .registry.capabilities.airconditioner import (
     HREF_WIND_STRENGTH as WIND_STRENGTH_HREF,
 )
 from .registry.capabilities.airconditioner import (
+    extend_option_code_bit,
     is_legacy_board,
+    option_code_bit,
 )
 from .registry.capabilities.common import normalize_temp_unit
 from .registry.entities import ClimateDesc
@@ -130,6 +133,10 @@ PRESET_AI_COMFORT = "ai_comfort"
 # with an echoed capability flag, not a genuine mode. Unlike _AI_COMFORT_MODE,
 # not modeled as a preset either: nothing confirms it's user-selectable.
 _NON_HVAC_OPTION_CODES = frozenset({"HOMECARE_WIZARD_V2"})
+
+# Seconds to let a legacy board settle into Cool before the WindFree token is
+# written after it -- see _legacy_preset_needs_cool for the measurement.
+_NANO_AFTER_MODE_DELAY = 3
 
 # Fan (wind strength): device codes "0".."4" -> HA standard fan constants where
 # a clean match exists so they auto-localize; "turbo" is custom (translated).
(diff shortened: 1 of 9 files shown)


custom_components/localthings/climate.py
```diff
@@ -267,17 +274,96 @@ def __init__(self, coordinator: LocalThingsCoordinator, bound) -> None:
...
+    # zero -- the appliance refuses both of those locally too.
+
+    def _legacy_preset_codes(self, hvac: str, options: list) -> list[str]:
+        """Comfort-mode codes this unit offers in this HVAC mode."""
+        rep = self._rep(MODE_HREF)
+        cool = hvac in ("Cool", _AI_COMFORT_MODE)
+        heat = hvac in ("Heat", "HeatClean")
+        # An absent bit map is not a claim that nothing is supported, so a board
+        # that publishes no OptionCode keeps the older unconditional list.
+        if option_code_bit(rep, 1) is None and extend_option_code_bit(rep, 31) is None:
```
mbillow: **This guard needs both maps missing, and one legacy board in the repo publishes only one.**

A board with `OptionCode` but no `ExtendOptionCode` sails past this check, and from there every `extend_option_code_bit()` read returns `None` → falsy → treated as "feature absent". That's exactly the conflation `option_bit`'s own docstring says must not happen: "no token at all says only that this board does not publish the map."

`tests/fixtures/airconditioner_artik051_dongle_fac_18k_device.json` is that board — `OptionCode_521`, no `ExtendOptionCode`, fixture mode `Auto`:

| | `preset_modes` |
|---|---|
| `main` | `none, nano, quiet, comfort, 2step, speed` |
| this branch | `none` |

WindFree is gone in *every* HVAC mode for that unit, and Auto/Dry/Wind collapse to `['none']` on its own. Rail 2 doesn't catch it either — the active code is `Comode_Off`, which is already in the list.

Suggest applying the guard per map rather than to the pair, so a missing `ExtendOptionCode` means "can't judge this, don't hide it" instead of 0. Something like an `option_code_bit(...) is not False` test for the eoc-gated rules, or a per-map presence helper.

Worth a test on this fixture too — nothing currently asserts its preset list, which is why all 1232 still pass.

---

custom_components/localthings/climate.py
```diff
@@ -267,17 +274,96 @@ def __init__(self, coordinator: LocalThingsCoordinator, bound) -> None:
...
+        if cool:
+            codes.append("2Step")
+        if cool and option_code_bit(rep, 2):  # d'light Cool
+            codes.append("DlightCool")
+        # Quiet reads oc[10] with no mode condition in the app, but the owner of
+        # the unit above sees it in Cool and Heat only, on the remote as well as
+        # in the app -- the observation wins over the reading.
+        if option_code_bit(rep, 10) and (cool or heat):
+            codes.append("Quiet")
+        if (cool or heat) and not (heat and not option_code_bit(rep, 12)):  # Comfort
```
mbillow: This is the same condition as `Speed` seven lines up, written a different way. `(cool or heat) and not (heat and not oc12)` ≡ `cool or (heat and oc12)` across every reachable input — they differ only at `cool and heat`, which can't happen since the two are built from disjoint mode sets.

Given that the whole point of this function is line-by-line comparability with `updateOptionsList()`, the two rules being identical in the app but spelled differently here is the kind of thing that costs a reader a minute each time. Suggest matching the `Speed` form.

---

custom_components/localthings/climate.py
```diff
@@ -267,17 +274,96 @@ def __init__(self, coordinator: LocalThingsCoordinator, bound) -> None:
...
+    # /mode/vs/0's options (OptionCode, ExtendOptionCode), and its own app gates
+    # each item on a bit plus the current mode. _legacy_preset_codes() below is
+    # that logic, transcribed from the app's updateOptionsList() so the two can be
+    # compared line by line, with the bit each rule reads named in the comment.
+    #
+    # Confirmed against an ARTIK051_KRAC_18K whose owner read the same lists off
+    # the remote and the app: WindFree in Cool/Dry/Fan and (being an 18K model)
+    # Auto but never Heat, Fast Turbo and Comfort in Heat because oc[12] is set,
+    # and no d'light or Single User anywhere because oc[2], oc[3] and oc[11] are
+    # zero -- the appliance refuses both of those locally too.
```
mbillow: **Single User is documented here and in the PR table, but never implemented.**

This comment says `oc[3]` and `oc[11]` are zero, and the PR body gives the rule as `oc[3] || (oc[11] && Heat)` — but nothing in `_legacy_preset_codes` reads bit 3 or bit 11, and no code path ever emits a `SingleUser` code.

That makes `assert "singleuser" not in presets` in `test_presets_follow_the_hvac_mode_and_the_capability_bits` vacuously true: it would pass no matter what those bits said, so it isn't testing the rule it looks like it's testing. Either implement the rule (and the test starts earning its keep) or drop the claim from the comment and the assertion from the test.

---

custom_components/localthings/climate.py
```diff
@@ -267,17 +274,96 @@ def __init__(self, coordinator: LocalThingsCoordinator, bound) -> None:
...
+        if hvac == "Auto":
+            nano_mode_ok = bool(extend_option_code_bit(rep, 30))
+        if extend_option_code_bit(rep, 31) and nano_mode_ok:
+            codes.append("Nano")
+        if cool or (heat and option_code_bit(rep, 12)):  # Fast Turbo, oc[12] in Heat
+            codes.append("Speed")
+        if cool:
+            codes.append("2Step")
+        if cool and option_code_bit(rep, 2):  # d'light Cool
+            codes.append("DlightCool")
```
mbillow, line 356: `DlightCool` has no translation entry. It goes through `_preset_to_ha` to `"dlightcool"`, and none of the six catalogs have that key — a board with `oc[2]` set would show the raw code in the UI.

This unit has the bit clear so it never surfaces here, but the rule is live for any board that sets it. (`smart` is fine — already present in all six.) Same would apply to `singleuser` if that rule gets implemented.

---

custom_components/localthings/registry/capabilities/airconditioner.py
```diff
@@ -324,6 +359,59 @@ def write(payload, rep, href=None):
...
+    as `Comode_NanoSleep`/`Sleep_16`, silently turning the user's two hours into
+    eight. Writing the pair keeps the two hours.
+    """
+    sleep = _option_token(rep, "Sleep")
+    running = sleep not in (None, "0")
+    if code == "Nano" and running:
+        code = "NanoSleep"
+    if code in ("Sleep", "NanoSleep"):
+        return [f"Comode_{code}", f"Sleep_{sleep if running else _DEFAULT_SLEEP_HALF_HOURS}"]
+    return option_write("Comode", code)
```
mbillow, line 436: **Leaving a sleep mode via a preset doesn't zero `Sleep_`, but the Number does — and the stale token then feeds back into this function.**

`_good_sleep_write(0, …)` sends `["Comode_Off", "Sleep_0"]`, while selecting the `none` preset lands here and sends `["Comode_Off"]` alone. If the board doesn't zero the token by itself, the next nano selection reads `running=True` off that leftover value:

```
preset Off  while sleep running  -> ['Comode_Off']
then preset Nano (stale Sleep_4) -> ['Comode_NanoSleep', 'Sleep_4']
```

So a plain WindFree selection silently re-arms a two-hour timer the user just turned off — the same class of surprise this function exists to prevent, just from the other direction.

You have the unit, so this is one measurement: does a bare `Comode_Off` zero `Sleep_`? If it does, no change needed and it's worth a line in the docstring saying so. If it doesn't, the preset path should carry `Sleep_0` when leaving `Sleep`/`NanoSleep`.

---

custom_components/localthings/climate.py
```diff
@@ -622,6 +708,29 @@ async def async_set_swing_mode(self, swing_mode: str) -> None:
...
+        reason (its nano command carries `modes: Cool` only in Auto/AIComfort).
+
+        The pause is measured, not padding: back to back (same session, no gap at
+        all) the token was dropped again, two seconds apart it held. Three is that
+        with a little margin, and it only ever runs for this one preset in this
+        one HVAC mode.
+        """
+        if not self._legacy_preset() or code != "Nano":
+            return
+        if _first(self._rep(MODE_HREF).get(_MODES_FIELD)) not in ("Auto", _AI_COMFORT_MODE):
```
mbillow: The `_AI_COMFORT_MODE` half of this tuple is unreachable. In AIComfort, `nano_mode_ok` is False, so `Nano` never lands in the codes list, so `async_set_preset_mode("nano")` never finds it in `_supported(CONVENIENT_HREF)` and never reaches this helper. Confirmed — AIComfort presets come out as `none, speed, 2step, quiet, comfort, sleep, nanosleep`.

The docstring says the app carries `modes: Cool` "in Auto/AIComfort", which suggests the intent was to offer nano in AIComfort too and reach it the same way. Either do that (drop `hvac != _AI_COMFORT_MODE` from `nano_mode_ok`) or drop AIComfort from this check, so the two halves agree on whether it's reachable.

---

custom_components/localthings/climate.py
```diff
@@ -267,17 +274,96 @@ def __init__(self, coordinator: LocalThingsCoordinator, bound) -> None:
...
-        return {}
+        active = next(
+            (o.split("_", 1)[1] for o in options if isinstance(o, str) and o.startswith("Comode_")),
+            None,
+        )
+        if active is None:
+            return {}
+
+        codes = self._legacy_preset_codes(
+            str(_first(self._rep(MODE_HREF).get(_MODES_FIELD))), options
```
mbillow: Minor: `str(_first(...))` turns a missing `modes` field into the literal string `"None"`, which matches no branch in `_legacy_preset_codes` — so an unknown mode degrades to a derived-but-wrong list (`['none', 'nano']` on this fixture) rather than falling back to `_LEGACY_PRESET_CODES` the way an absent bit map does.

Needs a partial or malformed rep to hit, so low priority, but the safe-fallback instinct that governs the rest of this function seems worth extending to it — an unrecognized mode is the same kind of "can't judge" as an absent map.

---

mbillow (changes requested):
Reviewed both commits — we'll close #299 and take its change through here, so this covers the full diff.

The reverse-engineering is excellent, and deriving the list from the unit's own bits is clearly the right answer to my note on #298. The three safety rails are the right instincts too. But the first rail has a hole, and a fixture already in the repo falls through it.

## Blocking

**The absent-map guard is all-or-nothing.** It requires *both* maps missing, so a legacy board that publishes `OptionCode` but no `ExtendOptionCode` gets past it and then has every eoc-gated feature treated as unsupported. `airconditioner_artik051_dongle_fac_18k` is exactly that board, and it goes from six presets to `['none']` in its own fixture mode, losing WindFree in every mode. Details inline.

The reason CI is still green on this is that nothing asserts that fixture's preset list — worth adding alongside the fix.

## Worth fixing

- `Comfort` and `Speed` are the same condition written two different ways, which undercuts the line-by-line comparability the function is built for.
- Single User is documented in the comment and the PR table but never implemented, which leaves its test assertion vacuous.
- `DlightCool` can be emitted but has no catalog entry.
- Leaving a sleep mode via a preset doesn't zero `Sleep_` while the Number does, and the stale token feeds back into `_preset_options` — one measurement on the unit settles whether this is real.

## Minor

- The AIComfort branch of `_legacy_preset_needs_cool` is unreachable.
- An unknown HVAC mode degrades to a derived-but-wrong list instead of the safe fallback.
- `option_code_bit(rep, 1)` as a bare presence probe would read better as a named helper.
- `_legacy_preset_needs_cool` hardcodes `"Cool"` where the module elsewhere uses `_device_code_for_hvac` precisely because codes vary per unit. Cosmetic for legacy boards, but inconsistent.
- A preset change silently moving hvac_mode Auto → Cool matches the app, but it will surprise people in HA — worth a changelog line.

## Verified

1232 pass; `ruff format --check` and `ruff check` clean over `custom_components` and `tests`. Bit indices all check out (`35882` → oc[12]=1, oc[10]=1, oc[2]=0; `ExtendOptionCode_7` → eoc[31]=eoc[30]=1), `option_bit` correctly rejects over-wide and negative values, there's no recursion between `_rep(CONVENIENT_HREF)` and `_legacy_convenient`, and the active-code rail behaves. `artik051_krac_energy` — the other legacy board carrying both maps — comes out sensibly across all five modes.

## On the judgement call

Keep `Smart`. You have direct device evidence of `min_temp` and the target moving and reverting, which beats the app's `showSaverOption` flag as a statement about the hardware, and Cool-only keeps the blast radius small. Happy to close #298 as you suggest.

---

perseus177 (commented):
Thank you for finding the hole — that fixture would have shipped broken. All nine points are addressed: the code in `eed04fa`, and the one that needed the hardware is measured, with the result in `b820a96`. Details below.

## Blocking: fixed, but not the way you suggested

Your diagnosis was right and I went further than the per-map guard, because that alone still breaks that board. `airconditioner_artik051_dongle_fac_18k` has `oc[10]` set to an **explicit 0**, so treating only `None` as permissive still removes Quiet from it — and its `OptionCode` is **521**, three orders of magnitude below every RAC-class value these bit positions were read from (35880–56636). The FAC and CAC dumps in the repo all sit in that small range, and the function those positions come from is the RAC plugin's, with its author guarding neighbours with *"only for Single RAC"*. Reading RAC bit offsets out of a FAC field is meaningless; it just happens to yield zeros.

So the derived path now requires **both** maps, via named helpers (`has_option_code` / `has_extend_option_code`, which also replaces the `option_code_bit(rep, 1)` presence probe you flagged). `ExtendOptionCode` calls itself *"Single RAC new option code, as old option code is full"*, so its presence is the closest available proxy for "this is the family those offsets were documented for" — a proxy, stated as one in the docstring, and used only to decide whether to read the map at all.

Inside the derived path a bit that cannot be read is still treated as permission rather than denial, for the codes the unconditional list already carried. A parsing failure should not cost a working preset.

Tests: the dongle fixture is asserted to keep exactly `main`'s six presets across five HVAC modes, and `artik051_krac_energy` — the other board carrying both maps — is asserted to *stay* on the derived path, so the guard cannot quietly swallow everything.

## Worth fixing

- **Comfort** now reads `cool or (heat and oc12)`, identical in form to `Speed`.
- **Single User** claim dropped from the comment, and the vacuous assertion removed. Worth recording why it cannot simply be implemented: the app's own `getSingleUserCommand()` sends **`Comode_Smart`** — the Smart Saver token — with a hardcoded `desired: 24` beside it. There is no distinct token to write, and no name the appliance would recognise. And this is not academic: on `artik051_krac_energy` those bits are actually **set** (`oc[3]`=1, `oc[11]`=1), so implementing the rule with an invented code would have handed that board a preset that does nothing.
- **DlightCool** now has an entry in all six catalogs (`d'light Cool` — Samsung's own spelling, and the same string in each, being a product name).
- **Sleep_ on leaving a sleep preset: measured, and the board zeroes it itself** — so no code change, just the docstring saying so (`b820a96`). Starting from a verified `Comode_Sleep` / `Sleep_4`, a bare `["Comode_Off"]`:

  ```
  pred   Comode_Sleep  Sleep_4   power=On
  WRITE  ['Comode_Off']          2.04 Changed
  +8s    Comode_Off    Sleep_0
  +38s   Comode_Off    Sleep_0
  ```

  So the `none` preset cannot leave a stale duration for the next nano selection to read as a running timer. Worth adding that the first attempt at this measurement failed and reported rather than answering: the host holding the DTLS client had rebooted, its static ARP entry for the unit was gone, and every write came back a handshake timeout — the script's precondition gate refused to measure and said so, which is the only reason there is no plausible-looking wrong number in this comment.

## Minor

- The `_AI_COMFORT_MODE` half of the check is gone. The app's command builder does cover AIComfort, but `updateOptionsList()` disables the WindFree button there, so the pairing can never fire — and `_legacy_preset_codes()` does not offer `Nano` there either. The docstring now states that, and a test asserts AIComfort offers no nano and writes no mode change.
- An unknown HVAC mode falls back to `_LEGACY_PRESET_CODES`, alongside the absent-map case, via a `_LEGACY_KNOWN_HVAC` set. I also dropped the `str()` that was turning a missing `modes` field into the literal `"None"`.
- `_legacy_preset_needs_cool` now resolves the code through `_device_code_for_hvac(HVACMode.COOL)` like the rest of the platform.
- **Changelog:** there is no changelog file in the repo, so flagging it here for the release notes — *selecting WindFree while a legacy 18K unit is in Auto now switches it to Cool, because the appliance will not accept WindFree in Auto.*

## Two corrections to what I told you

- I offered to **rebase this onto `main` once #299 landed**. Since you closed #299 in favour of this PR, that rebase would delete its change — so the first commit stays where it is. Happy to squash the three into one, or reorder, if you would rather review it differently.
- I nearly told you this push was CI-verified only, because the machine I run the suite on was down when I pushed it. It came back: **1236 pass locally**, plus `ruff format --check` and `ruff check` clean, on top of the green CI you can see.

Thank you also for keeping `Smart` on the device evidence — and for taking #299's change through here.

PR #316, opened 2026-08-06
feat(airconditioner): add support for ACA-KR-TP2-21-AN9000 (updated for v0.19.0)

teia1397-prog:
## Description
I have updated my custom changes to support the Samsung `ACA-KR-TP2-21-AN9000` model so that they are fully compatible with **v0.19.0**.

This PR adds the missing capabilities required for this specific device in `by_type_airconditioner.py`.

## Changes
- **Updated for v0.19.0 compatibility**
- **AC Capabilities**: Added `AI_CLEAN`, `WINDFREE`, and `WINDSLEEP`
- **Purifier Capabilities**: Reused `air_purifier.HEPA_FILTER` and `air_purifier.DEVICE_ACTIVE`
- **Standalone Power**: Included `common.POWER`

## Request
Could you please review and merge this support for `ACA-KR-TP2-21-AN9000` so it can be included in the next release? 

Thank you!


mbillow:
Thanks for putting this together and for the detailed writeup on what you found for the System Fresh Air Ventilator (ACA-KR-TP2-21-AN9000) — genuinely useful device intel.

I'm going to close this one rather than merge it as-is: the branch is quite a few months stale against `main`, so most of the diff (the bulk of the 1200+/1100- line change) is drift — it reverts a lot of unrelated work that's landed since your fork point (filter/PM1 entities, several sensors, config-flow error message translations, etc.), not intentional changes.

That said, I pulled out what genuinely holds up and adapted it into #323 (adds `ventilation_mode`, `windfree`/`windsleep`, a CO2 sensor, and reuses `air_purifier.py`'s HEPA filter/device-active capabilities):

**Kept, credited to this PR:**
- `ventilation_mode` select for the Purification/Ventilation/SmartVentilation vocabulary, gated so it can only ever bind on a device using exactly that mode set (verified against every AC fixture in our test corpus so it can't affect a real air conditioner's climate card).
- `windfree`/`windsleep` switches on their own dedicated hrefs.
- HEPA filter / device-active reuse from `air_purifier.py`.
- A CO2 sensor, matched to `air_monitor.py`'s already-confirmed `carbon_dioxide`/`ppm` descriptor for the same field shape.
- Your `/airlevelcheck/vs/0` finding turned out to be the most valuable part: it wasn't just wrong for your device, it was already wrong for two of our *existing* AC fixtures too (real, populated AI-Purify state that the old code comment had dismissed as "scheduler plumbing"). That's now fixed for everyone, not just your model — nice catch.

**Left out:**
- The dust/fine_dust/super_fine_dust unit + device_class mapping (`µg/m³`, pm10/pm25/pm1) — we already have an explicit, reasoned rejection of that exact mapping in `air_monitor.py` for the same three fields: Samsung's PM10/PM2.5 convention doesn't confirm where a third tier or a PM1 reading fits, and guessing wrong mislabels the reading forever. If you can get a raw diagnostics dump (not just an app screenshot) that pins down which tier is which, that'd change the calculus.
- A standalone power switch — this registry deliberately keeps power on the climate card only; see the module docstring for why.
- Independent fan-speed/swing selects for every AC user — a reasonable UX idea, but a bigger, more opinionated change than a single device's support PR should carry, and out of scope here.
- The model-name diagnostic sensor — `/information/vs/0` is already covered via the global ignore list, so this wasn't closing an actual gap.

One more thing: I couldn't build a proper fixture for your device since no raw `/device/0` diagnostics dump was ever attached here (I don't want to fabricate one from prose descriptions). If you're able to grab one (Settings → Devices & Services → this device → the menu → Download diagnostics) and open a fresh device-support issue with it, that'd let us lock in `ventilation_mode`/`windfree`/`windsleep` with real regression coverage and get their write contracts confirmed against your actual hardware.

Thanks again for the contribution!

---

PR #341, opened 2026-08-09
Normalize appliance enums for HA translations

lkn94:
## Summary

Translates previously English Samsung status values in the Home Assistant integration.

- Cycle progress, e.g. `Rinse` → “Rinsing”
- Drum-clean counter, e.g. `3 cycles`
- Diagnosis status, e.g. `Ready`
- Buzzer volume options, e.g. `Volume_High` → “High”

## Testing

- Validated translation catalogs and key structure
- Checked Python syntax

### DIFF
--- a/custom_components/localthings/registry/capabilities/operational.py
+++ b/custom_components/localthings/registry/capabilities/operational.py
@@ -24,8 +24,25 @@ def _to_ocf(v):
     return _SAMSUNG_STATE_TO_OCF.get(v, v) if v is not None else None
 
 
+_PROGRESS_STATES = {
+    "None": "idle",
+    "Weightsensing": "weight_sensing",
+    "Wash": "wash",
+    "Rinse": "rinse",
+    "Spin": "spin",
+    "Finish": "finish",
+    "Steaming": "steaming",
+    "Airwashing": "air_washing",
+    "Drying": "drying",
+    "Cooling": "cooling",
+    "Predrain": "pre_drain",
+    "Prewash": "pre_wash",
+}
+
+
 def _progress(v):
-    return "Idle" if v in (None, "None") else v
+    """Normalize known Samsung phase tokens for Home Assistant enum translation."""
+    return _PROGRESS_STATES.get(v, v)
 
 
 def _int(v):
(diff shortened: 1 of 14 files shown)


custom_components/localthings/registry/capabilities/operational.py
```diff
@@ -24,8 +24,25 @@ def _to_ocf(v):
     return _SAMSUNG_STATE_TO_OCF.get(v, v) if v is not None else None
 
 
+_PROGRESS_STATES = {
```
mbillow: Keeping a map like this in code makes adding translations harder. It would be simpler to just pass these into `lower()` at runtime and add translations for the known values, letting the raw ones just pass through.

lkn94:
@mbillow You are right. Thanks for this review. Resolved it now in the newest commit.

mbillow:
Thanks for this, and for turning the hardcoded map into the `lower()` approach — that was exactly the ask. Landing it via #371, which carries your commit with authorship intact rather than reimplementing it. Closing this one in favour of that branch.

Two things came out of rebasing onto current main, and the first is a trap you had no way to see: **issue #345 landed after you opened this.** It added a "sticky hold" that keeps `progress` showing Finish for a grace window after a cycle ends, and its held value is the device's raw `'Finish'`. Once your change normalizes `rep_fn` to `'finish'` and gives the sensor `device_class="enum"`, the held value is no longer in `options`:

```
options:         ('airwashing', 'cooling', …, 'finish', 'idle', …)
sticky_value_fn: 'Finish'   → in options? False
```

Home Assistant raises for an enum state outside `options`, which takes the entity out rather than just showing it untranslated — and the hold engages at the end of every cycle. Neither change shows this alone, only the combination.

Second, on `options` more generally. The review note about letting raw values pass through turns out to be load-bearing in a way that isn't obvious: with `device_class="enum"` they *can't* pass through, because HA rejects any state not in the list. I checked how likely that is to bite — every `supportedProgress` token across the shipped fixtures is covered by your catalog list, so nothing is broken today. But this registry's rule is that an unrecognized device value renders raw, and Samsung ships far more devices than we have dumps for. #371 handles it in the sensor platform instead of the capability: the live value is admitted into `options`, so known values translate and unknown ones display untranslated. That also covers the `diagnosis` sensors, where `options=("ready",)` would have had the same problem for any device reporting something other than `Ready`.

One useful detail for future work here: dropping `unit="cycles"` was necessary, not just tidy — HA raises if an entity declares both a native unit and a catalog `unit_of_measurement`. It resolves the catalog one against the *default* language, so the localized unit strings you added are inert; they stay only because every catalog has to mirror English key for key.

Appreciate the translation work, it's a real improvement to how these read in the UI.

---

PR #350, opened 2026-08-10
Fix AC setup timeout and verify appliance course mappings

galaxysj:
## Summary

- stop treating UUID-prefixed file-transfer namespaces as sibling appliance trees
- prevent affected single-unit air conditioners from probing every master resource under a UUID that never answers
- preserve the original appliance course-label and setup-timeout commits on top of the current `main`
- add a regression test from the reported dishwasher resource shape, covering its complete decoded course set: `82`, `8A`, `A7`, `80`, `A8`, `88`, `8C`, `8D`, `8E`, and `8F`
- add a regression test from the reported washer resources, covering its complete live course set: `69`, `6F`, `73`, `75`, `78`, `01`, `71`, `96`, `88`, `70`, `6D`, `6A`, `76`, `72`, `6C`, `6E`, `6B`, `77`, `74`, `79`, `F1`, and `F3`
- require every reported non-personal washer code to have a Table_02 label; keep `F1`/`F3` backed by their device-provided UTF-8 names

## Root cause

The affected `ARTIK051_FAC_TIME_21K` advertises UUID-prefixed file-transfer links in `/oic/res`, but no UUID-prefixed `x.com.samsung.da.multidevice` resource. Pattern C treated every UUID prefix as evidence of a sibling indoor unit, then tried `/<uuid>/device/0` and the master's full resource surface under that prefix. None of those paths answered, so first setup consumed its enumeration timeout budget.

Pattern C now accepts only UUID prefixes attached to the Samsung multidevice resource type. UUID file-transfer namespaces are ignored, while real multidevice links and IDs from `/subdevices/vs/0` continue through the existing discovery paths.

The dishwasher diagnostics also show that this board has no `/wm/editcourse/vs/0`; its course list is encoded only in `/course/vs/0` `supportedOptions`. The new regression test uses that exact non-identifying resource payload and pins every decoded code, rather than testing only `82` and `8A`.

The supplied diagnostics themselves are not committed and contain no PII in this PR.

## Validation

- `.venv/bin/python -m pytest -q`: 1412 passed
- `git diff --check`

Please preserve the original commit history when merging: use a merge commit, not **Squash and merge**.


galaxysj:
And also the A8, 88, 8C missing too on the dishwasher.

mbillow:
Yah, the pattern was a side effect of two disparate observations colliding, thanks for the cleanup.

Issue #345, opened 2026-08-09
DA_WM_TP1_21_COMMON Progress no longer reports State "Finish"


jayrage887:
Washer Progress used to report State "Finish" before State "Idle", now it goes straigt from "Spin" to "Idle", Version 0.18 did not have that quirk, Dryer with the same Device type still used Finish then Idle.

Noticed because that broke my automation for announcements.

Heads up, using .19, not .20 atm, have not seen an update notification

mbillow:
This shipped in [v0.21.0](https://github.com/mbillow/localthings/releases/tag/v0.21.0), just released — update via HACS and `progress`/`progress_percentage` will hold at Finish/100% for 5 minutes instead of skipping straight to Idle/0.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #359, opened 2026-08-12
laundry: a post-Finish running stage is the cycle ending, not a new one

### LINKED ISSUE #358: Dryer Progress State Problems DA_WM_TP1_21_COMMON
Since the latest update that fixed the finish behaviour on my Washer, the Dryer Automations now broke because the following happens: Dryer repoducibly goes through progress in the following order: "Drying, Cooling, Finish, Drying, Idle" expexted behavior would be "Drying, Cooling, Finish, Idle" as it was before .21.

Also another issue, maybe unrelated, the dryer now sometimes has problems with the handshake.

2026-08-12 14:15:29.656 WARNING (MainThread) [custom_components.localthings.coordinator.192.168.50.189] poll failed, reconnecting: session operation failed
2026-08-12 14:15:46.851 ERROR (MainThread) [custom_components.localthings.coordinator.192.168.50.189] poll failed after reconnect: session operation timed out
2026-08-12 14:17:43.893 WARNING (MainThread) [custom_components.localthings.coordinator.192.168.50.189] poll failed, reconnecting: session operation failed
2026-08-12 14:15:46.851 ERROR (MainThread) [custom_components.localthings.coordinator.192.168.50.189] poll failed after reconnect: session operation timed out

Attached also a picture of the new behavior

Image

mbillow:
### Summary

Fixes #358, a regression from #346. The reporting `DA_WM_TP1_21_COMMON` dryer's `progress` now reads **Drying, Cooling, Finish, Drying, Idle** instead of ending at Finish, Idle.

### What the reported timestamps show

The issue's history screenshot captures two full cycles, an hour apart, identical to the second:

| cycle 1 | cycle 2 | value | Δ |
|---|---|---|---|
| 12:07:27 | 14:00:02 | Cooling | |
| 12:08:27 | 14:01:02 | Finish | +60s |
| 12:08:51 | 14:01:26 | **Drying** | +24s |
| 12:08:55 | 14:01:30 | Idle | +4s |

That reproducibility is the tell — this isn't a race, it's the device's normal shutdown sequence. `progress` does not go `Finish -> None`; it replays a running stage (`Drying`, the first real entry in this dryer's `supportedProgress`) for ~4s while winding down, then settles.

### Why #346 didn't work

The tail itself isn't new. `progress`'s `rep_fn` has always returned `"Idle"` whenever `state` isn't active, so the tail was masked — which is exactly why the dryer read "Finish then Idle" before 0.19, as #345 reports. #346 un-masked it, in two compounding ways:

1. **`sticky_bypass_fn` (`_live_progress_code`) fired on any concrete non-`Finish` progress code, deliberately ungated on `state`.** That reasoning is sound for *arming* — a new cycle's progress can appear before `state` catches up — but as a *release* condition it can't distinguish "new cycle starting" from "old cycle ending". Both look like a concrete progress code.
2. **`sticky_live_fn` was a second, ungated view of the same field**, and the bypass returned it in `rep_fn`'s place — so the hold became a channel for publishing a value the entity otherwise never shows.

Traced through `_apply_sticky`, the screenshot falls out exactly: `Finish` arms the hold → `Drying` trips the bypass, which sets `_sticky_until = None` (destroying the hold outright) and returns `"Drying"` → `progress` goes `None`, hold already gone, so `raw` gives `"Idle"`. Finish was observable for 24s, not the intended 5 minutes.

### Changes

- **`registry/capabilities/operational.py`**: `_live_progress_code` becomes `_new_cycle_running`, additionally requiring `state == 'active'`. The asymmetry with the arm condition is the point — failing to arm loses the Finish entirely (#345), while releasing late costs nothing, since the hold expires on its own. `rep_fn` for both entities is unchanged.
- **`registry/entities.py` / `sensor.py`**: `sticky_live_fn` is removed. `rep_fn` is now the only definition of a live value; the hold decides only *whether* to freeze, and the bypass returns `rep_fn`'s own result. This is the structural half of the fix — it makes the class of bug unrepresentable rather than patching the one instance.
- **`sensor.py`**: an already-open window is no longer restarted, so a `progress` that flaps in and out of `Finish` can't ratchet the deadline forward. `sticky_seconds` is now genuinely measured from the first `Finish` of a cycle, as the docstring already claimed.
- Tests: the #358 trace end-to-end at entity level, the same tail against `progress_percentage`, the flap bound, and `_new_cycle_running`'s unit cases.

### Behavior change worth flagging

A new cycle that starts **paused** no longer cuts the hold short (it did under the ungated bypass). It cannot be told apart from this tail, and nothing live is withheld by waiting: `rep_fn` shows `Idle` while paused with or without a hold, so the only difference is a stale `Finish` expiring on schedule instead of early.

### Not addressed

The issue also reports handshake/poll failures. Those timestamps start at 14:15 — fourteen minutes after the cycle ended — and #346 touched nothing in the transport, so this looks unrelated and is better tracked separately with debug logs.

### Test

```bash
python -m pytest -q          # 1508 passed
ruff check custom_components tests && ruff format --check custom_components tests
ty check custom_components/localthings tests
```

`ruff check .` additionally reports a pre-existing `RUF200` (`pyproject.toml` missing `name`) and a formatting diff in `.claude/skills/adding-device-support/SKILL.md`; both are present on `main` and untouched here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CjxQ63cDAsDPquDDGQzdHW)_

### DIFF
--- a/custom_components/localthings/registry/capabilities/operational.py
+++ b/custom_components/localthings/registry/capabilities/operational.py
@@ -62,17 +62,18 @@ def _just_finished(rep):
     return rep.get("x.com.samsung.da.progress") == "Finish"
 
 
-def _live_progress_code(rep):
-    """progress/progress_percentage's sticky_bypass_fn: a concrete,
-    non-Finish progress code being reported right now -- e.g. a new
-    cycle's own real 'Wash'/'Spin' -- must win over a still-open hold from
-    the previous cycle immediately. Not keyed on `state` (unlike
-    _is_active): _just_finished's whole premise is that `state` can't be
-    trusted to still say 'active' while a fresh, real progress value is
-    already there, and the same applies to recognizing when it's moved on
-    to a new one -- including while paused, e.g. adding a sock mid-hold."""
+def _new_cycle_running(rep):
+    """progress/progress_percentage's sticky_bypass_fn: drop the #345 hold
+    early once a new cycle is genuinely running.
+
+    Gated on `state == 'active'`, unlike _just_finished's arm condition
+    above: issue #358's dryer replays a running stage ('Drying') for a few
+    seconds after Finish while `state` already reads idle, and a bypass
+    keyed on the progress code alone read that tail as a new cycle and
+    republished it. Releasing late costs nothing -- an unreleased hold
+    still expires on its own -- so this side takes the stronger signal."""
     v = rep.get("x.com.samsung.da.progress")
-    return v is not None and v not in ("None", "Finish")
+    return _state_is_active(rep) and v is not None and v not in ("None", "Finish")
 
 
 def _remaining_seconds(raw):
@@ -190,14 +191,10 @@ OPERATIONAL_STATE = Capability(
         # sticky_* (issue #345): once progress reads 'Finish', keep
         # showing Finish/100 for a grace window even after machine_state
         # reverts, rather than falling to Idle/0 the instant it does --
-        # see sensor.py's _apply_sticky. rep_fn below is otherwise
-        # unchanged; the hold is entirely a read-side, per-entity concern,
-        # deliberately not gated on machine_state (_just_finished's
-        # docstring explains why). sticky_live_fn reads the raw field the
-        # same ungated way, for sticky_bypass_fn's benefit: a real
-        # progress value reported while paused (e.g. adding a sock
-        # mid-cycle) must win over a stale hold even though rep_fn itself
-        # would show "Idle"/0 there.
+        # see sensor.py's _apply_sticky. rep_fn below is unchanged and
+        # stays the only definition of a live value -- the hold decides
+        # only *whether* to freeze. A second, ungated one here is what
+        # let issue #358's post-Finish tail reach the entity.
         SensorDesc(
             key="progress",
             icon="mdi:progress-wrench",
@@ -208,8 +205,7 @@ OPERATIONAL_STATE = Capability(
             ),
             sticky_fn=_just_finished,
             sticky_value_fn=lambda rep: "Finish",
-            sticky_live_fn=lambda rep: _progress(rep.get("x.com.samsung.da.progress")),
-            sticky_bypass_fn=_live_progress_code,
+            sticky_bypass_fn=_new_cycle_running,
         ),
         SensorDesc(
             key="progress_percentage",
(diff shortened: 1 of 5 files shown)


mbillow:
Moved to draft — the diagnosis this rests on isn't confirmed yet.

The PR asserts that #358's post-Finish `Drying` is reported by the device with `state` already idle. That was inferred, not measured: `rep_fn` masks `progress` whenever `state` isn't active, so the reporter's statement that pre-0.21 the dryer ended at "Finish, Idle" implies `state` was non-active during the tail. The screenshot shows `Progress` only.

It matters, because the fix here gates `sticky_bypass_fn` on `state == 'active'`. If `state` actually reads `Run` during that spurious `Drying`, `rep_fn` renders it directly and this PR doesn't fix #358 — it's only a hardening of the bypass.

Asked for the `machine_state` history over the same window in #358; holding until that comes back.

One thing worth recording separately, found while re-checking this. `/operational/state/vs/0` is `poll_tier="hot"`, refreshed by `_run_subpolls` every 3s and applied immediately. The 30s summary poll then applies **every** href from the `/device/0` batch unconditionally (`coordinator.py:1347`), hot ones included, and `ObserveManager.apply()` merges with no freshness check — so if the batch's copy of an href lags a direct GET of that same href, the sweep moves the cache backwards. `_poll_subdevice_flat_hrefs` already skips hot/warm hrefs for exactly this reason ("strictly more current than this once-per-summary-poll pass could offer", `coordinator.py:777-781`); the main sweep has no such guard.

That is *not* what caused #358, as far as I can tell — a lag big enough to produce `Drying` at 14:01:26 would also have shown a spurious `Drying` during the Cooling minute, and that part of the reporter's history is clean. Noting it as a latent inconsistency rather than folding it in here.

---

mbillow:
Confirmed by the reporter's `machine_state` history — out of draft.

Cross-referencing the two entities over the same window:

| time | Progress | Machine state |
|---|---|---|
| 12:08:27 | Finish | Active |
| **12:08:51** | **Drying** | **→ Idle** |
| 12:08:55 | Idle | Idle |
| 14:01:02 | Finish | Active |
| **14:01:26** | **Drying** | **→ Idle** |
| 14:01:30 | Idle | Idle |

The spurious `Drying` lands on the exact second machine state flips to idle, in both cycles. That settles it: `rep_fn` returns `Idle` whenever state isn't active, so it cannot have produced that value, and the only remaining path was the ungated `sticky_live_fn` the bypass returned in its place.

Replaying the sequence against the pre-fix path reproduces the report verbatim, and against this branch it holds:

```
sequence:  (Run, Cooling) -> (Run, Finish) -> (Ready, Drying) -> (Ready, None)
pre-fix:   Cooling, Finish, Drying, Idle     <- issue #358
this PR:   Cooling, Finish, Finish, Finish   <- held, then Idle at expiry
```

Both halves of the fix independently block it: the `state` gate stops the bypass firing, and with `sticky_live_fn` gone the bypass could only ever return `rep_fn`'s own `Idle` anyway.

This also rules out the sweep-staleness theory I noted above — that would have required `state` to read active at 14:01:26, and it reads idle. Still worth filing separately as a latent inconsistency, but it isn't this.

One incidental finding: the `Active → Idle → Active` blip at cycle start (11:15:50 / 11:15:57 / 11:16:04) is on `machine_state` itself, so it's the appliance, not us. `Progress` mirrors it exactly because `rep_fn` masks on state. Unchanged by this PR.

---

Issue #325, opened 2026-08-07
Request: Use Home Assistant PM sensor device classes for air purifier particulate sensors


ericksoegiantoro-a11y:
### Appliance model

ARTIK051_TVTL_18K

### What's missing or wrong

While testing the ARTIK051_TVTL air purifier family (AX60R5080WD/SE), I noticed that /sensors/vs/0 already exposes air quality data that could be represented using Home Assistant’s standard particulate sensor metadata.
The resource reports:
* Dust
* FineDust
* SuperFineDust
* Odor
* CleanLevel
The existing helper sensor_item_value() in registry/capabilities/common.py already parses this resource correctly, so no parser changes are required.

I modified custom_components/localthings/registry/capabilities/air_purifier.py by updating the AIR_QUALITY capability to use Home Assistant’s standard metadata for the particulate sensors:
```python
SensorDesc(
    key="dust",
    field="x.com.samsung.da.items",
    icon="mdi:blur",
    device_class=SensorDeviceClass.PM10,
    unit="µg/m³",
    state_class=SensorStateClass.MEASUREMENT,
    value_fn=lambda items: sensor_item_value(items, "Dust"),
),

SensorDesc(
    key="fine_dust",
    field="x.com.samsung.da.items",
    icon="mdi:blur",
    device_class=SensorDeviceClass.PM25,
    unit="µg/m³",
    state_class=SensorStateClass.MEASUREMENT,
    value_fn=lambda items: sensor_item_value(items, "FineDust"),
),

SensorDesc(
    key="super_fine_dust",
    field="x.com.samsung.da.items",
    icon="mdi:blur",
    unit="µg/m³",
    state_class=SensorStateClass.MEASUREMENT,
    value_fn=lambda items: sensor_item_value(items, "SuperFineDust"),
),
```

No changes to the parser, registry, coordinator, or discovery logic were required. The existing implementation of sensor_item_value() already extracts the values from /sensors/vs/0.

I tested this on two identical ARTIK051_TVTL_18K units. Both expose the same /sensors/vs/0 structure, and the entities are created and update correctly.

### Diagnostics file

[config_entry-localthings-01KYBVPEZJ6PP6T6WRDRGA611H.json](https://github.com/user-attachments/files/30837639/config_entry-localthings-01KYBVPEZJ6PP6T6WRDRGA611H.json)
[unofficial air_purifier.py](https://github.com/user-attachments/files/30837638/unofficial.air_purifier.py)

[...]

PR #365, opened 2026-08-13
washer: add 0A/B0 cycle labels; purifier: PM device classes

### LINKED ISSUE #363: Add wash cycle labels for Samsung WW90DG5G34ABLE
### Appliance model

WW90DG5G34ABLE (DA_WM_TP1_21_COMMON)

### What's missing or wrong

Hi!

As of `v0.21.2`, the following (non-cloud) wash cycle values do not have friendly names. Could you please add these?
- `0A`: Towels
- `B0`: Mixed Load

The already existing name mappings are correct for my washing machine.

### Diagnostics file

_No response_

JayChickenK:
Closes #363. Implements the confirmed mapping from #325.

## What

- Add Table_02 washer cycle labels `0A` (Towels) and `B0` (Mixed load), mirrored in all shipped languages.
- Give air-purifier particulate sensors HA `device_class`/`unit` (`Dust`→PM10, `FineDust`→PM2.5, `SuperFineDust`→PM1, µg/m³).

## Why

Issue #363 reported those two Table_02 codes as raw hex on a WW90DG5G34ABLE (`DA_WM_TP1_21_COMMON`). Existing Table_02 labels on that machine were already correct.

Issue #325 asked for PM device classes. A live same-moment LocalThings vs SmartThings read on ARTIK051_TVTL confirmed the three-tier mapping (`Dust` matched PM10 exactly; the other two were 1 µg/m³ off, same order). Air-monitor sensors stay untyped — that board still has no same-moment correlation of its own.

## Testing

- Translation catalogs all carry `0a`/`b0` and still match English key-for-key.
- Updated `tests/test_air_purifier_air_quality_statistics.py` for the 6-column sensor rows and the PM mapping.
- Could not run `pytest` in this environment (no Python 3.13+).


mbillow:
Thanks for this — both halves are landing, via #370, which carries your commit with authorship intact rather than re-implementing it. Closing this one in favour of that branch. Two things came out of review worth writing down:

**The unit was spelled with U+00B5 MICRO SIGN.** Home Assistant's `DEVICE_CLASS_UNITS` holds only the U+03BC GREEK SMALL LETTER MU spelling, so `'µg/m³' != 'μg/m³'` and the membership check fails. The two are indistinguishable in an editor, a terminal, and a review diff. The effect isn't a crash — HA logs a per-entity "not a valid unit for the device class" warning telling the user to report a bug against the integration. Easy to miss, and `tests/test_air_purifier_air_quality_statistics.py` hardcoded the same character, so the test agreed with the bug. It now takes the expected unit from HA's own constant, and there's a registry-wide guard checking every `SensorDesc` against HA's tables.

**The PM mapping got firmer evidence, so air_monitor is typed too.** The open question here was that naming order isn't proof. It turns out the device answers it: each dust reading's `value[]` is `[concentration, grade]`, and the grade bands aren't shared across the three fields. A reading of 18 grades one step above the floor as `SuperFineDust` (air_monitor fixture) but sits *at* the floor as `Dust` (range_hood fixture), on two families that both floor at 1. One threshold can't produce both, so the firmware itself treats them as three scales ordered coarse-to-fine — and where each crosses brackets the Korean CAI band for its tier. That's board-independent, which is why air_monitor gets the labels as well rather than being left as the odd one out.

One consequence worth knowing for future work in this area: these sensors had been recording long-term statistics *without* a unit, and HA doesn't just warn about that mismatch — it suppresses statistics generation for the entity until someone clicks through the repair. So adding the `device_class` on its own would have quietly frozen the history it was meant to label. #370 adds a config-entry migration that relabels the recorded metadata first (metadata only — the values were always µg/m³, so nothing is converted).

Appreciate you picking up two stale issues at once; the mapping work in particular is what unblocked #325 after it had been sitting.

---

Issue #166, opened 2026-07-28
Room A/C (ARxxTXFCAWKNEU)


HonbraDev:
### Appliance model

AR07TXFCAWKNEU, AR18TXFCAWKNEU

#### Diagnostic → Alarm code (`/alarms/vs/0`)

Alarm codes still get shown even if they're `OFF`. Examples: `ErrorCode_OFF`, `FilterAlarm_OFF`. A real filter replacement `.code` is `FilterAlarm` and has a `.state` of `Created`.

Not sure if this is how it's supposed to work, but I didn't get a notification/repair stutus for the air filter one.

#### Diagnostic → Filter status (`/filter/airdustfilter/vs/0`)

`.filterUsage` is `100` (of `500`; 20%) even when the `.filterStatus` is `wash`.

#### Configuration → Air purification

Air purification is in the Configuration section, even though the feature is not present and toggling the option does not trigger a beep.

#### Configuration → Beep

The option to turn beeping on/off is missing.

#### Options

The *Air conditioner options* in the official app are as follows:
- Sound Effect
- Lighting
- Auto Clean
  - At the end of the cooling operation, the internal moisture is dried and hygienically managed.
- Self-Check

#### Fast (Turbo), Quiet

These are mutually exclusive, and incompatible with Wind-Free. Both lock the *Wind strength* in the official app.

Add them to the *Fan mode* dropdown?

#### Wind-Free

This is implemented as a toggle in the official app (and a toggle button on the remote), but it's mutually-exclusive with Fast and Quiet, and is incompatible with the wind speed and direction options (since there are no blades to swing).

Add it as another "fan mode" option?

#### Good'Sleep

Can probably only be disabled via OCF (it uses the timer functionality only present on the remote).

### Diagnostics file

The Device Profile / Presentation ID for both units in SmartThings Advanced is `DA-AC-RAC-000003`.

`AR18TXFCAWKNEU` unit:
[config_entry-localthings-01KYMAVQJHEHFPF4B7SAT88FSY_f1_h.json](https://github.com/user-attachments/files/30463723/config_entry-localthings-01KYMAVQJHEHFPF4B7SAT88FSY_f1_h.json)

`AR07TXFCAWKNEU` unit:
[config_entry-localthings-01KYMB3NFJHAN6AF9VNSB1KP1P_f1_o.json](https://github.com/user-attachments/files/30463724/config_entry-localthings-01KYMB3NFJHAN6AF9VNSB1KP1P_f1_o.json)

[...]

PR #390, opened 2026-08-18
air_purifier: expose CO2 when the device lists it

### LINKED ISSUE #387: Samsung Air Purifier added CO2
### Appliance model

Samsung Air Purifier (TP1X_DA-AC-AIR-01031_0000)

### What's missing or wrong

_AIR_QUALITY_SENSORS = (
    ("dust", "mdi:blur", "Dust", "measurement", "pm10", "μg/m³"),
    ("fine_dust", "mdi:blur", "FineDust", "measurement", "pm25", "μg/m³"),
    ("super_fine_dust", "mdi:blur", "SuperFineDust", "measurement", "pm1", "μg/m³"),
    **("co2", "mdi:molecule-co2", "CO2", "measurement", "carbon_dioxide", "ppm"),**
    ("odor", "mdi:scent", "Odor", None, None, None),
    ("clean_level", "mdi:air-filter", "CleanLevel", None, None, None),
)

### Diagnostics file

_No response_

JayChickenK:
Closes #387.

## What

Add a `co2` sensor to `air_purifier.AIR_QUALITY` for boards that list a `CO2` item on `/sensors/vs/0`.

## Why this shape

The reporter's suggested change was to add CO2 to `_AIR_QUALITY_SENSORS`. That tuple is shared with `air_monitor.SENSORS`, which already has its own `co2` descriptor — putting the key in the tuple would create a second entity with the same key on the air-monitor family.

The existing `TP1X_DA-AC-AIR-01031_0000` fixture (same model string as #387) also does *not* list a CO2 item. Adding the sensor unconditionally would give every current purifier an empty entity.

So this follows the AC family's existing pattern instead:

- Same field/shape `air_monitor.SENSORS` already models (`device_class='carbon_dioxide'`, `unit='ppm'`).
- `exists_fn` gates on the type being listed, so boards that don't report it stay unchanged.
- Catalog entry `sensor.co2` is already shipped in every language.

A diagnostics dump from #387 would still be useful to confirm the live items[] shape, but the mapping isn't a unit guess — it's the contract already shipping for this field on air-monitor (and AC).

## Testing

- Existing purifier fixtures still flatten with no `co2` key and no unbound hrefs.
- Injecting a `CO2` item into the ARTIK051_TVTL dump produces `612`.
- `exists_fn` is false on an empty items list and true when the type is present.
- Shared tuple still does not contain `co2` (guards the air-monitor duplicate).
- Could not run the full `pytest` suite here (no MSVC for `lru-dict` / Home Assistant). Registry checks were run against a stubbed import of `registry/`.

### DIFF
--- a/custom_components/localthings/registry/capabilities/air_purifier.py
+++ b/custom_components/localthings/registry/capabilities/air_purifier.py
@@ -34,7 +34,13 @@ from ..entities import (
     SwitchDesc,
     TimeDesc,
 )
-from .common import epoch_to_utc, filter_usage_percent, int_or_none, sensor_item_value
+from .common import (
+    epoch_to_utc,
+    filter_usage_percent,
+    has_sensor_type,
+    int_or_none,
+    sensor_item_value,
+)
 from .laundry import bool_option_exists, bool_option_value, option_value, option_write
 
 # Newer TP1X_DA-AC-AIR-class boards (issue #130) report fan modes directly
@@ -49,18 +55,6 @@ HREF_WIND_STRENGTH = "/wind/strength/vs/0"
 
 def _has_top_level_modes(rep, resources):
     return isinstance(rep.get("x.com.samsung.da.supportedModes"), (list, tuple))
-
-
-def _has_sensor_type(type_):
-    """True when /sensors/vs/0's items[] lists an item of this type."""
-
-    def fn(rep, resources):
-        return any(
-            isinstance(i, dict) and i.get("x.com.samsung.da.type") == type_
-            for i in (rep.get("x.com.samsung.da.items") or [])
-        )
-
-    return fn
 
 
 # Columns: key, icon, device item type, state_class, device_class, unit.
@@ -134,7 +128,8 @@ AIR_QUALITY = Capability(
         # CO2 (issue #387) -- same field/shape air_monitor.SENSORS already
         # models with device_class='carbon_dioxide'/unit='ppm'. Gated on the
         # type being listed so boards that don't report it (every current
-        # fixture) don't grow an empty entity.
+        # fixture) don't grow an empty entity. Disabled by default for the
+        # same reason as airconditioner.AIR_QUALITY (issue #166).
         SensorDesc(
             key="co2",
             field="x.com.samsung.da.items",
@@ -142,7 +137,8 @@ AIR_QUALITY = Capability(
             device_class="carbon_dioxide",
             state_class="measurement",
             unit="ppm",
-            exists_fn=_has_sensor_type("CO2"),
+            exists_fn=has_sensor_type("CO2"),
+            enabled_default=False,
             value_fn=lambda items: sensor_item_value(items, "CO2"),
         ),
     ),
--- a/custom_components/localthings/registry/capabilities/air_purifier.py
+++ b/custom_components/localthings/registry/capabilities/air_purifier.py
@@ -51,6 +51,18 @@ def _has_top_level_modes(rep, resources):
     return isinstance(rep.get("x.com.samsung.da.supportedModes"), (list, tuple))
 
 
+def _has_sensor_type(type_):
+    """True when /sensors/vs/0's items[] lists an item of this type."""
+
+    def fn(rep, resources):
+        return any(
+            isinstance(i, dict) and i.get("x.com.samsung.da.type") == type_
+            for i in (rep.get("x.com.samsung.da.items") or [])
+        )
+
+    return fn
+
+
 # Columns: key, icon, device item type, state_class, device_class, unit.
 # state_class is what makes Home Assistant keep long-term statistics --
 # without one, a reading is only in the short-term recorder history and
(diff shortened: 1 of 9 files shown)


custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -51,6 +51,18 @@ def _has_top_level_modes(rep, resources):
...
+def _has_sensor_type(type_):
+    """True when /sensors/vs/0's items[] lists an item of this type."""
+
+    def fn(rep, resources):
+        return any(
+            isinstance(i, dict) and i.get("x.com.samsung.da.type") == type_
+            for i in (rep.get("x.com.samsung.da.items") or [])
+        )
+
+    return fn
```
mbillow: Two things about this helper.

**Missing the stub-rep carve-out.** Every other presence-gating `exists_fn` in the repo opens with `is_stub_rep(rep) or ...`. Without it, a `/device/0` that returns the not-yet-fetched stub `{"href": "/sensors/vs/0"}` at forward time drops `co2` for the whole entry load while its five siblings still register — and since platforms enumerate `bound` exactly once, nothing adds it back until a reload. I reproduced this against the branch. It's the failure `common.ENERGY_METER` documents for issue #127.

**It's a byte-identical copy of `airconditioner._has_sensor_type`, minus the docstring.** `common.py`, next to `sensor_item_value`, is the natural single home for it. The dropped docstring matters too — it's what explains why the AC entities behind this gate are `enabled_default=False` (see the separate note below).

---

custom_components/localthings/registry/capabilities/air_purifier.py
```diff
@@ -106,17 +118,33 @@ def _has_top_level_modes(rep, resources):
...
-            unit=unit,
-            value_fn=lambda items, t=sensor_type: sensor_item_value(items, t),
-        )
-        for key, icon, sensor_type, state_class, device_class, unit in _AIR_QUALITY_SENSORS
+            icon="mdi:molecule-co2",
+            device_class="carbon_dioxide",
+            state_class="measurement",
+            unit="ppm",
+            exists_fn=_has_sensor_type("CO2"),
+            value_fn=lambda items: sensor_item_value(items, "CO2"),
```
mbillow, line 142: This borrows the air conditioner family's presence gate but not its `enabled_default=False`, which that gate's own docstring says it needs: a board can list every sensor type it knows about and report a permanent zero for hardware it doesn't actually have (issue #166). Paired with `state_class="measurement"`, such a board would record long-term statistics of a constant 0 ppm — which is worse than not having the entity, because the recorder keeps it.

Matching the AC family here (`enabled_default=False`) costs a user one toggle on a device that really does report CO2, and costs nothing on one that doesn't.

---

mbillow (commented):
Thanks — the shape of this is right. Keeping `co2` out of the shared `_AIR_QUALITY_SENSORS` tuple does genuinely avoid a duplicate key against `air_monitor.SENSORS`, the `carbon_dioxide`/`ppm` pairing matches what `air_monitor` already ships and passes the global device-class/unit scan, and the `sensor.co2` catalog entry is there. Full suite passes on the branch (1628), with goldens untouched because no fixture lists a CO2 item.

Two inline notes: the `exists_fn` needs the `is_stub_rep` carve-out, and the entity should probably follow the AC family's `enabled_default=False` since it's using that family's presence gate.

One thing that isn't a code issue but decides whether this closes #387: the only fixture we have for that model (`air_purifier_tp1x_da_ac_air_device.json`, TP1X_DA-AC-AIR) lists exactly five items with no `CO2`, and no other href on that board looks like a CO2 source. You note in the PR that no dump was available — so whether the reporter's unit actually lists the type is still open. A diagnostics dump from them, plus a fixture and golden, would settle it and give this a regression test; without one we're shipping a gate nothing exercises.

---

custom_components/localthings/registry/capabilities/air_purifier.py
JayChickenK: Addressed in 171316bd954946442443b58803a8672d132486cc: airconditioner._has_sensor_type is now common.has_sensor_type (next to sensor_item_value), with the original docstring and an is_stub_rep(rep) or carve-out. Both airconditioner.AIR_QUALITY and this purifier co2 descriptor use it. A stub {"href": "/sensors/vs/0"} keeps the entity; a genuinely empty {} or empty items[] still drops it.

custom_components/localthings/registry/capabilities/air_purifier.py
JayChickenK: Addressed in 171316bd954946442443b58803a8672d132486cc: this descriptor now sets enabled_default=False, same as airconditioner.AIR_QUALITY. has_sensor_type only proves the type is listed, so a board that advertises CO2 with a permanent 0 ppm (issue #166) will not record long-term statistics unless the user turns the entity on.

JayChickenK:
Pushed 171316b for the two inline notes: has_sensor_type now lives in common.py with the stub-rep carve-out, and purifier co2 is enabled_default=False.

On whether this closes #387: still no diagnostics dump from the reporter, so there is no fixture/golden that lists a CO2 item. The decode is covered by injecting that item into the ARTIK051_TVTL dump, and the gate (including the stub carve-out) is unit-tested, but nothing in the corpus exercises a real board that reports it. Happy to add a fixture once a dump lands; until then I would not treat #387 as confirmed on hardware.

custom_components/localthings/registry/capabilities/air_purifier.py
mbillow: Verified at `171316b` — this is the right shape, and extracting it to `common.has_sensor_type` next to `sensor_item_value` is better than what I asked for.

Recording why the carve-out is correct, since it does change behavior for the AC family's five existing gated entities and someone will reasonably question that later: `flatten()` re-evaluates `exists_fn` on every poll against live data, so including a stub at forward time isn't a permanent phantom — it resolves itself as soon as real data lands. Excluding it *is* permanent, because platforms enumerate `bound` exactly once. That asymmetry is why the carve-out is the convention here, and `AI_ENERGY_LEVEL` — the one gate in `common.py` that deliberately omits it — documents both the reason and the cost ("the entity not appearing until a reload if the first poll stubs this cold-tier href"). Your docstring addition covers this well.

The AC widening is benign in practice too: those five entities are `enabled_default=False`, so a stub-created entity that never populates isn't even visible unless someone enables it.

---

custom_components/localthings/registry/capabilities/air_purifier.py
mbillow: Verified at `171316b`. This matches the precedent that governs this gate: the AC family already ships a CO2 sensor (PR #316, ACA-KR-TP2-21-AN9000) behind the same `has_sensor_type("CO2")` with `enabled_default=False`, and the helper's docstring states the rule for anything gated on it.

Worth being explicit about the cost, since it's a product decision rather than a code one: #387's reporter won't see a CO2 entity after upgrading — they'll have to enable it from the device page. The alternative reading is that `air_monitor` ships CO2 enabled, so why not here; but that descriptor has no `exists_fn` at all, because that family always reports the type. Enabled-by-default goes with "this family always has it"; gated-and-disabled goes with "the type is merely listed." This is the second case, so I'd keep it as you have it.

---

mbillow (approved):
Both points addressed, and the helper extraction is a genuine improvement over what I suggested — one gate, one docstring, shared by both families. I've replied on each thread with the reasoning I verified, so the next person to question either choice has it recorded, and resolved them.

Re-checked at `171316b` beyond the deltas: no duplicate `co2` key (air_monitor keeps its own SENSORS capability and no registry binds two AIR_QUALITY capabilities), the shared `_AIR_QUALITY_SENSORS` tuple correctly did *not* gain a CO2 row, the `sensor.co2` translation key exists, the v2→v3 particulate statistics migration regex doesn't match `co2`, no stale `_has_sensor_type` references remain in `airconditioner.py`, and `sensor_item_value` returns an int so `state_class="measurement"` is honoured. Full suite passes (1632) with goldens untouched.

Approving. One thing that stays open and isn't yours to close: no fixture exercises the new gate, because the only dump we have for the model in #387 lists five items with no `CO2`. If the reporter can supply a diagnostics dump, a fixture and golden would turn this into something with a regression test behind it — worth asking on the issue.

---

PR #391, opened 2026-08-18
Keep asking the appliance for readings that never send live updates

### LINKED ISSUE #92: Subscribed-but-silent hrefs get no sub-polling in observe mode
## Summary

`try_enter_observe_mode` knows exactly which subscribed hrefs never emitted a notify during the grace period, then throws that information away. Those hrefs are counted as fully covered by push for the rest of the session, but `_run_subpolls` skips them because we're in observe mode — so they only refresh on the 30s `/device/0` summary sweep. Any device-initiated change on one of them (physical remote, schedule, the unit restoring its own state) is invisible for up to 30 seconds.

Came out of the discussion on #91, where it surfaced as "the card briefly shows the last cached setpoint right after a power-on" on a TP1X cool-only RAC. That's the visible symptom, but it isn't specific to that board or to writes — it's the general case of a resource we subscribe to but the device never pushes.

## Mechanism

`observe.py:170-205` — `try_enter_observe_mode` records per-href notify activity in `self._notified`, reduces it to a single global `fraction`, and if that clears `SUCCESS_FRACTION` (0.8) sets `subscribed_hrefs = subscribed` — every href, including the ones that stayed silent. The per-href detail is dropped.

`coordinator.py:265` — `_run_subpolls()` returns immediately when `mode == MODE_OBSERVE` and `force` is unset, on the (normally correct) assumption that observe-mode hrefs are covered by push.

Net effect: an href can be silent from the very first grace period and still get zero sub-poll coverage forever.

`fallback_hrefs` (`observe.py:264`) already looks like the right home for this, but it's only ever populated by `downgrade_to_poll()`.

`log_sweep_discrepancies` partly compensates by forcing subpolls for one cycle when the sweep disagrees with the cache — but that only fires *after* the sweep has already corrected the value, so it does nothing for the window the user actually notices.

## Repro

Driving `ObserveManager` directly with 12 hot/warm hrefs, where the device pushes on 10 and stays silent on `/power/vs/0` and `/temperature/desired/0`:

```
entered observe mode : True | mode = observe
notify fraction      : 0.83          <- clears the 0.80 threshold
subscribed_hrefs     : 12 hrefs
silent but "subscribed": ['/power/vs/0', '/temperature/desired/0']
fallback_hrefs       : []            <- silence never recorded
```

The two silent hrefs are then reported as subscribed in diagnostics and get no sub-polling.

This is also visible in the field: `observe_href_freshness_s` in the diagnostics payload will show a handful of hrefs sitting at a much higher age than the rest while `observe_mode` reads `observe` (see the dump in #89 for the shape).

## Proposed fix

Record the silence and keep sub-polling it:

1. In `try_enter_observe_mode`, after computing `fraction`, keep `subscribed - self._notified` as a set of known-silent hrefs (either a new `silent_hrefs` attribute or reuse `fallback_hrefs`, which is otherwise idle while in observe mode).
2. In `_run_subpolls`, instead of returning early in observe mode, fall through with just that set — hot/warm cadence, so 3s/6s slots on two or three hrefs rather than all of them.
3. Surface it in diagnostics next to `observe_subscribed_hrefs` so a user's dump says which resources aren't pushing.

Bounded extra traffic, no new registry surface, no per-device tuning. Also worth considering whether an href that goes silent *later* (subscription lapses without a reconnect) should get demoted into the same set — `StateCache.freshness_s()` already gives us the signal, and `log_sweep_discrepancies` is arguably a cruder version of the same idea.

## Alternative considered

Let entities/capabilities declare related hrefs to re-read after a write — the AC already has the list in `CLIMATE_CONSUMED_HREFS`, and `_poll_hrefs_blocking` already does targeted GETs.

Not proposing that as the primary fix: it only covers resources that change *as a result of our own write*, and does nothing when the change is device-initiated (physical remote, schedule), which is the broader half of the problem. It also needs a new descriptor field, discovery plumbing, and a post-write delay constant that would want tuning per board.

It does layer cleanly on top later if the warm sub-poll cadence turns out to be too slow for the moment right after a write, and by then the declaration costs nothing.

JayChickenK:
Closes #92.

## The problem

The integration can listen for live updates from the appliance. If most readings arrive that way, it stops asking for the rest and assumes they will be pushed too.

A few readings never send those live updates (power, setpoint, and similar). After that, they can sit outdated for up to 30 seconds — including when you change something on the machine itself.

## The fix

Remember which readings stayed silent when we first started listening. Keep asking *just those* every few seconds. Readings that already update live are left alone.

## How it is wired

Observe mode only needs 80% of subscribed hrefs to notify, then treated every subscribed href as push-covered. `_run_subpolls` then returned immediately, so a silent href waited for the next `/device/0` sweep.

`fallback_hrefs` was already unused while in observe mode (it only filled in on `downgrade_to_poll`). On enter, it now holds `subscribed - notified`. `_run_subpolls` polls only that set on the normal 3s/6s cadence. Diagnostics already reports this as `observe_fallback_hrefs`.

If every subscribed href notified, `fallback_hrefs` stays empty and sub-polling is still a no-op.

## Testing

- `enter_observe_mode` records only the silent hrefs; an all-notify grace period leaves `fallback_hrefs` empty.
- The partial-notify success-fraction test now asserts the silent remainder.
- Coordinator: observe + silent set polls only those hrefs; observe + empty set does not poll or sleep.
- `ruff format` / `ruff check` clean. Could not run the full pytest suite here (no MSVC for Home Assistant).

### DIFF
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -975,16 +975,23 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
     # ------------------------------------------------------------------
 
     async def _run_subpolls(self, force: bool = False) -> None:
-        """Poll hot/warm hrefs in the gaps between summary polls. No-op in
-        observe-primary mode (those hrefs are already covered by push)
-        unless `force` is set -- set when this cycle's sweep found the
-        cache disagreeing with a still-live observe session (see
-        log_sweep_discrepancies): a bounded fallback for a channel gone
-        silent without a reconnect."""
+        """Poll hot/warm hrefs in the gaps between summary polls.
+
+        In observe-primary mode this is a no-op for hrefs the device is
+        actually pushing, unless `force` is set (sweep disagreed with the
+        cache -- see log_sweep_discrepancies). Hrefs that were subscribed
+        but stayed silent through the grace period (issue #92) stay on
+        the hot/warm cadence via `fallback_hrefs`.
+        """
         if self._observe.mode == MODE_OBSERVE and not force:
-            return
-        hot = self._hot_hrefs
-        warm = self._warm_hrefs
+            silent = self._observe.fallback_hrefs
+            if not silent:
+                return
+            hot = [h for h in self._hot_hrefs if h in silent]
+            warm = [h for h in self._warm_hrefs if h in silent]
+        else:
+            hot = self._hot_hrefs
+            warm = self._warm_hrefs
         if not hot and not warm:
             return
         step = self._SUBPOLL_STEP_S
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -998,6 +998,8 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
         for i in range(1, 10):  # slots 1..9  (T+3 s … T+27 s)
             await asyncio.sleep(step)
             hrefs = list(hot) + (list(warm) if i % 2 == 0 else [])
+            if not hrefs:
+                continue
             async with self._session_lock:
                 try:
                     await self.hass.async_add_executor_job(self._poll_hrefs_blocking, hrefs)
--- a/custom_components/localthings/observe.py
+++ b/custom_components/localthings/observe.py
@@ -96,7 +96,8 @@ class ObserveManager:
         self._last_notify_ts: float | None = None
         # Wakes try_enter_observe_mode's grace wait early once enough hrefs
         # have notified. Guards `_notified` mutations, the `wait_for`, and
-        # fallback_hrefs discards from on_notification.
+        # fallback_hrefs (enter_observe_mode assignment, on_notification
+        # discard).
         self._notify_cond = threading.Condition()
         # Idle while polling, except after downgrade_to_poll (every href
         # that was subscribed). While in observe mode this is the set of
@@ -280,14 +281,13 @@ class ObserveManager:
         #294) -- committing against a session a reconnect already replaced
         would claim observe mode with nothing left to notice it's dead."""
         self.subscribed_hrefs = set(subscribed)
-        with self._notify_cond:
-            notified = set(self._notified)
         # Issue #92: subscribed-but-silent hrefs are counted as covered by
         # push if we drop this, but they never emit a notify. Keep them on
         # the poll cadence via fallback_hrefs (otherwise idle in observe).
-        # This is the 80% quorum snapshot; a later notify discards the
-        # href in on_notification.
-        self.fallback_hrefs = set(subscribed) - notified
+        # Same lock as on_notification's discard so a notify in this window
+        # cannot land on a set object that is about to be replaced.
+        with self._notify_cond:
+            self.fallback_hrefs = set(subscribed) - self._notified
         self._set_mode(MODE_OBSERVE)
         self.start_refresh_task(session)
 
--- a/custom_components/localthings/observe.py
+++ b/custom_components/localthings/observe.py
@@ -97,6 +97,11 @@ class ObserveManager:
         # Wakes try_enter_observe_mode's grace wait early once enough hrefs
         # have notified. Guards only `_notified` mutations + the `wait_for`.
         self._notify_cond = threading.Condition()
+        # Idle while polling, except after downgrade_to_poll (every href
+        # that was subscribed). While in observe mode this is the set of
+        # hrefs that were subscribed but never notified during the grace
+        # period (issue #92) -- they stay on the hot/warm sub-poll cadence
+        # instead of waiting for the 30s /device/0 sweep.
         self.fallback_hrefs: set[str] = set()
         self._on_applied: Callable[[str, dict, str], None] | None = None
         self._refresh_task: ObserveRefreshTask | None = None
@@ -270,6 +275,12 @@ class ObserveManager:
         #294) -- committing against a session a reconnect already replaced
         would claim observe mode with nothing left to notice it's dead."""
         self.subscribed_hrefs = set(subscribed)
+        with self._notify_cond:
+            notified = set(self._notified)
+        # Issue #92: subscribed-but-silent hrefs are counted as covered by
+        # push if we drop this, but they never emit a notify. Keep them on
+        # the poll cadence via fallback_hrefs (otherwise idle in observe).
+        self.fallback_hrefs = set(subscribed) - notified
         self._set_mode(MODE_OBSERVE)
         self.start_refresh_task(session)
 
--- a/custom_components/localthings/observe.py
+++ b/custom_components/localthings/observe.py
@@ -95,13 +95,14 @@ class ObserveManager:
         self._notified: set[str] = set()
         self._last_notify_ts: float | None = None
         # Wakes try_enter_observe_mode's grace wait early once enough hrefs
-        # have notified. Guards only `_notified` mutations + the `wait_for`.
+        # have notified. Guards `_notified` mutations, the `wait_for`, and
+        # fallback_hrefs discards from on_notification.
         self._notify_cond = threading.Condition()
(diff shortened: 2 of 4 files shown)


custom_components/localthings/observe.py
```diff
@@ -270,6 +275,12 @@ def enter_observe_mode(self, session, subscribed: set[str]) -> None:
         #294) -- committing against a session a reconnect already replaced
         would claim observe mode with nothing left to notice it's dead."""
         self.subscribed_hrefs = set(subscribed)
+        with self._notify_cond:
+            notified = set(self._notified)
+        # Issue #92: subscribed-but-silent hrefs are counted as covered by
+        # push if we drop this, but they never emit a notify. Keep them on
+        # the poll cadence via fallback_hrefs (otherwise idle in observe).
+        self.fallback_hrefs = set(subscribed) - notified
```
mbillow: The set is snapshotted the instant `await_observe_notifies` returns — which is as soon as 80% of hrefs have notified, not after the full 15s grace period. Anything slower than that quorum is classified silent even when it pushes perfectly well.

The reliable victim is a multi-block resource. `smartthings_local` withholds the initial 2.05 for those and re-delivers it through a serialized, paced refetch (`_refetch_one`, `_REFETCH_TIMEOUT_S = 15.0`), so something like `/course/vs/0` is essentially guaranteed to be absent at snapshot time — and then gets a full blockwise GET every 3s for the life of the session, which is the opposite of what this PR is for.

That's compounded by `on_notification` (observe.py:212) never pruning `fallback_hrefs`, so a misclassification can't recover: an idle setpoint href that notifies the moment the user changes something stays double-covered (push *and* 3s polls) until the next reconnect. A `self.fallback_hrefs.discard(href)` alongside the `self._notified.add(href)` there would make the set self-correcting, and would largely neutralize the snapshot-timing problem too — a genuinely silent href never notifies, so it stays in the set, which is exactly the property you want.

---

custom_components/localthings/coordinator.py
```diff
@@ -975,16 +975,23 @@ def _poll_hrefs_blocking(self, hrefs: list[str]) -> dict[str, dict]:
...
+            if not silent:
+                return
+            hot = [h for h in self._hot_hrefs if h in silent]
+            warm = [h for h in self._warm_hrefs if h in silent]
+        else:
+            hot = self._hot_hrefs
+            warm = self._warm_hrefs
         if not hot and not warm:
             return
         step = self._SUBPOLL_STEP_S
```
mbillow, line 997: This whole-cycle guard is now doing less work than it looks like it is. Once `hot` and `warm` are filtered down to the silent set, the common observe-mode case is a non-empty `warm` with an empty `hot` — and then every odd slot in the loop below builds an empty `hrefs` list, takes `_session_lock`, and dispatches a no-op executor job. Five times per cycle, contending with the poll and command paths for the lock each time.

The shape is pre-existing, but this change makes it the normal path rather than a corner case. An `if not hrefs: continue` right after `hrefs` is built covers it.

---

mbillow (commented):
Thanks — the core idea is sound and the plumbing is right. Full suite passes on the branch (1628). I confirmed the parts that could easily have gone wrong: the observe-mode no-op is preserved when nothing is silent, `force=True` still polls everything, and `fallback_hrefs ⊆ hot ∪ warm` holds because `_attempt_observe_mode` subscribes exactly `_hot_hrefs + _warm_hrefs`.

My concerns are all about *how* an href ends up classified as silent, not about the sub-polling itself — see the note on `enter_observe_mode`. The one-line `discard` in `on_notification` suggested there is the highest-value change: it turns the classification from a single snapshot taken at an awkward moment into something self-correcting, which makes the rest of this robust.

---

custom_components/localthings/observe.py
JayChickenK: Addressed in 43076352aeae28eac92eb84176f2c29a1e682769: on_notification now does self.fallback_hrefs.discard(href) next to _notified.add. The enter_observe_mode snapshot is still the 80% quorum, but a late first push (blockwise refetch, slower than quorum) drops the href so it is not GET-polled for the rest of the session. A genuinely silent href never notifies and stays in the set.

custom_components/localthings/coordinator.py
JayChickenK: Addressed in 43076352aeae28eac92eb84176f2c29a1e682769: if not hrefs: continue after the slot list is built, so an odd slot with an empty hot set does not take _session_lock or dispatch a no-op executor job.

custom_components/localthings/observe.py
```diff
@@ -270,6 +280,14 @@ def enter_observe_mode(self, session, subscribed: set[str]) -> None:
...
         would claim observe mode with nothing left to notice it's dead."""
         self.subscribed_hrefs = set(subscribed)
+        with self._notify_cond:
+            notified = set(self._notified)
+        # Issue #92: subscribed-but-silent hrefs are counted as covered by
+        # push if we drop this, but they never emit a notify. Keep them on
+        # the poll cadence via fallback_hrefs (otherwise idle in observe).
+        # This is the 80% quorum snapshot; a later notify discards the
+        # href in on_notification.
+        self.fallback_hrefs = set(subscribed) - notified
```
mbillow: The `discard` in `on_notification` is the right fix, but the snapshot and the assignment here are no longer atomic with respect to it:

```python
with self._notify_cond:
    notified = set(self._notified)      # lock released here
                                        # ← a notify landing in this window
                                        #   discards from the OLD set object…
self.fallback_hrefs = set(subscribed) - notified   # …and this overwrites it
```

`on_notification` runs on the DTLS reader thread, and this code runs immediately after the subscribe burst — the moment the device is pushing hardest, since a subscription is normally answered with the current representation. An href that notifies inside that window has its `discard` applied to the set object being replaced, so it lands in the new `fallback_hrefs` anyway despite having just proved it pushes. It then only escapes on its *next* notification, which for a change-only resource (door contact, alarm) can be hours of 3s/6s blockwise GETs against a perfectly live push channel — the exact case the new comment at lines 214-216 says must not happen.

Moving the assignment inside the `with` block closes it; `on_notification` already takes the same lock, so they serialize:

```python
with self._notify_cond:
    self.fallback_hrefs = set(subscribed) - self._notified
```

While you're here, two smaller things in the same area, neither blocking:

- The quorum residue the `discard` mitigates but doesn't remove: an href merely *slower* than the 80% quorum still gets pinned until its next push, and for change-only resources that's the same long tail. Snapshotting after the full grace period instead of at `await_observe_notifies`' return would make the classification match what the set claims to mean.
- `fallback_hrefs` is derived from `subscribed`, so an href whose `session.subscribe()` raised (currently just a warning) is in neither `subscribed_hrefs` nor `fallback_hrefs` — no push, no sub-poll, only the 30s sweep, even when it's a hot href. Not a regression from this PR, but seeding the set from the hrefs *attempted* rather than the ones that took would cover it for free.

---

mbillow (commented):
Both changes from the last round landed and are correct in substance — the `discard` makes the classification self-correcting, and `if not hrefs: continue` removes the empty-slot lock churn (nicely covered by the new test, including cancelling the background sub-poll before driving it directly). Suite passes at `35f144a` (1630).

The `discard` did introduce one thread-safety gap in `enter_observe_mode`, though — details inline. It's a one-line move, and with that in place I'm happy with this.

---

custom_components/localthings/observe.py
JayChickenK: Addressed in 4edd5b7866a0ca550437fb966f90fb4674f48276: allback_hrefs is now assigned inside the same with self._notify_cond that on_notification takes for the discard, so a notify in that window serializes instead of landing on the set object being replaced.

On the two non-blocking notes: I left them. Snapshotting after the full grace period would keep the subscribe lock-free wait sitting for the remaining ~15s even after the 80% quorum already returned, which is a bigger change than this PR's silent-href classification. Seeding from attempted rather than successful subscribes is a pre-existing coverage gap for session.subscribe() failures and is independent of observe-mode fallback. Happy to take either in a follow-up.

Issue #192, opened 2026-07-29
Cannot Connect to Device - Samsung Smart TV


Rustymage:
### Appliance model

QE55S95DATXXU

### What's missing or wrong

I have a working dryer & washer, yet when attempting to add the TV I get the following error `Cannot Connect to Device`

TV, dryer & washer are on the same segregated VLAN.

```
Starting Nmap 7.95 ( https://nmap.org ) at 2026-07-29 13:11 BST
Nmap scan report for 192.168.2.100
Host is up (0.0077s latency).

PORT      STATE         SERVICE
49152/udp closed        unknown
49153/udp closed        unknown
49154/udp open|filtered unknown
49155/udp closed        unknown
49156/udp closed        unknown
49157/udp closed        unknown
49158/udp closed        unknown
49159/udp closed        unknown
49160/udp closed        unknown

Nmap done: 1 IP address (1 host up) scanned in 4.66 seconds
```

### Diagnostics file

```
2026-07-29 13:09:08.367 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Live DTLS port candidates on 192.168.2.100: [49155, 49156, 49160]
2026-07-29 13:09:09.279 DEBUG (MainThread) [custom_components.localthings.coordinator.192.168.2.133] Finished fetching localthings_192.168.2.133 data in 5.323 seconds (success: True)
2026-07-29 13:09:20.682 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] port 49155 failed: DTLS handshake timeout to 192.168.2.100:49155
2026-07-29 13:09:24.757 DEBUG (MainThread) [custom_components.localthings.coordinator.192.168.2.145] Finished fetching localthings_192.168.2.145 data in 1.085 seconds (success: True)
2026-07-29 13:09:32.995 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] port 49156 failed: DTLS handshake timeout to 192.168.2.100:49156
2026-07-29 13:09:40.309 DEBUG (MainThread) [custom_components.localthings.coordinator.192.168.2.133] Finished fetching localthings_192.168.2.133 data in 1.353 seconds (success: True)
2026-07-29 13:09:45.310 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] port 49160 failed: DTLS handshake timeout to 192.168.2.100:49160
```

mbillow:
Thanks for the nmap output alongside the debug log — that combination is what made this diagnosable at all.

[...]

Issue #188, opened 2026-07-29
Feature Request: Generate certificate programmatically inside Config Flow


hmmbob:
**Is your feature request related to a problem? Please describe.**
Currently, users are required to execute an external script to generate the required TLS certificate prior to adding the integration, and then manually copy-paste the resulting keys/PEM files into the Config Flow.

Executing external scripts presents a usability barrier, especially for non-technical users or those running Home Assistant OS without easy access to a terminal environment. I've seen some requests on "how do I do this" in fora and Discord servers I attend.

 **Describe the solution you'd like**
Embed the certificate generation code directly into the integration's Config Flow execution path.

Possible flow - when a user adds the integration through the UI:
 1. The user inputs the device IP address
 2. The Config Flow programmatically executes the certificate creation logic (e.g., using Python's cryptography library) directly within Home Assistant. Config flows allow for multiple steps.
 3. The generated certificate and key are automatically saved to persistent storage and linked to the device config.
 4. Setup completes seamlessly without requiring the user to run external scripts or manually copy and paste PEM strings.

Integrating the certificate generation logic directly into the setup code makes `localthings` feel like a native, zero-friction Home Assistant integration while maintaining exact parity with what the standalone script currently does. You also wouldn't be shipping the assets/certs directly, which would mitigate DMCA risks.

mbillow:
Sorry, I’m not following, you are describing exactly how LicalThings works today. 

https://github.com/mbillow/localthings/blob/119a4f443cf2b995eb7ccef10044738fb84ce7cc/custom_components/localthings/config_flow.py#L76

There is no external script you need to run. You just need to give it the CA.

Mind clarifying what the actual ask here is?

hmmbob:
Ah, the "give the CA" part is what I meant.

Basically, step 4 above:

> Setup completes seamlessly without requiring the user to run external scripts or manually copy and paste PEM strings.

And "part 2" of the installation instructions here.

hmmbob:
But I think the answer is in #141 so closing this. Didn't see that one before opening this issue.

Love the work you do here - as mentioned before.

mbillow:
> The Config Flow programmatically executes the certificate creation logic (e.g., using Python's cryptography library) directly within Home Assistant. Config flows allow for multiple steps.

[...]

Issue #141, opened 2026-07-27
Companion integration: automatic AC14K_M CA bootstrap for Localthings


ervede:
Hi! I’ve created a small companion integration for Home Assistant that automates the AC14K_M CA certificate/key step required when adding Samsung appliances through Localthings.

**Repository:** https://github.com/ervede/localthings_ca_bootstrap

## What it does

- Fetches and verifies the AC14K_M CA bundle (same source as `setup_cert.py`)
- Provides the cert + key directly to Localthings’ config flow
- Lets users add an appliance by IP without running scripts or copy/pasting PEM blocks
- Optionally saves a local backup of the CA bundle for resilience
- Stores no persistent credentials itself — all long‑term storage remains in Localthings’ own config entry

This keeps Localthings unchanged while removing the last manual step for users.

## Note from the author

I’m not a developer — I built this with some AI assistance and by learning from existing Home Assistant integrations.  
If you think this is useful, I’m happy to improve documentation, adjust anything, or discuss whether any part of it should be upstreamed or linked from your README.

Thanks for your work on Localthings!

mbillow:
I have no intention of including our automating ways to get the certificate. This repo is intentionally clean from any proprietary artifacts to ensure it stays unambiguously compliant with US intellectual property laws. I’m not a lawyer, so I’m just trying to be as cautious as I can be.

PR #392, opened 2026-08-18
Let new users set up LocalThings entirely inside Home Assistant

JayChickenK:
## What problem this solves

Today, adding a Samsung appliance is hard for people who are not technical.

The README asked them to do two things **outside** Home Assistant first:

1. **Part 1 — is my appliance compatible?**
   Open a terminal and run `nmap` against the appliance. If the right UDP ports answer, it is new enough. If only port 8888 is open, it is older firmware and this integration cannot control it.

2. **Part 2 — get the CA certificate and key**
   Go to another GitHub project, run a Python script (`setup_cert.py`), then copy two separate PEM blocks (the certificate and the private key) into Home Assistant.

That is fine if you are comfortable with a terminal. It is a real wall if you run **Home Assistant OS** on a small box and only use the UI.

Part 3 (add the integration in Home Assistant) already worked. This PR moves Parts 1 and 2 into that same UI, so a new user can stay in Home Assistant the whole time.

## What a new user does now

1. Install LocalThings (HACS or copy the folder) and restart Home Assistant.
2. Go to **Settings → Devices & Services → Add Integration → LocalThings**.
3. Type the appliance IP address. That is the only thing on the first screen.
4. Home Assistant checks the appliance itself. You do not run `nmap`.
   - If the appliance speaks the local protocol (DTLS on UDP 49152–49160), setup continues.
   - If it only answers on TCP 8888, the form says it is older firmware and cannot be added. You find that out immediately, before pasting any certificates. (This needs the appliance to actively refuse the UDP range. A firewall that drops those packets silently instead leaves the diagnosis to the handshake, one step later.)
   - If nothing is at that IP, you get a normal “cannot reach this address” style error and can try again.
5. **First appliance only:** the next screen asks for the AC14K_M CA file. You need this once. LocalThings does not include it and does not download it for you.

   **What the file is**
   One text file in PEM format. It can be a *combined* file (the certificate and the private key in the same file). That is the easy option. You paste the whole file into the first box and leave the second box empty.

   **How to get it (no terminal)**
   The `smartthings-local` project already documents this. Its [`setup_cert.py`](https://github.com/QuiteYellow/SmartThings-Local/blob/main/setup_cert.py) downloads one public combined file: the CA certificate and the private key together.

   1. On a phone or computer, open that script and find the `cert.pem` download URL it uses (it is near the top, named `BRAYSTORM_URL`).
   2. Open that URL in the same browser. You should see a page of text starting with `-----BEGIN` and ending with `-----END`.
   3. Select all of the text, copy it, and paste it into the **first** box in Home Assistant (“CA certificate or combined bundle”).
   4. Leave the second box (private key) empty.
   5. Continue.

   **How to get it if you already have Python**
   Run [`setup_cert.py`](https://github.com/QuiteYellow/SmartThings-Local/blob/main/setup_cert.py). Then either paste the original combined `cert.pem` into the first box, or paste the split certificate and key into the two boxes. Combined is enough.

   Home Assistant splits a combined file and checks that the certificate and the key belong together. You only do this for the first appliance.
6. After that, Home Assistant talks to the appliance and creates the device, same as today.
7. **Every later appliance** only asks for the IP. The CA is reused. That part is unchanged.

## What changed in the UI

**Before (first device)**
One form: IP + CA certificate + CA private key, all at once. Compatibility was only checked after you already pasted the CA. The README told you to use `nmap` and `setup_cert.py` first.

**After (first device)**
Two short forms:

1. **IP address** — Home Assistant checks compatibility first.
2. **CA credentials** — one-time paste. Combined file is OK.

**Later devices**
Still just the IP. No change.

The README no longer tells people to run `nmap` or to split the PEM with a Python script. Setup is described as a Home Assistant walkthrough.

## Why the CA is still pasted by the user

This PR does **not** download the CA from the internet, and it does **not** ship the CA in the repo.

That matches the existing project decision in #141, #168, and #188. The CA is leaked Samsung key material. Using it for interoperability is one thing; this project distributing it or fetching it automatically is another.

So the user still has to obtain the CA themselves (the README still points at `setup_cert.py` as an example of how). The difference is they no longer need a terminal to *use* it: they can paste the whole file into Home Assistant.

## Technical notes for reviewers

- New host-step helper `_assess_compatibility()` runs the existing UDP / ClientHello scan **before** any CA is asked for.
- If every local-API UDP port is refused **and** TCP 8888 is open, we raise `LegacyFirmware` (`legacy_firmware` in the form). That is the “older firmware” case from the old Part 1.
- If ClientHello already found a real DTLS server, we do **not** treat 8888 as a reason to reject. Some boards can expose both. This also avoids calling a working device “legacy” when the cheap UDP sweep is wrong (same concern as #192).
- New helper `_parse_ca_credentials()` accepts either a combined PEM bundle in the certificate field, or the old split cert + key fields.
- If both PEMs load as real keys, we check they pair. Mock / unreadable PEMs still fall through to the existing minting error, so tests that use placeholder PEMs keep working.
- The later handshake still uses the scan from the host step, so we do not probe the appliance twice.
- New strings are in every language catalog (`en`, `de`, `cs`, `es`, `it`, `ko`, `nl`).

## Follow-up after a review pass

Second commit, addressing four things found reviewing the first:

- **A probe failure now retries on the step that can fix it.** The first-device error path pointed at the CA form for *every* failure, and that form has no host field — so a mistyped IP that got past the (deliberately lenient) compatibility check could only be corrected by abandoning the flow and starting over. It routes by what the failure blames now: the CA form for `InvalidCA` and `CertRejected`, the host form for everything else. Coming back through the host step keeps the CA paste instead of re-asking for the file, so the one-time step stays one-time.
- **The CA paste is parsed in the executor.** `_ca_pair_mismatch` imports `cryptography` lazily and parses a certificate and a private key, and a lazy import on the event loop is what HA's blocking-call detection flags. Every other caller of that module in this file already ran in the executor.
- **A passphrase-protected key is named** — new `encrypted_ca_key`, in all seven catalogs. The key regex matches `ENCRYPTED PRIVATE KEY`, but `load_pem_private_key` cannot open one without the passphrase and HA has nowhere to ask for it, so the pair check swallowed the failure and the user got the generic `invalid_ca` at mint time.
- **The docs hedge where the 8888 check cannot be certain**, per the note added to step 4 above. `_assess_compatibility` also records that an ICMP admin-prohibited reply lands in `refused`, which costs such a host #192's preferred-port rescue when 8888 is open too.

New tests: an address failure returns to the host step with the CA kept, a cert rejection stays on the CA step, and a passphrase-protected key is named.

## Tests

- Existing first-device tests now go **host → CA → connect**.
- New tests for: combined bundle paste, split cert + key, mismatched pair, missing key, 8888-only “legacy firmware” message, and “do not probe 8888 after DTLS is already confirmed.”
- `legacy_firmware` is in the error catalog, so the UI never shows a raw error key.

Hassfest, HACS, lint, and pytest are green.


custom_components/localthings/config_flow.py
```diff
@@ -196,6 +221,65 @@ def _normalize_pem(text: str) -> str:
...
+    key_field = _normalize_pem(key_field or "")
+    certs = _PEM_CERT_RE.findall(cert_field)
+    keys = _PEM_KEY_RE.findall(cert_field) or _PEM_KEY_RE.findall(key_field)
+    if not certs:
+        raise InvalidCA("No certificate found in CA PEM")
+    if not keys:
+        raise InvalidCA("No private key found in CA PEM")
+    ca_cert_pem = "\n".join(certs)
+    ca_key_pem = keys[0]
+    if ca_key_pem.startswith("-----BEGIN ENCRYPTED"):
```
JayChickenK: **[suggestion]** EncryptedCAKey is raised only when the extracted block starts with -----BEGIN ENCRYPTED. That covers PKCS#8 (BEGIN ENCRYPTED PRIVATE KEY), which the new regex already matches. A traditional OpenSSL passphrase-protected key (BEGIN RSA PRIVATE KEY / BEGIN EC PRIVATE KEY with Proc-Type: 4,ENCRYPTED) is also matched by _PEM_KEY_RE, fails the header check, and then _ca_pair_mismatch swallows TypeError/ValueError from load_pem_private_key(..., password=None) and returns False. _parse_ca_credentials therefore succeeds, _async_finish_probe runs, and _mint_credentials fetches a Samsung cloud UUID before _mint_leaf_cert finally raises generic invalid_ca. That is the "fail later with the unhelpful message" path this PR added encrypted_ca_key to avoid; the README also claims the flow "says so rather than failing later."

**Suggestion:** After extracting ca_key_pem, treat it as encrypted if the header is BEGIN ENCRYPTED, if the body contains Proc-Type: 4,ENCRYPTED, or if load_pem_private_key(..., password=None) raises the cryptography "Password was not given but private key is encrypted" error. Raise EncryptedCAKey in those cases so a passphrase-protected key never hits _fetch_samsung_uuid.

README.md
```diff
@@ -69,28 +69,25 @@ Other Tizen RT / DAWIT-family appliances almost certainly speak the same protoco
...
+4. The flow then sends a DTLS `ClientHello` to every port in the `49152-49160` range at once and keeps the one that answers -- a real DTLS server identifies itself in about one round trip, and the probe stops there, so nothing is left behind on the appliance. Only that port is then given a real certificate handshake: it fetches the current UUID from Samsung's cloud gateway, mints a leaf cert signed by your CA, and reads the device's identity and `/device/0`. On success it creates the config entry, already knowing the appliance's serial, model, and type.
```
JayChickenK: **[suggestion]** Step 2 already documents the UDP ClientHello / 8888 compatibility probe, which is what _assess_compatibility → _scan_ports actually does on the host form. Step 4 then says that after the CA paste "the flow then sends a DTLS ClientHello to every port in the 49152-49160 range at once and keeps the one that answers." That was true of the old single-step probe; _probe_and_validate now takes the host-step scan and does not scan again. The handshake / UUID / mint / /device/0 part of step 4 is still right; the second ClientHello sweep is not.

**Suggestion:** Rewrite step 4 to say the flow reuses the port found in step 2, then mints the leaf and does the certificate handshake (cloud UUID on the first device, reused leaf afterward). Leave the ClientHello explanation on step 2 only.

README.md
```diff
@@ -69,28 +69,25 @@ Other Tizen RT / DAWIT-family appliances almost certainly speak the same protoco
...
+2. Enter the appliance's IP. The flow probes UDP `49152-49160` for a DTLS server (the old nmap step), and checks TCP `8888` when nothing in that range answers:
+   - A DTLS answer in that UDP range: newer firmware (Tizen RT 3.x, DAWIT 3.0+). This is what the integration talks to. Most devices answer on `49154`/`49155`, but some builds bind lower (e.g. `49153`). The live port is detected automatically.
+   - Only `8888/tcp` open (token-based HTTPS): older firmware (roughly 2018-2022). **Not supported here.** The flow usually names this for you, on the IP screen, before it asks for anything else. It needs the appliance to actively refuse the UDP range to be sure, though, so a firewall that drops those packets silently instead leaves it to the handshake, which fails a step later with "nothing there speaks DTLS".
```
JayChickenK: **[nit]** The silent-drop hedge says the later handshake fails with "nothing there speaks DTLS." That string is config.error.no_dtls_server. A firewall that drops the UDP range (no ICMP) makes the sweep mark every port live; _classify_handshake_failure then raises NoResponse ("No response from that IP address — nothing came back on any port…"), not NoDtlsServer. 
o_dtls_server is the mixed live/refused case.

**Suggestion:** Quote the 
o_response wording, or say the handshake fails a step later with a reachability error, without naming the DTLS-specific string.

JayChickenK (commented):
## Summary

The host-then-CA split is implemented correctly: compatibility runs in the executor before any paste, 8888 is only consulted when ClientHello found nothing and the UDP range was actively refused, first-device CA parse accepts the public combined bundle (key + AC14K_M first, bag attributes ignored by the PEM regex), and _probe_and_validate(scan=...) reuses that scan so the appliance is not ClientHello'd twice. Retry routing, stored-CA reuse, unique-id / CONF_OCF_DEVICE_ID write path, and all seven translation catalogs (including ca, encrypted_ca_key, legacy_firmware) check out; the main merge did not damage the new flow. Remaining issues are a hole in the new passphrase-protected-key path (traditional OpenSSL PEM still reaches the Samsung UUID fetch) and README text that still describes a second ClientHello after the CA step.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

mbillow:
Closing this since the config flow has drifted so far and the next version won’t need the CA for most devices anyway. The spirit of these changes is already on main. Thanks for the contribution regardless!

PR #401, opened 2026-08-19
Add AddWash controls and sensors for washers

vmonkey:
Adds the three AddWash tokens on `/course/vs/0`'s `x.com.samsung.da.options`, which the washer registry previously ignored entirely.

```
AddWashSet_<0-7>         the alarm setting, and the only writable one:
                         a 3-bit mask over the moments it fires
                         (bit 0 rinse, bit 1 final rinse, bit 2 spin)
AddWashAvailable_<0-7>   the same three bits, but what the running cycle permits
AddWashIndicator_On/Off  the panel lamp: laundry may go in right now
```

The three are kept as separate entities rather than collapsed into one, because they answer different questions and an automation would otherwise be ambiguous.

## Entities

| entity | platform | category |
|---|---|---|
| `add_wash_alarm` | switch | config |
| `add_wash_alarm_rinse` / `_final_rinse` / `_spin` | switch | config |
| `add_wash_available` | binary_sensor | diagnostic |
| `add_wash_indicator` | binary_sensor | — |

Each self-gates on its own token via `laundry.bool_option_exists`, so washers that never report AddWash gain nothing. Reads and writes reuse `common.option_value` and `laundry.option_write` — no new parsing or transport code.

The mask is the only state; nothing is remembered HA-side. Turning the last moment off lands on 0 (alarm off) and turning one on from 0 turns the alarm back on. The master switch writes 7 / 0.

## Evidence

Confirmed on a WW6500 (`DA_WM_A51_20_COMMON`), not inferred:

- **Bit order** — over one wash cycle `AddWashAvailable` shed exactly one bit as each moment passed: `7` through Rinse, then `6`, `4`, and `0` as Spin began, returning to `7` at the end.
- **Independence** — the lamp read `On` mid-Rinse while the alarm was `0`, so it is not a view of the setting.
- **The write** — `AddWashSet_7` written, read back as `7`, restored to `0`. The single-token write merged by prefix, leaving the sibling tokens untouched.

`AddWashAvailable` deliberately gates nothing: it describes the cycle, not the setting.

## Tests

New `tests/test_washer_add_wash.py` covers all eight mask values, per-bit set/clear without disturbing the others, the outgoing payload, malformed and out-of-range values, and per-token capability detection. Full suite passes; `washer_dryer_combo`'s golden gains the six keys and no other golden changes.

Adds a scrubbed `washer_ww6500` fixture — the first plain washer in the corpus with AddWash tokens and an empty `editCourseList`. Serial and otnDUID are `**REDACTED**`; it contains no IPs, MACs, UUIDs, or key material.


tests/test_golden_regression.py
```diff
@@ -389,6 +389,22 @@ def test_registry_reproduces_golden_state_keys_for_washer_dryer_combo():
     )
 
 
+def test_registry_reproduces_golden_state_keys_for_washer_ww6500():
+    """DA_WM_A51_20 front-loader, routed by /oic/d's oic.d.washer rather
```
JayChickenK: This dump is not routed by `/oic/d`'s `oic.d.washer`.

`_new_state_keys("washer_ww6500", resources)` is called with no `device_types`, the fixture has no `/oic/d`, and `DA_WM_A51_20_COMMON_WW6500` types as washer via `_CONSUMER_PREFIX_TO_KEY` (`WW`). Same path as other laundry boards that share `DA_WM_` with dryers.

Please rewrite the docstring to say that -- these comments are how the next A51 dump gets typed. The empty `editCourseList` / `supportedOptions` fallback note is the useful part and should stay.

custom_components/localthings/registry/capabilities/washer.py
```diff
@@ -350,5 +434,29 @@ def validate(p, rep, resources):
...
+            entity_category="config",
+            exists_fn=bool_option_exists("AddWashSet"),
+            rep_fn=_add_wash_any("AddWashSet"),
+            write_fn=_add_wash_alarm_write,
+        ),
+        _add_wash_bit_switch("add_wash_alarm_rinse", "mdi:water", 0),
+        _add_wash_bit_switch("add_wash_alarm_final_rinse", "mdi:water-check", 1),
+        _add_wash_bit_switch("add_wash_alarm_spin", "mdi:sync", 2),
+        BinarySensorDesc(
+            key="add_wash_available",
```
JayChickenK, line 468: This is the 3-bit remaining-phase mask OR'd into a single bool. Confirmed behaviour on the WW6500: `AddWashAvailable` went 7 -> 6 -> 4 -> 0 through the cycle and reset to 7 at the end, so this sensor is on while idle and only goes off when spin starts.

That's a coherent split with `add_wash_indicator` (the "add now" lamp) and I'm fine keeping it as a diagnostic. Worth a one-liner in the entity comment so nobody automates "allowed" expecting the door to be open. Per-bit available sensors would match the alarm bits if anyone actually needs "rinse window still coming" -- not asking for that here.

JayChickenK (commented):
Nice work - this is a hardware-confirmed, well-tested addition, not a speculative decode.

The mask handling is right (missing/malformed/out-of-range stays `None`, never `0`), writes reuse the one-token prefix merge, and each entity self-gates on its own token so pre-AddWash washers gain nothing. Tests cover all eight masks, per-bit set/clear, the outgoing payload, and the WW6500 dump; the combo golden only grows these six keys.

Not a blocker for merge once CI is green, but please fix the routing comment called out inline - these goldens are load-bearing documentation.

I could not approve the waiting fork workflow from this account (403). A maintainer still needs to click **Approve and run workflows** on https://github.com/mbillow/localthings/actions/runs/32231515887

Optional, not blocking:
- `add_wash_available` collapsing the 3-bit remaining-phase mask to a bool is coherent with `add_wash_indicator` as the "add now" signal, but idle is `AddWashAvailable_7` so "allowed" is on at rest. Don't automate on it expecting the door lamp.
- The three moment switches could be `enabled_default=False` so a typical washer doesn't grow three extra config toggles. Fine either way.
- Master `On` always writes `7`. Documented, and there's no HA-side memory of a previous subset.

custom_components/localthings/registry/capabilities/washer.py
```diff
@@ -264,6 +264,90 @@ def validate(p, rep, resources):
...
+def _add_wash_set_write(mask):
+    return ["course", "vs", "0"], {
+        "x.com.samsung.da.options": option_write("AddWashSet", str(mask)),
+    }
+
+
+def _add_wash_alarm_write(p, rep, href=None):
+    if p not in ("On", "Off") or not rep.get("x.com.samsung.da.options"):
+        return None
+    return _add_wash_set_write(0b111 if p == "On" else 0)
```
mbillow, line 327: This writes `AddWashSet_7` without ever consulting `_add_wash_mask`, so the master switch overwrites a mask the module has just refused to interpret.

On a device reporting `AddWashSet_15` (or a non-numeric `AddWashSet_On`), `_add_wash_mask` returns `None`. All four switches are still created and all read `unknown`, and the three per-moment switches correctly no-op through `_add_wash_bit_switch`'s `mask is None` guard — but `turn_on` here still emits `AddWashSet_7`, truncating bit 3 and destroying the moment the device supported. That's the write `_add_wash_mask`'s own docstring rules out: "a mask this model can't represent is a wrong model rather than something to write back." Verified by calling the descriptors directly against a rep carrying `AddWashSet_15`.

Gating it the same way the per-moment writes are gated — return `None` when `_add_wash_mask(rep, "AddWashSet") is None` — keeps the unrecognized-mask case read-only, which is what the rest of the module already promises.

---

tests/test_golden_regression.py
```diff
@@ -389,6 +389,22 @@ def test_registry_reproduces_golden_state_keys_for_washer_dryer_combo():
...
+    than a board token. Reports the AddWash tokens and an empty
+    /wm/editcourse/vs/0, so its cycle list comes from supportedOptions."""
```
mbillow: The docstring says this device is "routed by `/oic/d`'s `oic.d.washer` rather than a board token", but the fixture carries no `/oic/d` and `_new_state_keys` passes `device_types=()`, so that path never runs in this test. The WW6500 is typed solely by the consumer prefix in `x.com.samsung.da.description`.

Worth correcting because of what the wrong description hides: `DA_WM_A51_20` is not a registered board token, so an A51 unit whose description prefix isn't one the registry recognizes doesn't reach this registry at all. The test is pinning the description-prefix route, which is the fragile one here — saying so makes the coverage gap visible instead of implying a `/oic/d` route that would be robust to the description changing.

---

mbillow (commented):
Thanks for this — the shape is good and it's carefully done. I ran the full suite on the branch (1700 passed) and probed the new descriptors directly.

Two things inline: one real bug (the master alarm switch writes over a mask `_add_wash_mask` has rejected, which can truncate a 4-bit mask down to 3 bits on hardware that reports one), and one docstring correction on the golden test.

Things I checked that hold up: the read-modify-write on the mask isn't racy — `async_send_command` computes the body and applies the optimistic `merge_options_field` with no await in between, so back-to-back per-moment toggles compose correctly. `option_value`'s `prefix + "_"` anchoring keeps `AddWashSet` / `AddWashAvailable` / `AddWashIndicator` from colliding with each other or with the fixture's `QuickWashSet_`. Translations are complete across all 7 locales, and the new fixture is properly scrubbed with no unbound hrefs.

One deliberate behavior worth a second look rather than a defect: turning the master switch off and back on writes 7, so an automation that toggles the alarm nightly would silently reset a rinse-only setting to all three moments. The PR body documents this as intended given the mask is the only state — I'd just want that trade-off stated in the code near `_add_wash_bit_switch`, since it's the kind of thing a future reader will otherwise treat as a bug.

---

vmonkey:
Thanks for the suggestions, all incorporated.

custom_components/localthings/registry/capabilities/washer.py
```diff
@@ -264,6 +264,96 @@ def validate(p, rep, resources):
...
+    }
+
+
+def _add_wash_alarm_write(p, rep, href=None):
+    # Gated on the mask being readable, like the per-moment writes: a device
+    # reporting a wider mask than these three bits would otherwise have it
+    # truncated to 7 here, silently dropping a moment it supports.
+    if p not in ("On", "Off") or _add_wash_mask(rep, "AddWashSet") is None:
+        return None
+    return _add_wash_set_write(0b111 if p == "On" else 0)
```
mbillow, line 327: The unreadable-mask gate is right, thanks. There's a narrower version of the same problem still here: "On" writes `0b111` whenever the mask is *readable*, without checking whether the alarm is already on.

Home Assistant doesn't short-circuit `switch.turn_on` for an entity that's already on, and applying a scene re-asserts every captured state. So a scene or automation that asserts `add_wash_alarm: on` over a rinse-only `AddWashSet_1` silently rewrites it to `AddWashSet_7` — and the master switch shows no state change at all, because it was already on. The three bit switches flip in the UI, but nothing points at what caused it.

This is distinct from the off→on case you documented in `_add_wash_bit_switch` (which genuinely has nothing to restore): here the subset is still on the device, and the write throws it away. Returning `None` when the mask is already non-zero keeps "on" idempotent and leaves that documented behavior untouched:

```python
mask = _add_wash_mask(rep, "AddWashSet")
if p not in ("On", "Off") or mask is None:
    return None
if p == "On" and mask:
    return None  # already on; don't widen the moments the user picked
return _add_wash_set_write(0b111 if p == "On" else 0)
```

---

mbillow (commented):
Both points from the last round are properly addressed — I re-reviewed the whole diff at `51bce3c`, not just the deltas.

- The master write is now gated on a readable mask, with a comment that explains the truncation it prevents. Verified against a rep carrying `AddWashSet_15`: the write is refused rather than narrowing to 7.
- The golden docstring now names the description-prefix route and says what happens when the description doesn't match — which is the part a future A51 dump needs.
- The off→on trade-off is documented in `_add_wash_bit_switch`, which is exactly where someone hits it.

Full suite passes on the head (1704), `ruff check` and `ruff format --check` clean. I also re-confirmed the things most likely to break under a shared register: the bit math, the absence of a read-modify-write race (no `await` between `entity_rep` and the optimistic merge in `async_send_command`), `option_value`'s `prefix + "_"` anchoring against `AddWashAvailable`/`AddWashIndicator`/`QuickWashSet_`, and the translations across all 7 locales.

One inline note left — the same shared-register hazard in a narrower form (asserting "on" while already on widens a rinse-only mask). One line to fix, and then I think this is good to go.

---

Issue #254, opened 2026-08-02
Devices showing as unavailable after Homeassistant Restart


jayrage887:
Whenever I restart my Homeassistant (not full reboot, but simple restart) some of the devices are unavailable and i need to manually reload them individually, sometimes it takes more than one reload of a device, then it keeps working perfectly again. Sometimes it is the dryer, sometimes the washer, sometimes the dryer and one of the indoor splits of my AC system.

Most Recent localthings stable version on most recent stable Homeassistant core and os.

pendragon1:
I can confirm I have the same issue with the latest versions installed.

mbillow:
Anything in the logs? With the latest versions, I’d expect this to happen less given the work @QuiteYellow did for maintaining client ports across restarts.

QuiteYellow:
@mbillow, my dryer and oven: both honor RFC 6347 §4.2.8, reconnecting on the same source port should evict the stale association left by an unclean shutdown and completes in ~3s. The port fix is doing its job on these two, and tracing the coordinator's reconnect path, but we are finding out there's lots of variability between devices.

One thing that does look like a bug: the coordinator never sends a DTLS close_notify on shutdown. `async_close()` only runs on entry unload (reload/remove), and HA doesn't unload entries on a Core restart, so a plain restart leaves the old association orphaned on the device. Wiring `coordinator.async_close()` to an `EVENT_HOMEASSISTANT_STOP` listener would fix that. It still shouldn't be the cause of #254 by itself, though, because the fixed port evicts that orphan on the next reconnect anyway (measured above), but if devices aren't doing this, it might help.

Logs would help narrow it, does the reconnect handshake complete then fail, or time out entirely? That separates "the device reconnects fine but something downstream breaks" from "the device's DTLS service is just down for a while."

QuiteYellow:
Traced it further, and I think this might be the actual mechanism.

[...]

Issue #388, opened 2026-08-18
Would you want OCF media devices (TV, soundbar) in LocalThings?


vmonkey:
### Appliance model

Soundbar HW-S61B, TV UE50TU7172

### What's missing or wrong

I'd like to add a Samsung soundbar (`oic.d.networkaudio`) and a Tizen TV (`oic.d.tv`), and want to check whether you'd want that upstream before building anything.

Both work over DTLS today with PSK credentials. The protocol layer needs nothing new: `smartthings-local` already ships `PskAuth` and `DtlsCoapSession(auth=...)`. I've dumped every advertised resource on both devices over an authenticated session.

Four differences from appliances:

1. **PSK instead of certificates.** Both devices report sct: 1 in /oic/sec/doxm — symmetric pair-wise key is the only credential type they support, so the AC14K_M cert path can't apply to this device class at all. devowneruuid is readable from doxm unauthenticated, so the config flow needs only the host and the per-device key: no CA upload, no leaf minting, no rekey. _connect_session() is cert-only today.

3. **Ports change on every power-on.** Neither device answers in `PROBE_PORT_RANGE`. Across one TV power cycle the DTLS port moved 46800 → 51039 and the plaintext port 38784 → 51163. The DTLS port is discoverable as `p.port` in `/oic/res`, but reaching `/oic/res` needs the plaintext port, which moves too.
3. **No `/device/0`.** Authenticated GET returns 4.04, with `/nonexistent/xyz` also 4.04 as calibration — so it's really absent. `/information/vs/0` likewise. Resources are individually addressed and enumerated through `/oic/res` (64 on the soundbar, 67 on the TV). `/EasySetupResURI` is a real collection but holds only WiFiConf/DevConf/CoapCloudConf, so it's no substitute.

4. **The TV is offline whenever it's off,** and OCF can't wake it — the stack is down, so `/sec/tv/switch/binary` is off-only and power-on needs WoL or the 8001/8002 websocket.

The registry side looks like it needs no core changes: these devices expose the standard types (`oic.r.switch.binary`, `oic.r.audio` with `volume`/`mute`, `oic.r.mode`, `oic.r.media.input`) alongside Samsung vendor ones, so `by_type/tv.py` and `by_type/soundbar.py` would be data in the existing `Capability` model. The friction is entirely in points 1–3, which touch auth, port resolution in the reconnect path, and the enumeration model.

**Would you want this in LocalThings, or is it better as a fork?**

[...]

Issue #241, opened 2026-07-31
Onebody washer+dryer (AWM-WW-AID-26-ONEBODY): washer subdevice not discovered — third multidevice pattern (UUID from /oic/res multidevice link, no /subdevices/vs/0)


Hyunook:
### Appliance model

**AWM-WW-AID-26-ONEBODY** — Samsung "One Body" all-in-one washer+dryer (KR market). `one_ui_version: 7.0 Laundry Hub`, Tizen Lite 5.0, RT-OCF on UDP 49154. Integration v0.17.0.

### What's missing or wrong

The unit registers as a single **dryer** device (31 entities), but the board actually exposes **two complete subdevice trees** — the **washer subdevice is not discovered**, so all washer-side state/controls (washer power, washer course, washer operational state, water consumption) are missing. SmartThings cloud splits this same unit into two devices (washer + dryer); the local API has the same split, but neither of the two multidevice patterns from #177 matches this board — it looks like a **third discovery pattern**.

**Evidence from diagnostics:**

- `/multidevice/vs/0` answers with `x.com.samsung.da.numofsubdevice: "2"` (diagnostics `multidevice` key)
- but `subdevices: []` — Pattern A probes fail (`subdevice_probes: {"/device/1": false, "/device/2": false, "/multidevice/vs/0": true}`)
- Pattern B doesn't match either: there is **no `/subdevices/vs/0`** (no `subdeviceIdList` anywhere)
- however `/oic/res` advertises a **UUID-prefixed multidevice link**:
  `/58b7d338-15c5-97d3-b562-000000000001/multidevice/vs/0` (rt `x.com.samsung.da.multidevice`) — the UUID appears **only** in that link; nothing else in `/oic/res` carries the prefix

**Live confirmation (DTLS GET against the appliance, leaf-cert session):**

`GET /58b7d338-15c5-97d3-b562-000000000001/device/0` → **code 69, full Collection batch with 32 hrefs** — a complete, live **washer** tree, clearly distinct from the main (dryer) tree:

- `/information/vs/0` → `modelNum: "AWM-WW-AID-26-ONEBODY|80014641|20010001001A2342DAC30296009B0300"`, `description: "AWM-WW-AID-26-ONEBODY_**WF80H**/DC91-00153A_0174"`, **its own serial number** (redacted) — the main tree's description is `..._**DV80H27H**/DC91-00145A_305B` (dryer)
[...]

PR #405, opened 2026-08-22
Explore identity-bound OCF endpoint selection for a WD86 washer

atc722:
Hi @mbillow,

Thank you for the work you and the other contributors have been putting into
LocalThings, particularly the recent washer, device-identity, and config-flow
improvements.

I’m opening this as an early draft so the prototype is visible and I can ask
whether the direction would be useful to the project before taking it any
further. This is not a request to merge or begin a full review yet.

I would be very happy to adjust the design, wait for the related work to
settle, or keep this only as supporting evidence if another approach would fit
LocalThings better.

## Background

I have been investigating local connectivity for one Samsung washer-dryer
reporting the internal device family `AWM-KR-M64-24-WD86`.

In redacted, read-only testing, the physical host appeared to expose multiple
logical OCF devices. The washer was represented by its own top-level OCF
device UUID (`di`), while the host’s representative/root OCF identity was
different.

The matching logical-device container advertised a secure endpoint, and that
endpoint answered a stateless DTLS probe. This raised the possibility that an
IP-only setup flow may not always preserve the relationship between:

- the intended logical device;
- its exact OCF `di`;
- the discovery response source; and
- its advertised secure endpoint.

I may be overlooking a simpler integration point, so I would appreciate
guidance before treating this prototype as a final design.

## What the prototype currently does

The branch adds an optional advanced OCF device UUID field to the setup flow.

When an exact `di` is supplied, the flow:

- uses Home Assistant’s route-selected IPv4 interface;
- performs identity-aware discovery for that exact device;
- accepts the advertisement only from the host entered by the user;
- probes only the advertised secure-port candidates;
- requires one unambiguous live DTLS endpoint; and
- verifies the authenticated `/oic/d.di` before requesting `/device/0` or
  creating the config entry.

When the advanced field is left empty, the existing LocalThings setup path
remains unchanged and does not incur additional discovery work.

## Why I’m sharing this now

This prototype may provide a concrete LocalThings consumer for the
identity-aware discovery work that was held out of
QuiteYellow/SmartThings-Local#36.

I also noticed that #392 is revising the config flow and that #388 discusses
PSK credentials and automatic endpoint rediscovery. I would prefer to
coordinate with those efforts rather than duplicate work or make either PR
more difficult to integrate.

Depending on your preference, I can:

- wait for #392 and rebase the prototype onto the resulting flow;
- adapt it to a shared endpoint-resolver abstraction discussed in #388;
- reduce it to consumer-side tests for the SmartThings-Local helper; or
- leave it on the branch until there is a clearer integration point.

## Current limitations

For clarity, this draft does not yet make the WD86 connectable through
LocalThings.

The appliance rejected the existing certificate authentication path with
`unknown_ca`. A bounded diagnostic probe showed that it selected the expected
PSK cipher profile, but I do not currently have a legitimate OwnerPSK for the
device.

This prototype therefore does not currently:

- acquire, derive, store, or provision an OwnerPSK;
- perform ownership transfer or write OCF security resources;
- add PSK credentials to a LocalThings config entry;
- provide a general way for users to obtain the exact OCF `di`;
- rediscover an endpoint during reconnect;
- add WD86-specific capabilities or entities; or
- demonstrate an authenticated `/oic/d`, `/device/0`, or completed Home
  Assistant registration.

It would also not help an appliance that exposes no local DTLS listener.

## Current validation

- 1,769 LocalThings tests pass against the current released
  `smartthings-local` package, covering the unchanged normal path and mocked
  adapter boundaries.
- 1,769 tests pass with the local identity-aware discovery follow-up connected.
- Ruff format and lint pass.
- `ty check` and `git diff --check` pass.
- The diff contains no real device UUID, address, serial number, certificate,
  key, or account identifier.

The advanced path still depends on an unreleased SmartThings-Local API.
Hassfest, HACS validation, released-package integration, and real authenticated
device acceptance remain pending.

## Guidance welcome

Whenever convenient, I would appreciate your thoughts on the following:

1. Would an explicit, opt-in exact-`di` setup path be useful to LocalThings?
2. Would you prefer that I wait for #392 before adapting the config flow?
3. Should this be aligned with the PSK and reconnect-time endpoint work
   discussed in #388?
4. Is there another existing or planned implementation that this prototype
   should defer to?

There is no urgency from my side. I will keep the PR in draft and will not
request merge review until the surrounding direction is clearer.

Thank you for taking the time to look at this, and please feel free to suggest
a smaller or different scope if that would be more helpful to the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ This integration uses the [`smartthings-local`](https://github.com/QuiteYellow/S
 
 ### What you get
 
-Adding a device just needs a host IP and your CA credentials in the UI. The integration reads the appliance's identity and picks the matching capability registry on its own, so there's no per-model descriptor to write for a new unit of a type that's already supported.
+Adding a device normally just needs a host IP and your CA credentials in the UI. The integration reads the appliance's identity and picks the matching capability registry on its own, so there's no per-model descriptor to write for a new unit of a type that's already supported. For the uncommon case where one physical host exposes several logical OCF devices, the advanced setup path can bind endpoint selection to one exact OCF device UUID.
 
 Credential setup is one-time. The first device you add asks for a CA certificate and key (see Part 2); every device after that reuses the same stored CA and only asks for the host IP, minting its own per-device leaf cert automatically.
 
@@ -81,9 +81,13 @@ This repo doesn't include the needed CA bundle. For an example of how to obtain 
 1. Copy `custom_components/localthings/` into your HA config's `custom_components/` directory. (Or add this repo as a custom repository in HACS — `Integration` category — and install it from there.)
 2. Restart HA.
 3. **Settings > Devices & Services > Add Integration > LocalThings.**
-4. First device: paste the appliance's IP, plus the contents of the CA private and public key from Part 2.
-5. The flow sends a DTLS `ClientHello` to every port in the `49152-49160` range at once and keeps the one that answers -- a real DTLS server identifies itself in about one round trip, and the probe stops there, so nothing is left behind on the appliance. Only that port is then given a real certificate handshake: it fetches the current UUID from Samsung's cloud gateway, mints a leaf cert signed by your CA, and reads the device's identity and `/device/0`. On success it creates the config entry, already knowing the appliance's serial, model, and type.
+4. First device: paste the appliance's IP, plus the contents of the CA private and public key from Part 2. Leave **Exact OCF device UUID (advanced)** blank for normal setup.
+5. With the advanced field left blank, the flow sends a DTLS `ClientHello` to every port in the `49152-49160` range at once and keeps the one that answers -- a real DTLS server identifies itself in about one round trip, and the probe stops there, so nothing is left behind on the appliance. Only that port is then given a real certificate handshake: it fetches the current UUID from Samsung's cloud gateway, mints a leaf cert signed by your CA, and reads the device's identity and `/device/0`. On success it creates the config entry, already knowing the appliance's serial, model, and type. Supplying an exact OCF device UUID uses the identity-bound path below instead and never scans or falls back to the fixed port band.
 6. Every subsequent device only asks for the host IP. The stored CA credentials are reused, and so is the leaf cert itself -- every appliance accepts the same one -- so adding a second appliance doesn't depend on Samsung's cloud being reachable at all. If a device rejects the reused cert (the UUID behind it does rotate), the flow mints a fresh one and retries by itself.
+
+For the uncommon case where one IP exposes several logical OCF devices, the advanced field accepts the exact OCF device UUID (`di`). This is the UUID reported by the target's authenticated OCF `/oic/d` resource, not the SmartThings platform device ID or a Samsung cloud certificate UUID. Use this path only when that exact value is already available from authenticated OCF diagnostics; LocalThings does not derive it from SmartThings metadata. The flow runs identity-aware multicast discovery on Home Assistant's route-selected IPv4 interface, accepts only a stable exact-`di` advertisement from the submitted IP, and statelessly probes only the advertised secure ports. After authentication it reads `/oic/d` and requires the same exact UUID before it requests `/device/0` or creates an entry. It never falls back to `pi`, a serial number, the host, or the normal fixed port band on this explicit path.
+
+This advanced field changes setup-time endpoint selection and identity verification only. It does not acquire credentials, perform ownership transfer, add reconnect-time rediscovery, or make an appliance compatible with the certificate authentication described in Part 2 if that appliance requires a different authentication method.
 
 Entities appear under one HA device per appliance, named for the appliance's type and model. Rename freely: the device is keyed on the appliance's own OCF device ID, not its name. (Some Samsung models ship the same serial number on every unit of a model, so the serial can't tell two of them apart -- the OCF device ID can.)
 
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -52,6 +52,7 @@ from .const import (
     CONF_LEARN_MODES,
     CONF_MANUFACTURER,
     CONF_MODEL,
+    CONF_OCF_DEVICE_ID,
     CONF_PORT,
     CONF_SERIAL,
     DEFAULT_CLOUD_COURSES_ENABLED,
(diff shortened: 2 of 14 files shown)


mbillow:
I’ll take a deeper dive into this at some point, but I have a generic question out of the gate.

> I may be overlooking a simpler integration point, so I would appreciate guidance before treating this prototype as a final design.

Is there not some way to determine all available devices on the platform? Asking the user for some arbitrary UUID seems like a poor UX. I’d much rather just ask for the IP, discover all device IDs and then use the existing sub-device patterns to automatically expose all sub-devices of the composite platform.

atc722:
Thank you for taking a look and for raising this. I agree that asking the user to provide an OCF UUID would be poor final UX.

The UUID field was intended only as a conservative prototype around the current held SmartThings-Local helper, which accepts one exact target `di`. It was not intended as the final setup design.

The public multicast directory response from this host already contained multiple top-level `di` containers, so I believe an IP-only flow should be feasible:

1. accept only the host IP;
2. enumerate the stable, source-bound `(di, advertised secure ports)` candidates reported by that IP;
3. statelessly probe the advertised ports;
4. after authentication, require `/oic/d.di` to match the advertised identity; and
5. expose the confirmed identities through the existing sub-device model where they are reachable through the same session.

The remaining uncertainty is the session topology. LocalThings’ current sub-device implementation assumes that the logical devices are reachable through one physical connection and DTLS session. For this WD86 host, I have confirmed the separate public `di` advertisements and live DTLS candidates, but without an OwnerPSK I have not yet been able to determine whether those identities are available through one authenticated session and UUID-prefixed paths, or whether they require separate DTLS sessions.

If they share one session, reusing the existing sub-device patterns seems like the right fit. If they require separate sessions, I would rather discuss an appropriate multi-session or per-identity shape than force them into the current sub-device abstraction.

If this matches the direction you had in mind, I’ll revise the prototype around IP-first enumeration and remove the user-supplied UUID field. Thank you again for pointing me toward the better setup experience.

mbillow:
I wonder if @QuiteYellow has opinions here, but I don’t think it is a horrible idea for us to just instantiate and maintain a SmartThings-Local client per device on a multi-device platform. The downside is that we lose per-physical-device control of things like throttling and request pacing.

QuiteYellow:
@mbillow, the per-device client shape is fine, but the pacing loss you named is the smallest of three costs.

First, `_local_source_port(host)` in `coordinator.py` is keyed on the host IP alone, and its docstring already says the port must stay unique per device on that host. That holds today because one host is one device. N clients per host collapses them onto one source port, which turns the RFC 6347 §4.2.8 eviction that fixed #254 against you. Re-keying on `(host, di)` is a prerequisite.

Second, session capacity is unknown. One of my appliances has a finite session table on its management plane and wedges for minutes once it fills. Whether these hosts hold several concurrent DTLS associations is untested on hardware I have.

Third, pacing. It is session-scoped by construction: `_min_req_interval` and `_last_send_ts` are fields on `DtlsCoapSession`, and the 5 req/s default was set against ceilings I measured at roughly 14/s on my dryer and 8/s on my oven. Two sessions is already past the oven figure. Scoping the limiter to the host address, shared by every session to it, is the fix, and it belongs on my side of the line before any N>1 work.

`registry/subdevices.py` already binds several logical devices over one session, in three patterns. Pattern C came from the `AWM-WW-AID-26-ONEBODY` combo in #241 and sources sibling UUIDs from `/oic/res` link prefixes alone. @atc722's unit reports `AWM-KR-M64-24-WD86`, the same combo shape, so the first question is whether Pattern C already covers it. This PR cannot settle that: everything reported here is pre-authentication, and Patterns A to C were all derived from post-authentication resource shapes.

@atc722, read `docs/ocf-pki-laundry.md` in SmartThings-Local before more design work. §5 records the working carrier on this family as server-authenticated DTLS with no client identity leaf. The certificate path you tried is a different authentication mode from the one that works, so `unknown_ca` is what that mismatch looks like from the certificate side. §6 derives the OwnerPSK from that session's own master secret, so it is an output of the transaction. §4 is the part that blocks you, and it blocks everyone.

@Jason-Morcos, one question you can hopefully help with, whenever you have a moment. §7 of that doc records 39 complete protected representations on each validated WD53. Do those contain a second identity, either UUID-prefixed hrefs or a sibling `di`, or is it one flat namespace? Yours is `AWM-US-M64-24-WD80` against atc722's `AWM-KR-M64-24-WD86`, the closest post-authentication evidence anyone has. A flat namespace would mean the per-device client is solving something this family does not have.

Jason-Morcos:
> @Jason-Morcos, one question you can hopefully help with, whenever you have a moment. §7 of that doc records 39 complete protected representations on each validated WD53. Do those contain a second identity, either UUID-prefixed hrefs or a sibling `di`, or is it one flat namespace? Yours is `AWM-US-M64-24-WD80` against atc722's `AWM-KR-M64-24-WD86`, the closest post-authentication evidence anyone has. A flat namespace would mean the per-device client is solving something this family does not have.



For the two AWM-US-M64-24-WD80 units, it is one flat namespace.
Each has the same 39 protected representations. None of the hrefs is UUID-prefixed, no representation carries a sibling di, and there is no /device/1 or second device tree. The washer and dryer surfaces coexist under ordinary unqualified hrefs such as /st/washercourse/vs/0, /st/dryercourse/vs/0, /washer/vs/0, /course/vs/0, and /power/vs/0. The 72-link public directory is flat in the same way.
So Pattern C would find no subdevice in the WD80 data, and a client per advertised di would be solving a topology we did not observe on either unit.
I would keep that conclusion qualified to WD80, though. The multiple pre-authentication di containers on atc722’s WD86 may reflect a real regional or firmware difference, or they may collapse into one authenticated namespace. I think an authenticated WD86 /device/0?if=oic.if.b result is needed before choosing an N-session design. The closest post-authentication evidence we have says one host, one DTLS session, one OCF identity, and one combined washer/dryer namespace.

vmonkey:
​@atc722 In my case, the local device UUIDs match the Device IDs listed in the [SmartThings Advanced Web App](https://my.smartthings.com/advanced). How does that look for your multi-UUID device—does it show up as multiple device entries there for that single appliance?

​Also, if it would be helpful (and if you prefer a separate channel like Discord or email), I'd be happy to help you get through the local authentication / credential extraction part.

atc722:
Thanks @vmonkey . I checked the SmartThings Advanced Web App. My WD86 appears there as a single OCF device entry, not multiple entries. Its Device ID matches the SmartThings OCF di that I used to identify the washer during local discovery. I don’t see multiple device entries for this appliance in the web app.

I also really appreciate your offer to help with the local authentication and credential extraction. I’d like to review and try the documented path on my side first. If I’m still blocked after that, I’ll reach out so we can choose a convenient channel.

atc722:
Thanks again, @vmonkey , for offering to help. I’ve spent some time working through the documented authentication path, but I’ve realized that safely reproducing the owner-authorized transition and credential acquisition is likely to take me quite a while. Rather than make assumptions around an ownership-changing process, I would be very grateful for your guidance if your offer still stands.
Would it be okay if I contacted you at the email address listed on your GitHub profile?

vmonkey:
Yes, feel free to reach out.

atc722:
Thank you @mbillow, @QuiteYellow, @Jason-Morcos, and @vmonkey for the guidance on this draft.

The discussion here, together with the later authenticated, read-only WD86 result, showed that the useful next step is the smaller auth/session refactor discussed in #388 rather than extending this UUID-selection prototype.

I have opened #418 as that clean, no-behavior-change first step. It replaces this exploratory draft; OwnerPSK support and washer features remain separate follow-ups.

I am closing this draft to avoid leaving two active proposals for the same direction. Thank you again.

PR #407, opened 2026-08-23
Make the dryer's dry level a writable select

vmonkey:
## Summary

- Move the dryer's `dry_level` from a read-only sensor to a writable select, matching what washer/dryer combos have had since #33
- Add an `entity.select.dryer_dry_level` catalog across all seven locales
- Add a v4 → v5 config-entry migration that drops the entity-registry row the platform move orphans

Every dryer board in the corpus reports `x.com.samsung.da.supportedDryLevel` right next to `dryLevel` on `/washer/vs/0` — the same live option list `washer.py`'s combo select already writes against — but the dryer registry bound the field read-only. Both capabilities bind the same href and only one ever binds per device, so the asymmetry was historical rather than anything device-side.

## Why a separate translation key

`washer_dry_level`'s state table is that field's *other* meaning on a combo: a duration in minutes (`"30"` → `"30 min"`, `"240"` → `"4 hr"`). Reusing it would label a dryer's dryness setting in minutes, so the dryer gets `dryer_dry_level`.

The catalog covers the `Damp/Less/Normal/More/Very` vocabulary the TP1_21 dryers report. The numeric `None/1/2/3` that DV5000T (TP2_20) and DV6800N (A51_20) report is deliberately left uncatalogued, so `select._display` renders the digits raw — nothing confirms what they mean, and that's the same treatment unidentified course codes already get.

## Two deliberate differences from `washer.py`'s descriptor

**Field-gated, no `exists_fn`.** In `washer.py` the gate on `supportedDryLevel` discriminates a combo from a plain washer. Every dryer has a dry level, so the same gate here would only ever suppress the entity — including on a rep that is merely a stub at discovery, which `entity._is_included` deliberately admits so a sub-poll can populate it. Since `_is_included` is evaluated once at discovery, that would have been permanent for the life of the entry, and the migration below would already have removed the sensor.

**No `entity_category`.** Dry level is a per-load choice made alongside the cycle, not device configuration. It belongs in Controls next to the cycle select and `wrinkle_prevent`, which is also where the sensor it replaces already sat.

## ⚠️ Breaking change

`unique_id`s are scoped by (integration, platform), so a platform move cannot rewrite the old registry row — it can only be dropped.

- The v4 → v5 migration removes `sensor.<dryer>_dry_level`. Dashboards, automations, and the entity's rename/area need repointing at `select.<dryer>_dry_level`. The removal is logged at INFO naming the replacement, since it is irreversible.
- The entry version bump means **downgrading fails the entry outright** rather than losing a single entity, because the older release's guard is `entry.version > 4`. That is the downgrade guard working as designed, but it is worth weighing given 0.24.0 just shipped.

## Validation

- `ruff check custom_components tests`, `ruff format --check custom_components tests`, `ty check custom_components tests` — all clean
- 1740 tests pass, including the corpus-wide golden/unique-id/translation checks
- No golden file changes: the state key stays `dry_level` and `flatten()` is platform-agnostic, so the four existing dryer goldens are untouched
- Confirmed on a Samsung DV5000T (`DA_WM_TP2_20_COMMON`) through Home Assistant: the select lists the board's own `None/1/2/3`, and setting a level takes on the appliance. That board reports `isModelSettingWithoutSC` true, so the write lands with Smart Control off (see `common.remote_control_required_for_write`).

## Not included

`dry_time` stays a sensor. Dryer boards do report `supportedDryTime`, so a writable dry-time select is viable, but it is a separate change and would need its own entry migration for the same platform-move reason.

### DIFF
--- a/custom_components/localthings/__init__.py
+++ b/custom_components/localthings/__init__.py
@@ -193,6 +193,37 @@ def _relabel_particulate_statistics(hass: HomeAssistant, entry: ConfigEntry) -> 
     return True
 
 
+# Tail-matched for the same reason as _PARTICULATE_KEY_RE above. No
+# device-type gate: no registry binds a sensor-domain dry_level any more,
+# so the domain check plus this tail is already exact.
+_DRY_LEVEL_KEY_RE = re.compile(r"_dry_level(?:_\d+)?$")
+
+
+@callback
+def _drop_orphaned_dry_level_sensors(hass: HomeAssistant, entry: ConfigEntry) -> None:
+    """Remove the sensor row dryer.py's dry_level select supersedes.
+
+    unique_ids are scoped by (integration, platform), so a platform change
+    can't rewrite the old row -- it can only be dropped. dryer.py's select
+    is field-gated, not exists_fn-gated, precisely so it appears wherever
+    the sensor did and this removal is never left unmatched.
+    """
+    ent_reg = er.async_get(hass)
+    for registry_entry in er.async_entries_for_config_entry(ent_reg, entry.entry_id):
+        if registry_entry.domain != "sensor":
+            continue
+        if not _DRY_LEVEL_KEY_RE.search(registry_entry.unique_id):
+            continue
+        # INFO, not debug: this is irreversible and takes the user's rename,
+        # area and any automation reference with it.
+        _LOGGER.info(
+            "removing %s; dry level is now select.%s",
+            registry_entry.entity_id,
+            registry_entry.entity_id.partition(".")[2],
+        )
+        ent_reg.async_remove(registry_entry.entity_id)
+
+
 async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     """Migrate an entry to the current version.
 
@@ -214,11 +245,14 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     a UUID to rewrite *to*.
 
     The version is still bumped now rather than at that point, because the
-    bump's real job is the `entry.version > 4` downgrade guard below: an
+    bump's real job is the `entry.version > 5` downgrade guard below: an
     entry re-keyed onto a UUID and then loaded by a release that reads
     CONF_SERIAL as the key would silently orphan every entity it has.
+
+    v4 -> v5 drops the entity-registry row dryer.py's dry_level leaves
+    behind when it moves from a sensor to a select.
     """
-    if entry.version > 4:
+    if entry.version > 5:
         return False  # downgrade: this release doesn't know the newer shape
 
     if entry.version == 1:
@@ -257,6 +291,11 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             entry.entry_id,
             legacy_key,
         )
+
+    if entry.version == 4:
+        _drop_orphaned_dry_level_sensors(hass, entry)
+        hass.config_entries.async_update_entry(entry, version=5)
+        _LOGGER.debug("migrated entry %s to version 5 (dry_level moved to select)", entry.entry_id)
 
     return True
 
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -814,7 +814,9 @@ class LocalThingsConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
     # v4 keys the entry on the OCF device UUID (issue #381), which the probe
     # below resolves up front -- so a new entry is already on the v4 shape
     # and has nothing to re-key either.
-    VERSION = 4
+    # v5 drops the sensor-domain dry_level registry row a dryer's platform
+    # move orphans; a freshly created entry never had that row.
+    VERSION = 5
 
     def __init__(self) -> None:
         self._host: str = ""
--- a/custom_components/localthings/registry/capabilities/dryer.py
+++ b/custom_components/localthings/registry/capabilities/dryer.py
@@ -10,7 +10,7 @@ the /course/vs/0 cycle select -- lives in laundry.py.
 """
 
 from ..capability import Capability
-from ..entities import SensorDesc, SwitchDesc
+from ..entities import SelectDesc, SensorDesc, SwitchDesc
 from .common import diagnosis_status
 from .laundry import cycle_select, drum_clean_cycles_remaining, drum_clean_last_cleaned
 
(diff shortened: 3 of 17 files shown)


vmonkey:
I will get back to this soon. I will do another PR (decoding `supportedOptions` first to find which options are allowed for each program; that will be usable for other settings and devices as well).

mbillow (commented):
The entity change is right and I confirmed the central argument end-to-end. Discovering each laundry fixture through its resolved registry:

```
washer_dryer_onebody_awm  → dryer   dryer_dry_level   None/Damp/Less/Normal/More
dryer_tp1_21_drum_clean   → dryer   dryer_dry_level   None/Damp/Less/Normal/More
dryer_dve50a8600          → dryer   dryer_dry_level   None/Damp/Less/Normal/More/Very
dryer                     → dryer   dryer_dry_level   None/Less/Normal/More
dryer_dv6800n             → dryer   dryer_dry_level   None/1/2/3          (raw, by design)
washer_dryer_combo        → washer  washer_dry_level  None/Cupboard/30/…/240
```

The separate translation key is not just tidier, it's load-bearing: `washer_dryer_onebody_awm` is a combo that routes to the **dryer** registry and reports the word vocabulary, so reusing `washer_dry_level` would have labelled `Damp`/`Less`/`Normal` in minutes. Field-gating instead of `exists_fn` is also correct for the reason you give — every fixture has a populated `supportedDryLevel`, so the gate could only ever subtract. The migration test covers plain, subdevice-instanced, wrong-domain and wrong-tail rows, which is the right matrix.

## Don't bump the config entry version

You flagged this as worth weighing, and I'd go further: **there's no reason to bump it at all**, and the bump costs more than it looks.

The cost isn't losing one entity on downgrade. An older release evaluates `entry.version > 4`, returns `False`, and the **whole entry fails to load** — every entity on that device disappears, not just `dry_level`. For a single platform move, one release after 0.24.0 shipped, that's a steep price.

And the version gains nothing here, because the condition the migration keys off isn't a schema state — it's "a stale registry row may exist". That row can only ever be stale: after this PR nothing in the registry mints a sensor-domain `dry_level`, so the sweep is inherently idempotent and safe to re-run. A version gate on an idempotent cleanup is just a way to make it run exactly once, and running it every startup is cheap (one pass over the entry's registry rows, no I/O).

Concretely: keep `VERSION = 4` and the `entry.version > 4` guard exactly as they are, drop the `if entry.version == 4:` block, and call `_drop_orphaned_dry_level_sensors(hass, entry)` at the top of `async_setup_entry` — before the coordinator is constructed, alongside the existing `hass.data.setdefault(DOMAIN, {})`. It's a `@callback`, so it needs no await. Downgrade keeps working, upgrade behaves identically, and `test_migration_drops_the_superseded_dry_level_sensor` becomes a setup test rather than a migration one (the four-row matrix carries over unchanged).

That also lets you revert the churn in `test_identity_migration.py`, `test_statistics_migration.py`, `conftest.py` and the `version=5`→`6` future-version test, none of which is really about dry level.

## Smaller things

- **The log line invents an entity id.** `"dry level is now select.%s" % registry_entry.entity_id.partition(".")[2]` reuses the *sensor's* object id. After a user rename those diverge, so the message names a `select.` entity that doesn't exist — on the one log line whose whole job is telling the user where their entity went. Logging the removed `entity_id` and saying "dry level is now a select entity" avoids guessing.
- **`_DRY_LEVEL_KEY_RE` isn't device-type gated.** Exact today, as the comment says — I confirmed nothing else in the registry binds a sensor-domain `dry_level`. Worth noting the comment's reasoning quietly expires if a future device ever adds one, since this sweep keeps running. Moving it to setup makes that slightly more live, so the domain+tail check is worth a glance if that ever happens.

Validation reproduced: 1740 passed, `ruff check` / `ruff format --check` / `ty check` clean on py3.13, goldens untouched as claimed.

**Merge note:** conflicts textually with #409 in `tests/test_translations.py` — both append a test in the same spot. Trivial, but whichever lands second needs the rebase.

---

vmonkey:
Thanks — agreed on all of it, and the idempotency argument for dropping the version bump is convincing. I'll push the fixes shortly.

Before taking this out of draft, though, I'd like your read on where I was planning to go with it. #408 decodes `supportedOptions`, which tells us which options each program actually accepts. My plan was to land that first and then have this select's `options` follow the selected program, because today the entity offers levels the running course won't take and the board just ignores the write — no error, no state change, nothing the user can see.

Are you happy with options narrowing dynamically like that? It does mean `options` changes under the user at runtime, so a stored automation choice can fall outside the current list, and the entity's state can end up outside `options` on a program change until the board reports the new value. The alternative is keeping the full catalogue and surfacing the rejection some other way. I'd rather agree the shape before building it, and if you're happy with the dynamic version I'd keep this in draft until #408 lands rather than do the descriptor twice.

If that direction works, I'd like to follow the same pattern afterwards for the washer's spin/rinse/temperature and for the dryer's dry time, each in its own PR. Happy to rebase on #409 whenever it lands.

mbillow:
**Status:** #409 is merged. #408's decode is merged too, as #425 — it carries both of your commits unchanged plus the fixes from the second review round, so `course_option_mask` is on `main` and this PR is unblocked.

Three things in #425 that change what this select should do:

1. **The header nibble is the group count**, so the record width is stated rather than guessed — `1 + 2×hdr` bytes, true on all 17 dumps carrying a `supportedOptions`, and the split it gives is the one the smallest-that-parses scan already produced on every one (no course list moved, no golden changed). The `editCourseList` check went from a search to a check as a result, and it no longer returns `{}` when it can't be satisfied — a device listing a course `supportedOptions` doesn't carry would otherwise have switched the decoder off for every kind on that device, silently.
2. **`course_option_mask` matches a course code case-insensitively.** Worth knowing here specifically: `select.py` lowercases an entity's options for the translation catalog, so passing what the entity displays rather than the raw `Course_` token used to get a silent "no opinion".
3. **`0xD` is safe for *this* PR.** All five dumps routing to the dryer registry carry a `0xD` group; `washer_dryer_combo` is the `0xB` board and it routes to the washer registry. So the combo caveat bites `washer.py`'s `dry_level`, not this one.

## On narrowing options per course

Yes, go ahead — with three rules, because `select.py` computes `options` and `current_option` independently and HA's `SelectEntity.state` returns `None` when the current option isn't in the list:

- **Never let it go `unknown`:** `options = decoded ∪ {live value}`, so a course change landing before `dryLevel` catches up doesn't blank the entity.
- **Never narrow to empty:** an empty `_raw_options()` is the "unpopulated" contract that normally pairs with `exists_fn` suppression, and this descriptor deliberately has none — so empty gives a live entity with an empty dropdown rather than no entity. Decode returns `None` → full `supportedDryLevel`; course genuinely allows none → just the live value.
- Worth a line in the PR body when this comes out of draft: narrowing turns today's silently-ignored write into a visible `ServiceValidationError`. That's the win, but an automation that was quietly no-opping will start erroring, which is a behaviour change users will notice.

Same pattern afterwards for the washer's spin/rinse/temperature and the dryer's dry time sounds right, each in its own PR.

## Still outstanding from the first review

Independent of #408, so it can land whenever: keep `VERSION = 4`, drop the `if entry.version == 4:` block, and call `_drop_orphaned_dry_level_sensors` from `async_setup_entry` — the sweep is idempotent, so the version gate only buys running it once, at the cost of failing the whole entry on downgrade. And the log line reuses the sensor's object id to name a `select.` entity that may not exist after a rename.

## One process note

Apologies for not catching this before merging: `CONTRIBUTING.md` forbids AI co-author trailers, and the second commits on both #409 and #408 carried one. I've stripped them (`main` was rewritten once, identical tree) — if you rebase this branch, please drop the trailer from any new commits.

---

mbillow:
Thank you for this — and sorry it sat while the pieces it was waiting on landed. Closing it in favour of #442, which carries everything here that was still outstanding, but I want to be clear about what happened to each part.

**What had already merged.** #439 covered a DA_WM_TP1_21_COMMON washer/dryer pair and, in passing, made `dryLevel` a writable select on the same reasoning you gave here — `supportedDryLevel` right next to `dryLevel`, the same live option list `washer.py`'s combo select already writes against. It also did `dryTime`. That was landed without reference to this PR, which it shouldn't have been; the overlap is on me, not you.

**What was still missing, and is now in #442.** #439 landed the descriptor alone, so three things from here were still genuinely outstanding:

- The `dryer_dry_level` catalog. Without it the select shipped with a name and no state table, so the dropdown rendered a literal "None" instead of "Off", untranslated in every locale — on five of our seven dryer fixtures. **Your seven-locale commit is carried across as-is, with your authorship on it.**
- The sweep for the orphaned `sensor.<dryer>_dry_level` row, in the shape from my last review: `VERSION` stays 4, no `entry.version == 4` block, called from `async_setup_entry`. Extended to `dry_time`, which #439 also moved. Your point about the log line reusing the sensor's object id was right — it now names the removed row only.
- The per-course narrowing, with the three rules we agreed: `decoded ∪ {live value}`, never narrow to empty, and the `ServiceValidationError` note in the PR body. Built as a `laundry.course_narrowed_options` factory so the follow-ups can reuse it directly. On all five dryer fixtures it does the right thing — "None" leaves the dropdown, since no course accepts it.

I also ported your `test_presence_is_field_gated_not_narrowed_by_an_exists_fn`. It matters more in #442 than it did here: the sweep runs unconditionally on every setup rather than once in a migration, so a suppressed select would leave a user with neither entity, permanently and silently. Your reasoning for keeping the gate off is the thing holding that up.

**Two places #442 diverges from what you wrote**, both worth flagging since you may disagree:

- It keeps `entity_category="config"` from #439, matching `washer.py`'s `spin_level`/`rinse_cycles`, rather than your Controls argument. Yours isn't wrong — it's a per-load choice and that's where the sensor sat. Say so on #442 and I'll flip it.
- Your DV5000T confirmation is now recorded in the descriptor comment, which #439 had written as "inferred, not yet confirmed on a plain dryer". That's your hardware result doing the work; #438's ask on `dryLevel` is answered by it.

**What's left of the plan is genuinely new work**, not a rebase of this branch: the washer's spin/rinse/temperature and the dryer's dry time, each in its own PR, reusing `course_narrowed_options`. Those would be very welcome if you still want them — and the `supportedOptions` decode you did in #408 is what made the narrowing possible at all, so thank you for that too.

---

PR #408, opened 2026-08-23
Decode what each course allows from supportedOptions

vmonkey:
## Summary

- Decode the payload behind each `supportedOptions` record, not just the course code it starts with
- Name four option kinds — wash temperature, rinse, spin and dry — each against a device that could have contradicted them
- Fold the existing record splitting into one shared helper, and use a live `editCourseList` to pin the record width where one exists

`/course/vs/0`'s `supportedOptions` is already split up in `laundry.py`, but only far enough to recover the course codes; the payload behind each one is dropped. That payload is what says which wash temperatures, rinse counts, spin speeds and dry levels a given course actually permits — the thing you need to stop offering a dry level on a course that has none, or a 95° wash on a course that caps at 40°.

Three separate entities are about to want it, so it is decoded once here rather than three times in three places.

## The format

```
supportedOptions = <1 nibble header> + one record per course
record           = <course code:1B> then 2-byte groups
group            = <kind nibble><default nibble> <mask>
```

The mask indexes that option's own `supported<Option>` list, and so does the default nibble — but into the list, not into the mask: a dishwasher reports default `0` alongside a mask allowing only index `1`.

Every dump in the corpus fits `1 + 2×groups` bytes per record, from an AirDresser carrying no groups at all to an 11-byte WA8000T record.

## Why these four kinds and not others

| Kind | Named | Evidence |
|---|---|---|
| `0xD` | dry | A DV5000T's fourteen records reproduce exactly what its panel offers per course — 1/2/3 on Cotton, only 2 on Towels, only 1 on Wool, none on Quick Dry or the three air/time courses. Corroborated by the DV6800N on a different board family and code space |
| `0x8` | water temperature | On a WW6500 only this kind reaches bit 6, and `supportedWaterTemperature` is the only list seven entries long |
| `0x9` | rinse | Its owner's panel reading pins the set; `0xA` is ruled out as rinse by the row below, leaving this one |
| `0xA` | spin | On the `washer` dump `0xA` addresses index 6, which a six-entry `supportedRinseCycles` cannot hold — so `0xA` is not rinse, whatever an owner reports |

Corroborating all four: across every dump here, each live value sits inside the set its course's mask decodes to.

One thing worth being precise about, since it is easy to overstate: the WW6500 panel reading pins the *sets*, not which of `0x9` and `0xA` is which. On both courses that match it the two masks are `0x3F` alike, so swapping the constants leaves that reading satisfied. The list-length argument in the `0xA` row is what actually separates them, and it is asserted.

`0x0`, `0x5`, `0x6`, `0x7` and `0xC` stay unnamed — nothing pins them. So does `0xE`: it behaves like dry time on the one board carrying it, but no dump here proves it, and it can be named by the change that consumes it. `0xB` also stays unnamed, but not for want of evidence — see below.

A kind is not an entity name, either. `0xD` is "dry", not "dry level" — dishwashers carry it with no `supportedDryLevel` at all, since their dry setting is `heated_dry`. Callers key the mask against their own list rather than assuming what it counts.

## `0xB` is the dry dial on one board, and `0xD` is absent there

Six dumps carry a `supportedDryLevel`; five route to the dryer registry. `washer_dryer_combo` — a WW6600R on `DA_WM_TP2_20_COMMON`, eight entries — is the one routing to the **washer** registry, which is why `washer.py`'s `dry_level` select is the entity at stake. It carries **no `0xD` group at all**, and its `0xB` is unmistakably the dry dial:

| courses | default | allowed |
|---|---|---|
| wash+dry (`1C`, `1B`, `1E`, …) | None | None / Cupboard / 30 / 60 / 90 |
| dry-only (`36`, `38`, `39`) | Cupboard | Cupboard / 30 / 60 / 90 |
| wash-only (`24`, `30`, `32`, …) | None | *(empty)* |

`0xB` is still left unnamed, because what the corpus supports is narrower than "`0xB` is dry": `washer_dryer_onebody_awm` is also a combo and uses `0xD`, and the DV5000T that `0xD` was named on is `DA_WM_TP2_20_COMMON` — this WW6600R's own family. So it is neither a combo rule nor a board-family rule.

**The consequence to remember:** a gate keyed on `0xD` alone silently no-ops on the one board whose `dry_level` select it would be narrowing. Pinned by a test so the next change cannot miss it.

## Two caveats worth knowing before building on this

**A mask is not the appliance's last word.** A WW6500 offers only 2–5 rinses on Baby Care — on its own panel, not just in SmartThings — where the mask allows all six. Its other thirteen courses match dial for dial, default included. Gating on a mask can therefore offer slightly more than the machine really takes. That is the safe direction, since the device refuses what it will not do, but it is not a guarantee.

**An empty mask is not automatically "nothing selectable."** On a dryer's Quick Dry it really is none. But a washer's cloud `Download` slot reports empty masks for all three of its kinds while still holding live values, because the downloaded program supplies its own. Deciding between those is the caller's job, per entity, with its own evidence.

Both are recorded in the module comment rather than only here.

## Refactor half

`_course_codes_from_supported_options` keeps its behaviour exactly and now shares `_course_records`, so the course list and the payload can never end up read at different widths.

It also gains a guard that was free and unused: where the device publishes a live `editCourseList`, the split has to account for every code on it. That pins the width outright instead of leaning on the smallest-that-parses heuristic — the ONEBODY combo passes the old guards at four different widths, and only one accounts for its twenty listed codes.

## Validation

- `ruff check` / `ruff format --check` / `ty check` clean; 1748 tests pass
- No behaviour change: the golden files that depend on the course-list fallback are untouched
- Corpus-wide invariants added over every shipped dump — record shape, masks staying inside their own `supported*` list, and every live temperature/rinse/spin/dry value falling inside its course's decoded set
- The three guards that no fixture can exercise (the `editCourseList` width check, the group-scan stride, and the top mask bit) are pinned by constructed records; each was verified to fail the suite when the corresponding line is mutated

## Review follow-up

Addressing @mbillow's review, in a second commit:

- **`0xB` on the combo is recorded**, in the module comment, in the new investigation doc, and pinned by a test rather than left as prose. Independently re-derived before writing it down, along with two details the review did not have: the combo's `supportedDryLevel` mixes a dryness level with durations (`Cupboard`, then 30…240), which is presumably why that board has no separate `0xE` group; and no course addresses past index 4, so `120`, `180` and `240` are advertised by nothing that offers them.
- **The evidence narrative moved to `docs/investigations/course-option-groups.md`**, per `CONTRIBUTING.md`. The block comment is 56 → 34 lines; the two caveats stay inline. The doc also records the `QuickWashSet_5B847E933FA53F` corroboration from the review — verified byte-identical to course `5B`'s record.
- **The rinse/spin discrimination is now asserted structurally**, not left resting on one reading.
- **The stale `TestCourseOptionGroups` docstring** is fixed, and the five local `tests.conftest` imports are hoisted to module level.

Two corrections to the review's own framing, both verified: swapping the rinse and spin constants does **not** fail the WW6500 panel test (their masks are `0x3F` alike, so it passes either way) — the test it failed was the corpus live-value check. And the combo is not "the only board with a real multi-entry `supportedDryLevel`" — six dumps carry one, at lengths 4 to 8; what makes it the board that matters is that it is the only one routing to the washer registry.

Not done, per the review's own framing: caching `course_option_mask`'s divisor scan. It is premature while nothing consumes the decoder, and belongs in the change that adds the third caller.

## Not in this change

Nothing consumes the decoder yet — it is deliberately groundwork plus its evidence, so the decode gets reviewed once rather than three times. The dryer's dry level and dry time come next, and per the above should key on the kind the board actually carries rather than on `0xD`. The washer's own wash options are the obvious third, though narrowing what they offer is a behaviour change that deserves its own decision rather than being assumed here.

### DIFF
--- a/custom_components/localthings/registry/capabilities/laundry.py
+++ b/custom_components/localthings/registry/capabilities/laundry.py
@@ -250,6 +250,158 @@ def drum_clean_last_cleaned(rep):
         return None
 
 
+# ---------------------------------------------------------------------------
+# supportedOptions -- what each course permits.
+#
+# A 1-nibble header followed by one fixed-width record per selectable
+# course. A record is that course's own hex code, then zero or more 2-byte
+# groups; every fixture in the corpus fits `1 + 2*groups` bytes, from an
+# AirDresser carrying no groups at all to an 11-byte WA8000T record.
+#
+#   <hdr:1 nibble> ( <course:1> ( <kind:nibble><default:nibble> <mask:1> )* )*
+#
+# The mask indexes that option's own supported<Option> list, and so does
+# the default nibble -- but into the list, not into the mask: a dishwasher
+# reports default 0 with a mask allowing only index 1.
+#
+# Four kinds are pinned, each against a device that can contradict them:
+#
+#   0xD dry     A DV5000T's fourteen records reproduce exactly what its
+#               panel offers per course -- 1/2/3 on Cotton, only 2 on
+#               Towels, only 1 on Wool, none on Quick Dry or the three
+#               air/time courses.
+#   0x8 temp    A WW6500 owner read one course's three dials off the
+#   0x9 rinse   panel: Cold/20/30/40 with no "None" offered, every rinse
+#   0xA spin    count including none, and every spin including rinse-hold.
+#               Its masks are 0x1E/0x3F/0x3F, which decode to exactly
+#               those sets against supportedWaterTemperature (0x1E skips
+#               bit 0, the "None" the panel indeed omits),
+#               supportedRinseCycles and supportedSpinLevel. Only 0x8
+#               reaches bit 6 anywhere in that table, which none of the
+#               other two lists is long enough to hold.
+#
+# Corroborating all four: across every dump here, each live value sits
+# inside the set its course's mask decodes to. 0x0/0x5/0x6/0x7/0xB/0xC
+# stay unnamed -- nothing pins them, and 0xE looks like dry time on the one
+# board that carries it but no dump proves it.
+#
+# The kinds are not entity names. 0xD is "dry", not "dry level":
+# dishwashers carry it with no supportedDryLevel at all (their dry setting
+# is heated_dry). A caller keys the mask against its own list rather than
+# assuming what it counts.
+#
+# A mask is what supportedOptions advertises, and an appliance can enforce
+# more than it advertises: a WW6500 offers only 2-5 rinses on Baby Care --
+# on its own panel, not just in SmartThings -- where the mask allows all
+# six. Its other thirteen courses match dial for dial, default included.
+# So gating on a mask can offer a little more than the machine really
+# takes. That is the safe direction, since the device refuses what it will
+# not do, but it is not a guarantee and an entity should not present a
+# mask as the appliance's last word.
+#
+# An empty mask means the device declares no range, which is NOT
+# automatically "nothing is selectable". On a dryer's Quick Dry it really
+# is none (panel-confirmed above), but a washer's cloud Download slot
+# reports empty masks for all three of its kinds while still holding live
+# values, because the downloaded program supplies its own. Deciding
+# between those is the caller's job, per entity, with its own evidence.
+# ---------------------------------------------------------------------------
+
+OPTION_KIND_WATER_TEMPERATURE = 0x8
+OPTION_KIND_RINSE = 0x9
+OPTION_KIND_SPIN = 0xA
+OPTION_KIND_DRY = 0xD
+
+
+def _course_records(course_rep, must_cover=None):
+    """{course code: record hex} from supportedOptions, or {} if unreadable.
+
+    Record width is recovered the way it always has been -- see the guards
+    in _course_codes_from_supported_options, which this backs.
+
+    `must_cover` adds one more: the split's course codes must include every
+    code given. Callers that can see a live editCourseList pass it, which
+    pins the width outright on every dump here that has one; the
+    course-list fallback itself cannot, since it only runs when that list
+    is missing in the first place.
+    """
+    raw = course_rep.get("x.com.samsung.da.supportedOptions")
+    hexstr = raw[0] if isinstance(raw, list) and raw else raw
+    if not isinstance(hexstr, str) or len(hexstr) < 3:
+        return {}
+    body = hexstr[1:]
+    if len(body) % 2:
+        return {}
+    total_bytes = len(body) // 2
+    current = option_value(course_rep.get("x.com.samsung.da.options"), "Course")
+    for k in range(1, total_bytes + 1):
+        if total_bytes % k:
+            continue
+        n = total_bytes // k
+        if n < 2:
+            continue
+        records = [body[i * k * 2 : (i + 1) * k * 2] for i in range(n)]
+        firsts = [r[:2] for r in records]
+        if len(set(firsts)) != n:
+            continue
+        if current is not None and current not in firsts:
+            continue
+        if must_cover and not must_cover <= set(firsts):
+            continue
+        return dict(zip(firsts, records, strict=True))
+    return {}
+
+
+def course_option_mask(resources, kind, course=None):
+    """(default index, allowed indices) for `kind` on the selected course.
+
+    None when this device says nothing usable -- no supportedOptions, an
+    unrecognized course, or no group of that kind on the record. Callers
+    treat that as "no opinion" rather than "nothing allowed": several
+    boards carry supported<Option> lists with no supportedOptions groups at
+    all, and refusing everything there would be worse than not gating.
+
+    The default index is into the same supported<Option> list as the
+    allowed indices, but it is NOT necessarily one of them -- a dishwasher
+    reports default 0 alongside a mask allowing only index 1. Callers that
+    want a fallback value have to decide for themselves whether an
+    out-of-set default is usable.
+
+    Where the device also publishes an editCourseList, the split has to
+    account for every code on it (see _course_records). That check is free
+    here and it is the difference between mis-gating an entity silently and
+    declining to gate it at all.
+    """
+    course_rep = resources.get("/course/vs/0") or {}
+    edit_list = parse_edit_course_list(
+        (resources.get("/wm/editcourse/vs/0") or {}).get("x.com.samsung.da.editCourseList")
+    )
+    records = _course_records(course_rep, must_cover=set(edit_list))
+    if not records:
+        return None
+    if course is None:
+        course = option_value(course_rep.get("x.com.samsung.da.options"), "Course")
+    record = records.get(course)
+    if record is None:
+        return None
+    for i in range(2, len(record), 4):
+        group = record[i : i + 4]
+        if len(group) < 4:
+            break
+        try:
+            head = int(group[:2], 16)
+            mask = int(group[2:], 16)
+        except ValueError:
+            # Not hex after all, so this split is not a record table --
+            # the width guards alone can't tell. Say nothing rather than
+            # raise into whatever entity asked.
+            return None
+        if head >> 4 != kind:
+            continue
+        return head & 0xF, [bit for bit in range(8) if mask >> bit & 1]
+    return None
+
+
 def _course_codes_from_supported_options(course_rep):
     """Fallback for an empty/missing editCourseList: derive the selectable
     course list from /course/vs/0's own supportedOptions instead (issue #1:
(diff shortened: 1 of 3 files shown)


mbillow (commented):
Reviewed by re-deriving the decode independently over all 27 `supportedOptions` fixtures rather than reading the write-up, then mutation-testing the guards. The decode holds up well, and the tests are doing real work.

**Mutation results.** All three guards the PR claims are pinned do fail when mutated. I also mutated things the PR *doesn't* claim: swapping `OPTION_KIND_RINSE`/`OPTION_KIND_SPIN`, changing `WATER_TEMPERATURE` to `0xC`, changing `DRY` to `0xB`, dropping the `current not in firsts` guard, and reading the default nibble as `head >> 4`. Every one fails the suite. The kind assignments are genuinely tested, not merely asserted.

**One corroboration you didn't claim.** The WW6500's `QuickWashSet_5B847E933FA53F` option token is a standalone 7-byte record for course `5B`, byte-identical to that course's record in the table. That pins both the record width and the group structure from a completely separate field.

## `0xD` is not the dry kind on a combo — `0xB` is

This is the one finding I'd want addressed before merge, not because anything here is wrong but because the comment's framing will mislead the next reader.

`washer_dryer_combo` is the only board in the corpus with a real multi-entry `supportedDryLevel` (8 entries). It carries **no `0xD` group at all**. It carries `0xB`, and decoded against `supportedDryLevel` that group is unmistakably the dry dial:

| courses | default | allowed |
|---|---|---|
| wash+dry (`1C`, `1B`, `1E`, …) | `None` | None / Cupboard / 30 / 60 / 90 |
| dry-only (`36`, `38`, `39`) | `Cupboard` | Cupboard / 30 / 60 / 90 |
| wash-only (`24`, `30`, `32`, …) | `None` | *(empty)* |

The dry-only courses dropping `None` and defaulting to `Cupboard` is exactly the semantics you'd want, and `default == live dryLevel` holds. That's the same quality of evidence used to name `0xD` on the DV5000T.

Meanwhile the dishwasher — cited in the comment as support for `0xD` ("dishwashers carry it with no `supportedDryLevel` at all") — carries `0xB` and `0xD` with **byte-identical payloads in every record** (`B002 D002`, `B000 D000`; the cloud board adds a matching `0xC`). It cannot discriminate `0xD` from its neighbours, so it's the weakest evidence in the table while reading as confirming.

Leaving `0xB` unnamed is the safe call and I'm not asking you to name it. But the roadmap here is "the dryer's dry level and dry time come next," and a gate keyed on `0xD` alone will silently no-op on combos — which is also the one board family where `washer.py` already ships a writable dry-level select that would want narrowing. Worth a line in the comment recording what `0xB` does on the combo, so it isn't re-derived from scratch later.

## Comment length vs. CONTRIBUTING

`CONTRIBUTING.md` is unusually explicit on this ("keep the conclusion; cut the journey", "don't reproduce the full investigation — every dump checked, every hypothesis considered"). The new block comment is 59 comment lines against 86 lines of code and reproduces most of the PR body.

The two caveats earn their place inline — a mask isn't the appliance's last word, and an empty mask isn't automatically "nothing selectable". Both are things a caller gets wrong without them. The per-kind evidence narrative ("1/2/3 on Cotton, only 2 on Towels, only 1 on Wool…") is the journey, and the repo already has `docs/investigations/` for exactly this kind of record. Moving it there keeps the evidence without putting a page of prose between the reader and `_course_records`.

## Stale test docstring

`TestCourseOptionGroups` says:

> Only kind 0xD is claimed. … The washer kinds 0x8/0x9/0xA line up with temperature/rinse/spin by shape only, so nothing here names them.

The module names all four, and `test_confirmed_ww6500_course_matches_its_panel` panel-confirms them. This reads like it survived an earlier revision of the branch, and it understates the evidence you actually have.

## Smaller things

- **Rinse/spin discrimination rests on one test.** Swapping the two constants fails exactly one test (the WW6500 panel reading). The `washer` fixture gives you an independent *structural* argument: `0xA` reaches bit 6 there, while `supportedRinseCycles` is only six entries long — so `0xA` cannot be rinse, no owner report needed. Cheap to assert alongside the existing overflow check.
- **`from tests.conftest import _load_device` appears locally in five methods.** The file convention (and 40-odd other test modules) is a module-level import.
- **Non-blocking:** `course_option_mask` re-runs the divisor scan on every call — ~132 µs on the combo's table. Irrelevant now, and consistent with how `cycle_options` already works, but worth a cache once three entities are calling it per update.

Validation reproduced: 1746 passed, `ruff check` / `ruff format --check` / `ty check` clean on py3.13.

---

vmonkey:
Thanks for the feedback, it is fixed now.

vmonkey:
## How far do we trust the masks? — evidence from Samsung's own app

While validating the decode I dug into the official SmartThings washer plugin (`com.samsung.android.plugin.washer`, the AngularJS app Samsung ships for these boards). Notably, **Samsung does not trust its own `supportedOptions` table** — the app keeps a hardcoded patch layer on top of it, keyed by model-family string and course table generation:

- **Drum clean (our course `63`)** — `checkSpecialEcodrum()` forces the  temperature (`setTemp = "70"`) whatever the device reports, and on course select there is an explicit *"if the set reports the drum-clean default as 60°, force it to 70°"* branch (`isExceptTempCondidion`).
- **Baby Care (our course `5F`)** — `checkSpecialCase()` disables rinses 0–1 at 90°/95° and forces the rinse count (to 2, or even 4 on some EU families) — i.e. exactly the course where our corpus notes the  mask is wider than the panel.
- On top of that: US steam courses force temp + max soil with both controls disabled; soil level silently rewrites temp on US lifestyle courses; spin/dry/rinse cross-locks (rinse=0 + RinseHold + no dry is refused); dry-combo forces spin to its max; cloud Download courses get *all* option lists blanked; and dual-tub models get a whole extra interlock matrix (sanitize running on one tub caps the other tub to < 60°, forces Baby Care to 40°, and disables ~18 courses outright).

So the mismatch we see on Baby Care / Drum Clean is not a decode error and not a deferred device quirk — **the table is only ever a superset hint; the real per-course rules live in the appliance MCU, with the SmartThings app approximating them again in hardcoded JS per model family.** The masks stay exactly what this PR names them: an outer envelope, with the device as the final arbiter.

That leaves a design question for the integration, and I'd like your call, @mbillow: when a course-conditional entity lands (wash temp, rinse, spin, dry), do we

1. gate purely on the mask and let the device refuse/clamp the rest (what the mask is good at, zero maintenance, and the failure direction is safe), or
2. replicate Samsung's patch table for the few known courses (drum-clean temp floor, baby-care rinse floor at 90°+, steam-model soil/temp forcing) — nicer UX, but it is model-family×table-version keyed data we would have to source dump-by-dump and maintain against firmware changes.

mbillow:
Landed in #425 (merged as `20b479b`) rather than from this branch, so closing here — but the work is yours and both of your commits are on `main` with their authorship intact:

- `6d6aba5` Decode what each course allows from supportedOptions
- `ca40347` Address review: record what 0xB does, and move the evidence to docs

Thank you for this one. The decode holds up under every check I could think to throw at it, the `0xB` write-up is the kind of negative result that saves the next person a day, and being straight about what the WW6500 panel reading does and doesn't pin — rather than letting it carry more weight than it can — is what made the rest trustworthy.

Two reasons it was carried rather than merged directly. The second commit carried a `Co-Authored-By: Claude` trailer, which `CONTRIBUTING.md` forbids, so the message needed rewriting (content untouched — that's why the SHAs above differ from this branch's). And a third commit applies the fixes from the last review round, which was easier to do on top than to ask you for another push:

- **The header nibble is the group count**, so the record width is stated rather than guessed: `1 + 2×hdr` bytes, true on all 17 dumps carrying a `supportedOptions`, across all six widths. The split it gives is the one your scan already produced on every single dump, so no course list moved and no golden changed. The scan stays as the fallback and the stated width is still checked against your guards rather than trusted. Your own constructed fixture corroborates it, incidentally — the payload in `test_edit_course_list_overrides_a_narrower_passing_split` has header `1`, and three bytes is exactly the width the `editCourseList` recovers there.
- **`must_cover` no longer returns `{}` when it can't be satisfied.** That was the one thing in the change I'd have called a hazard: a device listing a course its `supportedOptions` doesn't carry — a personal or cloud slot, a stale list — would have switched the decoder off for every kind on that device, silently, which is the failure mode the module warns about most loudly elsewhere. A covering split still wins; failing that the smallest passing one stands. On the corpus the guard changes nothing either way (I checked all 8 dumps with a non-empty edit list), so this only ever mattered for a board none of us has seen.
- **`course_option_mask` matches a course code case-insensitively**, since `select.py` lowercases an entity's options for the catalog.
- Two corrections: the one-byte mask can't address past index 7, and three boards carry an 11- or 13-entry `supportedDryTime` (none of them with a group for it) — worth knowing before dry time gets gated. And bit 7 isn't "unexercised anywhere in the corpus" as the stride test said: `washer_wa8000t`'s unnamed `0x6` reaches it with mask `0xA1` on eight of its thirteen courses. No *named* kind does.

Each of those is pinned by a test that fails when the change is reverted. #407 is unblocked and has a reply on the dynamic-options question.

---

Issue #343, opened 2026-08-09
Washer cycles swapped


Dillton:
### Appliance model

DA_WM_TP1_21_COMMON (WW11BB744DGES7)

### What's missing or wrong

Hi, I've noticed that `Towels` and `Bedding` cycles are swapped means that `selecting Towels in HA selects Bedding on Washer and vice versa`.

Checked all other cycles and only these two are wrong plus checked my Dryer (DA_WM_TP1_21_COMMON (DV90BB7445GES7)) cycles and there it's correct. 😊 
### Diagnostics file

[config_entry-localthings-01KZ1JPBDWCNS5KMF3GMAGT3CP.json](https://github.com/user-attachments/files/30873783/config_entry-localthings-01KZ1JPBDWCNS5KMF3GMAGT3CP.json)

mbillow:
Fix is up in #344 — swapped the `24`/`33` labels back in `washer_cycle_table_02` across every locale catalog, confirmed against your diagnostics dump (`Course_24` selected, `Table_02`). Will close automatically when that merges.

---
_Generated by [Claude Code](https://claude.ai/code)_

mbillow:
Confirmed fixed and released in [v0.21.0](https://github.com/mbillow/localthings/releases/tag/v0.21.0) — `24`/`33` read Bedding/Towels correctly now, checked across every locale catalog. Update via HACS to pick it up.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #409, opened 2026-08-23
Add a WW6500's Table_00 washer course names

vmonkey:
## Summary

- Label the fourteen Table_00 courses a WW6500 (`DA_WM_A51_20_COMMON`) selects, none of which had a name — every one rendered as raw hex in the cycle select
- Twelve reuse a string the catalog family already carries for the same cycle, so every locale comes from an existing confirmed translation
- Two are new wording, one of them deliberately *not* shared with a similar-sounding Table_02 label

The codes are `5B`–`6B`, disjoint from the `01`/`55`–`78` codes another Table_00 device already contributed, so nothing here is a relabel.

| Code | Label | Code | Label |
|---|---|---|---|
| `5B` | Cotton | `63` | Drum Clean |
| `5C` | Extra Speed | `64` | Rinse + Spin |
| `5D` | Super Eco Wash | `65` | Wool |
| `5E` | Delicates | `66` | Bedding |
| `5F` | Baby Care | `67` | Synthetics |
| `60` | Outdoor | `68` | E Cotton |
| `61` | Dark Garment | `6B` | Spin Only |

## How the codes were assigned

Not by the order the owner happened to list them in. They reported, per course, the temperatures, rinse counts and spin speeds their machine offers plus each default — and **twelve of the fourteen match what the device's own `supportedOptions` advertises on all six values**, which pins the mapping one-to-one.

The distinctive ones carried the assignment: Drum Clean is locked to a single value on all three dials, Spin Only is the only course with no temperature and no rinse at all, Super Eco Wash is cold-only, E Cotton is exactly 40/60.

The two that did not match cleanly are explained rather than waved through:

- **`5C`** — its rinse "default" was read while it was the *selected* course, so the app showed the current setting (3) rather than the course default (2).
- **`5F` Baby Care** — the machine offers 2–5 rinses where `supportedOptions` advertises all six. Confirmed on the appliance's own panel, not just the app.

## Wording

Twelve reuse an existing catalog string for the same cycle, which keeps all seven locales on confirmed translations instead of fresh guesses. Nine of those borrow from the sibling `washer_cycle_table_02` — only `5E`, `64` and `66` have a same-cycle twin inside `washer_cycle_table_00` itself (`75`, `73`, `71`), so this is mostly cross-table reuse rather than a table that already carried these labels. A good check on the wording either way: the two courses the owner happened to give in Czech — *Máchání + odstřeďování* and *Čištění bubnu* — came back word for word what the catalog already had.

Only `5C` and `61` are new:

- **`5C` Extra Speed / `Extra rychlý`** — deliberately does **not** share Table_02's `Super Speed` / `Super rychlé`. This app calls it *Extra rychlý*, a different string, and treating two codes as the same cycle because they sound alike is how `24`/`33` got transposed in #343.
- **`61` Dark Garment / `Tmavé prádlo`** — nothing comparable existed in the catalog.

## Validation

- 1732 tests pass; `ruff check` / `ruff format --check` / `ty check` clean
- All fourteen codes resolve through the cycle select — and this is now asserted, not just checked by hand: the code set is derived from the shipped `washer_ww6500` fixture rather than retyped, so a course list that grows, or picks up a code the catalog has no label for, fails instead of rendering as raw hex
- The translation test follows the existing `test_confirmed_*` pattern and pins the twelve shared labels **in every locale**, not just English: a locale that translated one differently from the code it shares a meaning with would still pass the key-topology test, which is the gap #343 fell through
- It also asserts `5C` stays distinct from Table_02's `1D`, so a future tidy-up cannot quietly merge them

## Review follow-up

Addressing @mbillow's review, in a second commit:

- **The headline claim is now locked by a test.** `test_ww6500_table_00_course_codes_are_translated` derives the codes from the fixture via `cycle_options()` and also asserts the table id the device itself reports. Both directions verified by mutation: dropping `5b` from `en.json` fails the subset assert, and appending a fifteenth record to the fixture's `supportedOptions` fails the count.
- **The "twelve reuse" claim is corrected** here and in the test docstring, which also had the count wrong at eleven.
- **`washer_cycle_table_00` is now sorted by code** in all seven locales, not just the new entries slotted in — the existing eleven were in reporter order with no ascending run to insert into, and `washer_cycle_table_02` is the precedent as the one other sorted table. Verified as a pure reorder: each file's parsed JSON is unchanged.
- **The rinse-hold claim above is corrected.** Drum Clean (`63`) also excludes rinse-hold; `6B` is uniquely "no temperature, no rinse, spin range".

Not done, deliberately: the three intra-table twins (`5e`/`75`, `64`/`73`, `66`/`71`) agree in every locale today but are left unasserted. They are two different devices' names for the same-sounding cycle, and the `dryer_cycle_table_00.9a` vs `washer_cycle_table_02.1b` divergence noted in the review is the case for not assuming families must agree — pinning it would block a legitimate future correction to `75`. The Table_02 pinning is what carries the guarantee.

The commit message of the first commit is left as written; the second commit records the corrected topology.

## Provenance

Names were read from the SmartThings app, which is **not** a faithful mirror of the appliance — on this same machine it applies a rinse floor the API does not carry. Course names are lower risk than option ranges, but a later report from the same model should win over these.


mbillow (commented):
Checked the assignments against the shipped `washer_ww6500` fixture's own `supportedOptions` rather than the write-up. The identification holds up — the distinctive courses are exactly as distinctive as you say:

| course | record groups | reading |
|---|---|---|
| `63` Drum Clean | `8520 9204 A204` | single value on all three dials |
| `5D` Super Eco Wash | `8102` | mask `0x02` = Cold only |
| `68` E Cotton | `8430` | mask `0x30` = exactly 40/60 |
| `6B` Spin Only | `8000 9000 A53E` | no temperature, no rinse, spin minus rinse-hold |

`5C`'s rinse group is `923F` — default 2 against a live `rinseCycles` of 3, precisely the "read while selected" discrepancy you describe. `61` defaulting to 4 rinses is a good independent signal for Dark Garment. The reasoning is honest and the conclusions follow from the data.

Three things worth fixing.

## The headline validation claim isn't locked by a test

> All fourteen codes now resolve through the cycle select, verified against the shipped `washer_ww6500` fixture — none falls through to raw hex

I confirmed this is **true** — `cycle_options()` returns all fourteen and every one resolves. But nothing in the diff tests it. `test_reported_table_00_course_codes_are_translated` uses a hardcoded set from the WF45R6300AW (`01`, `55`, `57`, `70`–`78`) that doesn't include any of `5b`–`6b`, and the new test only checks label *consistency* for the shared twelve plus wording for `5c`/`61`. Nothing derives the code set from the fixture, so a fifteenth code or a changed course list goes unnoticed.

A fixture-derived assertion in the existing `confirmed <= translated_states(...)` shape would close it, and passes as-is today:

```python
codes = {c.lower() for c in cycle_options(_load_device("washer_ww6500"))}
assert codes <= translated_states("select", "washer_cycle_table_00")
```

Worth noting this path runs through `_course_codes_from_supported_options`, which #408 refactors — so this also pins the two PRs together.

## "Twelve reuse a string this catalog already carries" is stronger than it reads

Only **three** of the new codes have a twin inside `washer_cycle_table_00` itself (`5e`↔`75`, `64`↔`73`, `66`↔`71`). The other nine borrow from `washer_cycle_table_02`. That's completely fine — and the test is good, because it pins each code to a *specific* Table_02 twin in every locale rather than just matching English — but the PR body reads as though table_00 already carried these labels.

Related: the test docstring says "Eleven reuse a string" while `shared` has twelve entries. Looks like `5d` was added later and the count wasn't updated.

Two things this surfaced, neither yours to fix but both now inherited:

- `washer_cycle_table_02` already has `0d` and `79` both as "Spin Only" in English with **different** Korean strings (`탈수만` vs `탈수단독`). `6b` adopts `0d`'s. Fine, just no longer arbitrary once a third code points at it.
- Same English label diverges across families elsewhere — `dryer_cycle_table_00.9a` is `면` where `washer_cycle_table_02.1b` is `면의류`. You picked the washer side every time, which is right; the test locking the specific twin is what makes that durable.

## Ordering

The new codes are appended after `78` in all seven locales, so `washer_cycle_table_00` now reads `01, 70, 55, 71, 72, 77, 57, 73, 74, 75, 78, 5b, 5c, …`. The existing table isn't sorted either, so this is pre-existing, but on a catalog this long sort order is how you find a code — inserting `5b`–`6b` in place would be a small improvement while you're in here.

## One prose correction

> Spin Only is the only course with no rinse-hold

Drum Clean (`63`) also excludes it — its spin mask is `0x04`, bit 2 only. Doesn't weaken the identification at all, since `63` is single-valued on every dial and `6B` is uniquely "no temperature, no rinse, spin range". Just worth not carrying an over-strong claim into the commit message.

Validation reproduced: 1731 passed, `ruff check` / `ruff format --check` / `ty check` clean on py3.13.

**Merge note:** this conflicts textually with #407 in `tests/test_translations.py` — both append a test in the same spot. Trivial to resolve, but whichever lands second needs the rebase.

---

vmonkey:
Thanks for the feedback, it is fixed now.

PR #411, opened 2026-08-24
feat: add local Samsung TV (QN90B) control component

danalec:
Adds a small companion component, `samsung_tv_local`, to control Samsung
QN90B TVs entirely over the local network — no SmartThings, no cloud.

The QN90B doesn't speak the CoAP/DTLS protocol localthings uses for
appliances (no response on 3010/3011/5683/5684), so it talks the TV's own
remote WebSocket (wss://:8002) plus Wake-on-LAN instead.

It provides:
- media_player: power (WOL on, KEY_POWER off), volume/mute, source, app launch, media keys
- remote entity: any KEY_* command
- buttons: one-tap launch of Netflix/YouTube/Prime/Disney+/Spotify

Pairing is the TV's Allow popup (token granted automatically, no PIN); the
client name is Base64-encoded so the popup renders correctly, and the token
refreshes itself if the TV ever rejects it.

Tested live against two QN90B units (on/off, volume, mute, WOL wake-up).

It's a sibling component rather than a device type inside localthings, since
the transport is different — happy to reconsider the layout.

### DIFF
--- a/custom_components/samsung_tv_local/README.md
+++ b/custom_components/samsung_tv_local/README.md
@@ -1,0 +1,116 @@
+# Samsung TV Local (QN90B)
+
+A sibling component to LocalThings for the QN90B TVs, **100% local** — it
+talks to the TV directly over the remote-control WebSocket (port 8002, TLS)
+plus **Wake-on-LAN** — no SmartThings, no cloud, no fees.
+
+> LocalThings (mbillow) controls Samsung appliances over **CoAP/DTLS**
+> (fridges, washers, ...). The **QN90B does not speak CoAP/DTLS** (verified
+> against the TV), so this component uses the TV's own WebSocket transport
+> (the same protocol family as the stock HA `samsungtv` integration, but as
+> part of this fork).
+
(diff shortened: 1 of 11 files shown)


vmonkey:
Are you sure it does not speak ocf/coap? Checking 3010/3011/5683/5684 does not tell you much: the device most probably listens on a single port, but that port is assigned dynamically and can be anywhere in the range. Multicast discovery is what finds it. Can you try the multicast discovery in https://github.com/QuiteYellow/SmartThings-Local/pull/47 with the TV on (not standby)? Better run it from a host wired to the same subnet, multicast is often dropped over Wi-Fi.

I am definitely not the one to decide for the PR, but why not use a dedicated integration for that protocol? There are many, e.g.,
- https://www.home-assistant.io/integrations/samsungtv/
- https://github.com/ollo69/ha-samsungtv-smart

The first one already does the same WebSocket, WOL power on, volume, source, app selection and the KEY_* commands. What is missing there for your QN90B?

Issue #396, opened 2026-08-18
OBSERVE subscribe burst is unpaced and can wedge an appliance at connect


QuiteYellow:
Reporting this from the library side, with hardware evidence, because I had the same bug and it took a wedged dryer to find it.

### What happens

`subscribe_hrefs()` in `observe.py:239-246` registers every href back to back with no delay:

```python
for href in hrefs:
    segs = [s for s in href.strip("/").split("/") if s]
    try:
        session.subscribe(segs)
```

`DtlsCoapSession.subscribe()` does not pace internally. It goes straight to the datagram send, so nothing throttles that loop, and it runs at whatever speed the interpreter manages. The session's configured limit defaults to 5 requests per second and has no effect here.

The docstring is right that each call is fire-and-forget, but that is what makes the burst dense rather than what makes it safe. The device still has to process every registration and answer each with an initial 2.05.

### Evidence

My own bridge had the same bug with a hardcoded 50ms sleep, which is 20 rps against a firmware ceiling I measured at ~14 rps for that appliance. After a container recreate:

```
19:30:01  dryer  DTLS connected — subscribing 11 paths
19:30:03  dryer  seeded → 25 links; sensors live
19:30:05  dryer  sweep complete (25 links, 928ms)
19:30:10  dryer  poll /operational/state/vs/0 timeout    <- and continuously from here
19:31:18  dryer  liveness: no successful poll in the liveness window
19:32:51  dryer  forced reconnect, re-seeded fine
```

The session is healthy for about five seconds, then the appliance stops answering anything until something forces a new session. An oven did the same on a recreate on 2026-08-07.

Your loop has no delay at all, so it is denser than the one that produced that.

### Why it may matter to you specifically

This burst runs at connect, which is exactly when the appliances in QuiteYellow/SmartThings-Local#37 fail. That reporter's fridge comes up, then will not poll, and recovers only after a forced reconnect or a long wait. I am not claiming that is the same bug, since I am still waiting on a log from them, but the shape matches and it is cheap to rule in or out.

### Fix

The real fix belongs in the library, and most of it already exists. QuiteYellow/SmartThings-Local#51, from @Jason-Morcos, paces the block GET, `post()`, and `delete()` inside the session rather than leaving it to callers. That is the same conclusion you reached from the write side in #384, arrived at independently.

`subscribe()` is the one request path it does not yet cover, which is why your loop and mine both slipped through. I am tracking that in QuiteYellow/SmartThings-Local#53.

[...]

Issue #294, opened 2026-08-05
async_send_command swallows "socket closed" without retry, requires manual reload


mchiriciuc:
Appliance model: AR12TXFYAWKNEU (air conditioner)

What happens: Occasionally a command (e.g. power off from the climate card) does nothing. HA UI shows no error to the user, but the log contains:

ERROR (MainThread) [custom_components.localthings.coordinator.192.168.151.33] command failed for /power/vs/0: socket closed

Polling entities keep updating normally (so the coordinator's own reconnect logic in _async_update_data is doing its job), but no subsequent command works until I manually Reload the integration for that device. After reload, everything works again — until the next silent socket close.

Root cause (from reading the code): async_send_command around L1213 executes _do_put in an executor, catches any exception, logs it, and returns. There's no session reconnect / retry, so a POST landing on a session that Samsung's firmware closed between polls is permanently lost from HA's perspective. The poll path handles this exact scenario cleanly (close → pause → reconnect → retry); the command path doesn't.

Suggested fix: on a "socket closed" / connection-level exception in _do_put, close and reconnect the session, then retry the POST once (mirroring _async_update_data's pattern). Bonus: raise a HomeAssistantError on final failure so the user sees a red toast instead of silently losing the command.

Diagnostics attached.

[config_entry-localthings-01KYS2A70ZTSPHZZRFMFRJ199J.json](https://github.com/user-attachments/files/30729388/config_entry-localthings-01KYS2A70ZTSPHZZRFMFRJ199J.json)

mbillow:
Fixed in #306 — `async_send_command` now mirrors the poll path's reconnect-and-retry-once behavior on a dead session, and raises a `HomeAssistantError` on final failure instead of silently dropping the command. Closing, will ship in the next release.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #415, opened 2026-08-29
Bump smartthings-local to 0.1.11 and unbreak CI on HA 2026.9

mbillow:
Three commits: the `smartthings-local` floor moves `0.1.8 → 0.1.11`, the suite stops failing on Home Assistant 2026.9's device-registry deprecations, and the integration stops making the deprecated calls at all. The second and third are not caused by the first — `main` went red on its own overnight — but they have to land for this branch to be green, so they ride along here.

## 1. smartthings-local 0.1.11

Raises the floor in `manifest.json`, `requirements-dev.txt` and the `Dockerfile`, picking up 0.1.9, 0.1.10 and 0.1.11.

- **0.1.9** — CoAP request sends now go through the session rate limiter. `subscribe()` previously bypassed it entirely, so an eleven-href OBSERVE registration burst went out as fast as the loop ran and could wedge an appliance until a watchdog forced a new session (upstream #53, and #396 here). `post()` and the first block of a blockwise GET are paced now too.
- **0.1.10** — a MID-indexed exchange registry so token-less control messages (empty ACK, bare RST) correlate, and retransmissions reuse one Message ID per RFC 7252 §4.2. Adds `SessionResetError` and `SessionIdentifierError`. Also adds `discover_ocf_responder_ports()`, which we do not call.
- **0.1.11** — DELETE, Block1 POST uploads, request-option validation, `discover_ocf_secure_ports()` and `read_plaintext_ocf_resource()`. `post()` gained opt-in write retransmission.

**Why no integration changes.** I diffed the 0.1.8 wheel against the 0.1.11 sdist rather than reading the release notes: only `errors.py`, `coap.py`, `dtls_handshake.py` and `dtls_session.py` changed, and the three new modules are unused here. Every call site keeps its signature — `connect`/`get`/`post`/`subscribe`/`ping`/`refresh_observes`, plus `dtls_probe.probe` and `diagnose_dtls_handshake`; the additions are keyword-only with defaults. `write_max_attempts` defaults to `1`, so write retransmission stays off and `_connect_session` passes no override. `SessionResetError` and `SessionIdentifierError` both subclass `SessionError` → `ConnectionError`, so existing handling covers them.

**Two behaviors worth a reviewer's attention.** Paced OBSERVE registration runs under the coordinator's session lock — the hold issue #294 deliberately kept short — so that hold is now about one limiter interval per href (~2.2s for eleven at 5 req/s) rather than eleven back-to-back datagrams, and a command write can queue behind a resubscribe burst for a couple of seconds. Still bounded, still far shorter than the grace wait it was split from, so the design is unchanged and the cost is recorded in the `subscribe_hrefs` docstring, whose old claim ("each is a fire-and-forget UDP datagram") understated it. Expect connect-to-seeded to take a couple of seconds longer per device; upstream considers that the fix. Separately, option validation now raises `ValueError`/`TypeError` on an empty path segment instead of sending a zero-length Uri-Path — `_href_to_path_segs` filters those, and the two poll loops that don't already sit inside a per-href `except Exception`.

## 2. Home Assistant 2026.9 device-registry deprecations

The nightly `Validate` run on `main` went red on 2026-08-28 with no commit behind it ([run 33156264353](https://github.com/mbillow/localthings/actions/runs/33156264353), same head as this branch's base). `pytest-homeassistant-custom-component>=0.13.316` is a floor, so CI on Python 3.14 resolved 0.13.359 / `homeassistant 2026.9.0b1`, which deprecates `DeviceInfo`'s `via_device` (for `via_device_id`) and `device_registry.async_get_device` (for `async_get_device_by_identifier`), both breaking in 2027.8.

**The three CI failures were all in test code.** `report_usage` only errors when it cannot find an integration frame — exactly a test calling the registry directly; the integration's own calls are attributed to it and merely log. Fixed version-neutrally, since the two interpreters resolve different cores:

- Two assertions used `async_get_device(identifiers=...)` to say nothing is left keyed on an old identifier. They now scan the entry's own rows via `async_entries_for_config_entry`.
- One fixture linked a subdevice row with `via_device=`; it goes through a `link_to_parent` helper that passes whichever kwarg `async_get_or_create` accepts.
- One assertion narrows `async_get`'s result past 2026.9's new `ChildDeviceEntry`.

## 3. Dropping the deprecated calls (runtime feature detection)

That left the integration still calling both deprecated APIs — harmless (a WARNING naming 2027.8) but noisy: a user with a composite appliance would see one report per platform on every setup, and anyone re-keying would see another. A new `devices.py` removes them without touching `hacs.json`'s 2025.1 floor, following the same feature-detection pattern as `__init__.py`'s `UnitOfDensity` and `new_unit_class`:

- **`find_entry_device` needs no detection.** `async_get_device` searched every config entry — which is precisely why identifiers stopped being unique across them and why it is deprecated. Both callers (the re-key collision check, and the debug-write panel finding its own device) mean their own entry's row, so scanning `async_entries_for_config_entry` is both the compatible spelling and the more correct one. It also closes a real edge in `rekey_entry`, which could previously match another entry's device and move this entry's entities onto it before deleting the row they came from.
- **`set_via_device` does detect**, because there is no common spelling: a core with `via_device_id` in `DeviceInfo` needs the master's registry id, an older one takes only the identifier. Where the id is wanted but the master row does not exist yet, the link is left unset rather than guessed — HA *drops* an entity whose `via_device_id` names no registered device, while an unresolvable `via_device` on an older core silently leaves the row unlinked, so this matches the old behavior in the one case the two can differ, and the next entity added under that device picks the link up.

## Testing

Both interpreters resolve different cores, so everything was run twice.

| | full suite | ruff format | ruff lint | ty |
|---|---|---|---|---|
| Python 3.14.7, homeassistant 2026.9.0b1 (what CI runs) | 1730 passed | pass | pass | pass |
| Python 3.13, homeassistant 2026.2.3 | 1730 passed | pass | pass | pass |

Beyond the suite, I drove a composite appliance through a full config-entry setup on both cores and read the device rows and log records back. On each, the subdevice row comes out linked to the master (`via_device_id` set), with **0 deprecation reports and 0 errors**; the re-key path reports nothing on either. That probe is how the "all three failures were test-only" claim above was established too — before this change it printed four `via_device` reports attributed to our own `async_add_entities` frames, and no errors.

## Follow-ups, deliberately not here

- Whether to pin `pytest-homeassistant-custom-component` rather than floor it. Pinning would have kept `main` green through this, at the cost of not hearing about core deprecations until someone bumps it — worth a decision either way, but not one this PR should make quietly.
- `discover_ocf_secure_ports()` / `discover_ocf_responder_ports()` for the config flow's port probing, and `write_max_attempts`, which needs per-device numbers to justify enabling.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,4 @@ FROM ghcr.io/home-assistant/home-assistant:stable
 # repeats the install attempt on every container recreate. Baking
 # smartthings-local into the image keeps the dev container usable
 # offline and avoids relying on that runtime install path.
-RUN pip3 install --no-cache-dir "smartthings-local>=0.1.8"
+RUN pip3 install --no-cache-dir "smartthings-local>=0.1.11"
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -21,7 +21,6 @@ import voluptuous as vol
 from homeassistant import config_entries
 from homeassistant.config_entries import ConfigFlowResult
 from homeassistant.core import callback
-from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers.selector import (
     NumberSelector,
     NumberSelectorConfig,
@@ -65,6 +64,7 @@ from .const import (
     PROBE_PORT_RANGE,
     SERVICE_WRITE_RESOURCE,
 )
+from .devices import find_entry_device
 from .learned import persist as learned_persist
 from .learned import stored as learned_stored
 from .registry.capabilities.laundry import cycle_options, personal_course_labels
@@ -1545,8 +1545,8 @@ class LocalThingsOptionsFlow(config_entries.OptionsFlow):
             # panel's href dropdown already lists actual hrefs off
             # coord.last_resources, and MAIN.to_actual is identity, so
             # this preserves the panel's existing behavior byte for byte.
-            dev = dr.async_get(self.hass).async_get_device(
-                identifiers=coord.device_info["identifiers"]
+            dev = find_entry_device(
+                self.hass, self.config_entry.entry_id, coord.device_info["identifiers"]
             )
             if dev is None:
                 return self.async_abort(reason="not_loaded")
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -49,6 +49,7 @@ from .const import (
     DTLS_LOCAL_PORT_BASE,
     SUMMARY_INTERVAL_S,
 )
+from .devices import set_via_device
 from .learned import LEARNABLE, LearnedModes, persist
 from .observe import GRACE_PERIOD_S, MODE_OBSERVE, MODE_POLL, ObserveManager
 from .registry import CAPABILITIES
@@ -786,13 +787,7 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
             model=model or None,
             serial_number=serial,
         )
-        # Set through a plain mapping because HA 2026.9 dropped `via_device`
-        # from the DeviceInfo TypedDict for `via_device_id` (still accepted
-        # until 2027.8), while cores back to this integration's 2025.1 floor
-        # take only `via_device`. The identifier is also the safer half: HA
-        # drops an entity whose `via_device_id` names no registered device,
-        # where an unresolved `via_device` just leaves the row unlinked.
-        cast("dict[str, Any]", info)["via_device"] = (DOMAIN, self.device_key)
+        set_via_device(self.hass, self._entry.entry_id, info, (DOMAIN, self.device_key))
         return info
 
     @property
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -779,14 +779,21 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
                 if subdevice.kind == "indexed"
                 else "Secondary Subdevice"
             )
-        return DeviceInfo(
+        info = DeviceInfo(
             identifiers={(DOMAIN, f"{self.device_key}_{subdevice.key}")},
-            via_device=(DOMAIN, self.device_key),
             name=f"{base_name} {label}",
             manufacturer=self.device_info.get("manufacturer") or "Samsung",
             model=model or None,
             serial_number=serial,
         )
+        # Set through a plain mapping because HA 2026.9 dropped `via_device`
+        # from the DeviceInfo TypedDict for `via_device_id` (still accepted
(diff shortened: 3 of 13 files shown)


mbillow:
CI is red here, but not because of this PR — `main` is red the same way at the same commit.

**Failing:** `Lint, format, and type check` (4 `ty` diagnostics) and `Pytest` (3 failures).

**Why it isn't this PR's:** the scheduled `Validate` run on `main` from 2026-08-28 ([run 33156264353](https://github.com/mbillow/localthings/actions/runs/33156264353)) fails with the identical diagnostics and the identical three test names, at head `b5e25d7` — the same commit this branch is based on, a day before this branch existed. The run the day before that was green. This diff touches `manifest.json`, `requirements-dev.txt`, the `Dockerfile` and one docstring in `observe.py`; it touches nothing in `coordinator.py`'s `DeviceInfo`, `rekey.py`, `config_flow.py`, or any of the failing tests.

**Root cause:** `requirements-dev.txt` pins `pytest-homeassistant-custom-component>=0.13.316` as a floor, so CI (Python 3.14) resolved `0.13.359` / `homeassistant-2026.9.0b1`, which turns two device-registry deprecations into hard errors:

- `DeviceInfo`'s `via_device` key → `via_device_id`
- `device_registry.async_get_device` → `async_get_device_by_identifier` / `async_get_device_by_connection` / `async_get_devices`, because identifiers are no longer unique across config entries

**No fix exists to port** — no open PR or issue in this repo covers it, so there is nothing to pull into this branch. I'm deliberately not fixing it here: it is unrelated to the library bump, it touches the migration and re-key paths, and widening a dependency-floor PR into an HA-compatibility change would make both harder to review and to revert.

**Proposed patch, for a separate PR:**

- `custom_components/localthings/coordinator.py:784` — `device_info_for` passes `via_device=(DOMAIN, self.device_key)`. Needs the master device's registry id for `via_device_id`, which is not an identifier tuple, so this becomes a registry lookup at the call site rather than a rename.
- `custom_components/localthings/rekey.py:74` — `dev_reg.async_get_device(identifiers=fresh)` → `async_get_device_by_identifier`, scoped to the config entry. Worth care: the replacement's uniqueness contract is what the deprecation is about, and this is the branch that decides whether a device gets re-keyed or removed-and-merged.
- `custom_components/localthings/config_flow.py:1548` — same call, looking up the coordinator's own device to target the `write_resource` service; the narrowest of the three.
- Tests: `tests/test_subdevice_discovery.py:152,250` assert `info["via_device"]`, and `tests/localthings/test_identity_migration.py:341` builds a fixture with `via_device=`; both follow whatever the production change settles on.

Happy to do that as its own PR and then bring the fix into this branch so this one can go green — say the word.

---

PR #421, opened 2026-08-31
Add OwnerPSK credential-backed sessions

### LINKED ISSUE #388: Would you want OCF media devices (TV, soundbar) in LocalThings?

atc722:
## Depends on #418

This is the immediate step-2 follow-up discussed in #388. It is intentionally a draft while #418 is open.

The branch contains #418's session-factory patch replayed onto current `main`. After #418 merges, I will rebase this branch and drop the duplicated parent commit before requesting review.

## Summary

- add an explicit, fail-closed certificate/OwnerPSK credential model
- migrate legacy v4 entries to explicit certificate mode without changing endpoint, credential, identity, or registry values
- construct OwnerPSK sessions through `PskAuth` and bind the authenticated peer to the configured `/oic/d` device UUID before publishing the session
- add standard Home Assistant reauthentication plus a user-initiated reconfigure path for OwnerPSK replacement
- keep credential material out of representations, diagnostics, errors, and resource dumps

## Why #418 comes first

#418 establishes one provider-based DTLS session-construction seam while preserving existing certificate behavior. This PR selects `CertificateAuth` or `PskAuth` at that seam, so authentication does not branch independently across initial setup, reconnect, and recovery paths.

## Authentication boundary

This consumes an OwnerPSK that the caller already possesses. It does not discover, extract, derive, mint, provision, rotate, transfer ownership, perform OTM, reset a device, or write OCF security resources.

No OwnerPSK onboarding form is added here. This draft establishes the config-entry, runtime, identity-binding, and recovery contract for the concrete device-support work that follows.

Current SmartThings-Local releases classify both a rejected PSK handshake and some endpoint/backend failures as generic `SessionError`. This PR does not guess that every such error is bad credentials. Automatic reauthentication is triggered only for definite invalid configuration, authenticated device mismatch, or an explicit `AuthenticationError`; manual reconfigure remains available for credential replacement. A package-side error-classification improvement can tighten the automatic path later without changing the stored model.

## Follow-up

After #418 and this credential/runtime layer are reviewed, the next changes are:

1. pluggable `/device/0` versus `/oic/res` resource enumeration for #388;
2. a separate, sanitized WD86 capability/entity/status/control PR using this OwnerPSK path; and
3. endpoint rediscovery as its own bounded change.

The WD86 work needs this PR first because its authenticated read and control sessions require an explicit OwnerPSK-backed config entry, exact device binding, and a safe credential-replacement path. Keeping those generic concerns here avoids embedding authentication policy in one appliance profile.

## Integration note

#407 also currently reserves config-entry version 5. If it merges first, this draft will rebase and move this migration to version 6.

## Validation

- `python -m pytest -p no:cacheprovider tests/ -q`: 1,809 passed with SmartThings-Local 0.1.12
- focused credential/session/config-flow/coordinator/diagnostics/migration suite: 222 passed
- Ruff format and lint checks
- `ty check custom_components tests`
- translation JSON parse validation
- `git diff --check origin/main...HEAD`

Relates to #388. Depends on #418. This does not close #388.

### DIFF
--- a/custom_components/localthings/__init__.py
+++ b/custom_components/localthings/__init__.py
@@ -16,7 +16,16 @@ from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.typing import ConfigType
 
-from .const import CONF_DEVICE_TYPE, CONF_HOST, CONF_PORT, CONF_SERIAL, DOMAIN, PLATFORMS
+from .const import (
+    AUTH_CERTIFICATE,
+    CONF_AUTH_TYPE,
+    CONF_DEVICE_TYPE,
+    CONF_HOST,
+    CONF_PORT,
+    CONF_SERIAL,
+    DOMAIN,
+    PLATFORMS,
+)
 from .coordinator import LocalThingsCoordinator, snapshot_store
 from .registry.identity import resolve_serial
 from .rekey import rekey_entry
@@ -213,12 +222,16 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     what the coordinator rewrites *from* once a live poll finally hands it
     a UUID to rewrite *to*.
 
-    The version is still bumped now rather than at that point, because the
-    bump's real job is the `entry.version > 4` downgrade guard below: an
+    v4 -> v5 makes the authentication mode explicit. Every entry written by
+    an older release is certificate-backed, so this adds only the discriminator
+    and leaves all endpoint, credential, identity, and registry data untouched.
+
+    The v4 version is still bumped before a live UUID re-key completes because
+    that bump's real job is the downgrade guard below: an
     entry re-keyed onto a UUID and then loaded by a release that reads
     CONF_SERIAL as the key would silently orphan every entity it has.
     """
-    if entry.version > 4:
+    if entry.version > 5:
         return False  # downgrade: this release doesn't know the newer shape
 
     if entry.version == 1:
@@ -257,6 +270,14 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             entry.entry_id,
             legacy_key,
         )
+
+    if entry.version == 4:
+        hass.config_entries.async_update_entry(
+            entry,
+            data={**entry.data, CONF_AUTH_TYPE: AUTH_CERTIFICATE},
+            version=5,
+        )
+        _LOGGER.debug("migrated entry %s to version 5 (certificate auth)", entry.entry_id)
 
     return True
 
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -35,6 +35,7 @@ from homeassistant.helpers.selector import (
 )
 
 from . import cloudcourse
+from . import session as session_factory
 from .const import (
     CLIENTHELLO_PROBE_RETRIES,
     CLIENTHELLO_PROBE_TIMEOUT_S,
@@ -742,13 +743,16 @@ def _diagnose_failures(
 
 def _handshake_and_read(host: str, scan: _PortScan, cert_pem: str, key_pem: str) -> dict:
     """Handshake each candidate in turn, returning the first device that answers."""
-    from smartthings_local.protocol.dtls_session import DtlsCoapSession
-
     failures: list[tuple[int, Exception]] = []
     for port in scan.candidates:
         sess = None
         try:
-            sess = DtlsCoapSession(host, port, cert_pem=cert_pem, key_pem=key_pem)
+            sess = session_factory.create_certificate_session(
+                host,
+                port,
+                certificate_pem=cert_pem,
+                private_key_pem=key_pem,
+            )
             sess.connect()
             sess.start_reader()
             return _read_device(sess, host, port)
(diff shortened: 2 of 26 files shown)


vmonkey:
Super!

One idea from my side: the user is currently required to enter the owner UUID, but that value can be read without any authentication from `/oic/sec/doxm` (`devowneruuid`). I’m not sure how the result would differ from what @Jason-Morcos seemed to use, but what about at least pre-filling that field while still allowing the user to change the value? That would cover the common case (one field instead of two) while keeping the override for devices that don’t expose doxm (not sure if this happens), or where the user wants to bind a different identity.

EDIT: version bump in #407 will be removed

Jason-Morcos:
That is the value I used, but I treated it as a candidate rather than proof of the credential binding.

In my working path, `devowneruuid` selects the retained owner credential. I then require owned and internally consistent DOXM/PSTAT state, a successful OwnerPSK handshake with that UUID/key, an authenticated read from the expected device, and a fresh ownership-state check afterward. An editable UUID is useful when replacing an imported credential pair, but it is not really a way to choose a different identity independently of the key; a mismatched pair should fail authentication.

For this PR, the entry already has `owner_uuid` because this only adds reauth/reconfigure for an existing OwnerPSK entry. I think the smallest useful change is to suggest that stored UUID in the form, leave it editable, and continue never prefilling the PSK.

I would leave a DOXM lookup for the later onboarding/import flow. It is useful there as a best-effort candidate, but it cannot be required across device families: some firmware returns 4.01 to plaintext `/oic/sec/doxm` and only exposes it over an anonymous server-authenticated session. Either way, the candidate should not be persisted or treated as trusted until the PSK handshake succeeds and authenticated `/oic/d.di` matches the configured device.

atc722:
Thanks @vmonkey and @Jason-Morcos — I went with the smaller change discussed here. In `4b8fa64`, the reauth/reconfigure form now shows the UUID already stored on the entry as an editable suggestion, while leaving the PSK field empty.

I’ve left the DOXM lookup for the later onboarding/import work, where it can be used as a hint and then checked through the PSK handshake and authenticated `/oic/d.di`.

mbillow:
Same note as on #418 — I've opened #435 to settle the PSK / server-authenticated architecture in one place, and I'd like a full plan agreed there before merging anything in this area.

Two of the open questions bear directly on this PR. First, whether the credential-acquisition boundary is permanent: having read `ocf-pki-laundry.md` properly, §4 is an owner-authorized SmartThings transition and §6 writes OCF security resources, so I think the boundary you drew here is the correct one — but I'd like that confirmed rather than assumed, because it decides what the onboarding flow can ever look like. Second, how device identity gets bound, which interacts with what `CONF_DEVICE_KEY` actually holds (it's the first *usable* value from `di` → `pi` → `serialNum` → host, and the `pi` case in particular compares a platform UUID against `/oic/d`'s `di`).

You mentioned planning to rebase this onto a merged #418 and drop the duplicated parent commit — please hold off on that. I don't want you spending the effort while the sequencing is still open.

Worth saying that a lot here is right regardless of how the plan lands: keeping credential material out of diagnostics, the PASSWORD selector and never prefilling the PSK, storing only after a successful authenticated round-trip, and declining to treat a generic `SessionError` as bad credentials. Those hold up whatever shape the surrounding structure ends up taking.

---

PR #429, opened 2026-09-01
Add refrigerator proximity and night-light controls

Jason-Morcos:
## What this changes

Adds two representation-backed refrigerator controls:

- welcome-lighting proximity selection from the advertised `supportedLevels`, plus a default-disabled diagnostic sensor for `desiredSenseLevel`
- the missing `light.control.status` switch for the enhanced night-light schedule

The existing welcome-lighting switch, day/night brightness, schedule times, filter sensors, auto-door, temperature, ice-maker, and refrigeration entities stay unchanged.

On my refrigerator, `currentLevel` is the configured proximity sensitivity and `desiredSenseLevel` is the firmware's current sense bucket. The latter freezes when welcome lighting is disabled, so it is deliberately diagnostic rather than presented as motion detection. Proximity and night-light changes were round-tripped and restored on the appliance.

Water-filter reset is intentionally not included. A live test on a second refrigerator accepted `filterUsage: "0"` with CoAP 2.04 but retained the prior value on exact readback, so `filterResetType` alone does not establish a working local reset command.

## Validation

- existing sanitized refrigerator fixtures exercise both representations
- advertised-option, invalid-option, and missing-field gating
- exact write-payload coverage and switch-off regression coverage
- regression coverage that the unverified water-filter reset remains absent
- full test suite: 1,839 passed
- Ruff format/lint and ty

### DIFF
--- a/custom_components/localthings/registry/capabilities/fridge.py
+++ b/custom_components/localthings/registry/capabilities/fridge.py
@@ -16,6 +16,7 @@ segments.
 
 import datetime
 
+from ..batch import is_stub_rep
 from ..capability import Capability
 from ..entities import (
     BinarySensorDesc,
@@ -574,6 +575,63 @@ AUTOFILL = Capability(
     ),
 )
 
+
+_PROXIMITY_LEVELS = {
+    "0": "nearest",
+    "1": "near",
+    "2": "middle",
+    "3": "far",
+}
+_PROXIMITY_RAW_BY_LEVEL = {level: raw for raw, level in _PROXIMITY_LEVELS.items()}
+
+
+def _proximity_level(value):
+    return _PROXIMITY_LEVELS.get(str(value))
+
+
+def _proximity_options(resources):
+    rep = resources.get("/proximity/vs/0") or {}
+    supported = rep.get("supportedLevels")
+    if not isinstance(supported, (list, tuple)):
+        return []
+    return [str(value) for value in supported if _proximity_level(value) is not None]
+
+
+def _proximity_display(value, _resources):
+    return _proximity_level(value)
+
+
+def _proximity_write(value, rep, href=None):
+    value = str(value)
+    raw = value if value in _PROXIMITY_LEVELS else _PROXIMITY_RAW_BY_LEVEL.get(value.casefold())
+    supported_levels = rep.get("supportedLevels")
+    if not isinstance(supported_levels, (list, tuple)):
+        return None
+    supported = {str(item) for item in supported_levels}
+    if raw is None or raw not in supported:
+        return None
+    return ["proximity", "vs", "0"], {"currentLevel": raw}
+
+
+def _proximity_setting_exists(rep, resources):
+    return is_stub_rep(rep) or (
+        _proximity_level(rep.get("currentLevel")) is not None
+        and bool(_proximity_options(resources))
+    )
+
+
+def _proximity_sense_exists(rep, _resources):
+    return is_stub_rep(rep) or _proximity_level(rep.get("desiredSenseLevel")) is not None
+
+
+def _proximity_sense_options(resources):
+    rep = resources.get("/proximity/vs/0") or {}
+    supported = rep.get("supportedSenseLevels")
+    if not isinstance(supported, (list, tuple)):
+        return []
+    return [level for value in supported if (level := _proximity_level(value)) is not None]
+
+
 WELCOME_LIGHTING = Capability(
     href="/proximity/vs/0",
     poll_tier="warm",
@@ -588,6 +646,30 @@ WELCOME_LIGHTING = Capability(
                 ["proximity", "vs", "0"],
                 {"status": "On" if p == "On" else "Off"},
             ),
+        ),
+        SelectDesc(
+            key="welcome_lighting_proximity",
+            field="currentLevel",
+            icon="mdi:motion-sensor",
+            entity_category="config",
+            options=_proximity_options,
+            display_fn=_proximity_display,
+            exists_fn=_proximity_setting_exists,
+            value_fn=_proximity_level,
+            write_fn=_proximity_write,
+        ),
+        # desiredSenseLevel freezes while welcome lighting is off, so it is
+        # firmware diagnostics rather than a live motion-detection entity.
+        SensorDesc(
+            key="welcome_lighting_sense_level",
+            field="desiredSenseLevel",
+            icon="mdi:motion-sensor",
+            entity_category="diagnostic",
+            enabled_default=False,
+            device_class="enum",
+            options=_proximity_sense_options,
+            exists_fn=_proximity_sense_exists,
+            value_fn=_proximity_level,
         ),
     ),
 )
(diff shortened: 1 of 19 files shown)


custom_components/localthings/registry/capabilities/fridge.py
```diff
@@ -266,6 +270,41 @@ def _icemaker_write(field):
...
+        and any(
+            str(reset_type).casefold() in {"replaceable", "washable"} for reset_type in reset_types
+        )
+    )
+
+
+def _water_filter_reset_write(_payload, rep, href=None):
+    if not _water_filter_reset_supported(rep, {}):
+        return None
+    return ["filter", "waterfilter", "vs", "0"], {"x.com.samsung.da.filterUsage": "0"}
```
mbillow: This doesn’t seem to work on my fridge. Have you tested this write on a live device?

```yaml

action: localthings.write_resource
data:
  hold_session_lock: true
  device_id: <redacted>
  writes:
    - href: /filter/waterfilter/vs/0
      payload:
        x.com.samsung.da.filterUsage: '0'
  verify_after: 1
```

```yaml
device_id: <redacted>
results:
  - href: /filter/waterfilter/vs/0
    actual_href: /filter/waterfilter/vs/0
    code: '2.04'
    raw_code: 68
    accepted: true
    before:
      x.com.samsung.da.filterUsage: '100'
      x.com.samsung.da.filterUsageResolution: '1'
      x.com.samsung.da.filterResetType:
        - replaceable
      x.com.samsung.da.filterStatus: replace
      href: /filter/waterfilter/vs/0
    after:
      rt:
        - x.com.samsung.da.filter.water
      if:
        - oic.if.baseline
        - oic.if.a
      x.com.samsung.da.filterUsage: '100'
      x.com.samsung.da.filterUsageResolution: '1'
      x.com.samsung.da.filterResetType:
        - replaceable
      x.com.samsung.da.filterStatus: replace
    changed: false
verified:
  /filter/waterfilter/vs/0:
    code: '2.05'
    raw_code: 69
    rep:
      rt:
        - x.com.samsung.da.filter.water
      if:
        - oic.if.baseline
        - oic.if.a
      x.com.samsung.da.filterUsage: '100'
      x.com.samsung.da.filterUsageResolution: '1'
      x.com.samsung.da.filterResetType:
        - replaceable
      x.com.samsung.da.filterStatus: replace
    held: false
    read_error: null
```

Jason-Morcos: Thanks for testing this. I only tested a same-value filterUsage write on my refrigerator; it was accepted and read back unchanged. I deliberately didn’t send the destructive zero value against the in-use filter (I was planning on testing that part when I actually swap the filter out).
That established that the field accepted a write, but it wasn’t enough to prove the reset behavior. Your exact zero test shows the proposed payload is ignored on at least your firmware despite the 2.04 response, so the reset button shouldn’t ship on that evidence. I’ll remove the filter-reset portion from this PR and keep the proximity and night-light controls, which I did round-trip on my refrigerator and restore afterward.

mbillow (commented):
This one's in good shape — nothing below is a correctness defect, but a few things are cheap to fix now and expensive later.

**Verified locally against `7ad65b0`:** 1834 → 1839 tests, `ty` clean, `ruff` failures pre-existing on main.

I round-tripped the select by hand through `select.py`: `_proximity_options` returns post-`value_fn` names, `_display_option` passes them through the catalog unchanged, `async_select_option` maps display → raw → `_proximity_write` → `{"currentLevel": "2"}`, and the optimistic apply caches the raw value. Consistent end to end — though note it inverts `entities.py`'s documented "returns raw device option values" contract for callable `options`, so that docstring and this are now slightly at odds. The `is_stub_rep(rep) or …` idiom matches `common.py` exactly, goldens are updated for all five affected fixtures, and I confirmed `/cabinet/light/total/vs/0` carries `light.dimming.status` while `/cabinet/light/enhanced/vs/0` carries `light.control.status` — so the new switch genuinely isn't a duplicate of `cabinet_light_switch`.

---

## The device advertises the sense-level list, and we're ignoring it

`refrigerator_device.json`'s `/proximity/vs/0` includes `"supportedSenseLevels": ["1", "3"]`, right next to `supportedLevels`. The select correctly reads `supportedLevels`; the `current_proximity_level` sensor hardcodes `tuple(_PROXIMITY_LEVELS.values())` (`fridge.py:653`) — all four — for a field the device says only ever takes two of.

Not a bug: `sensor.py:76-81` already appends unexpected enum values. But it's the one spot in the PR doing the thing we generally avoid, which is hardcoding a list the device is willing to tell us. Narrowing it properly needs `SensorDesc.options` to accept a callable, which is a real change — so a comment explaining the choice is a fine resolution if you'd rather not go there.

## The two names read backwards from the fields they bind

- `welcome_lighting_proximity` (`fridge.py:637`) ← `currentLevel`
- `current_proximity_level` (`fridge.py:647`) ← `desiredSenseLevel`

Anyone reading the entity list will map "current proximity level" to `currentLevel` and be wrong. Something like `welcome_lighting_sense_level` would track the field it actually reads.

Same question for `night_light` (`fridge.py:754`), whose key reads like a light — next to the existing `cabinet_light_switch` — while its display name is "Night lighting schedule". Worth settling before merge; entity_ids break people's automations once released.

## The reasoning is all in the PR body, none of it in the code

`CONTRIBUTING.md` asks for one or two sentences of "why" with the confirming observation, and the most valuable one here is missing: `desiredSenseLevel` freezes when welcome lighting is off, which is *why* it's a disabled-by-default diagnostic rather than motion detection. That's exactly the non-obvious fact a future reader needs, and right now it lives only on this page.

## `test_unverified_water_filter_reset_is_not_exposed` is tautological

`filter_reset` and `filterResetType` appear nowhere in `custom_components/`, so the test asserts the absence of a key that was never added — it can't fail in a way that means anything, and it carries no comment explaining the negative result it's guarding.

The finding itself is genuinely useful and worth keeping: 2.04 accepted, prior value retained on exact readback. `CONTRIBUTING.md` is explicit that failed-attempt logs go to an issue or docs, and `docs/investigations/ac-filter-reset.md` is the precedent for exactly this shape. I'd move it there and drop the test.

## Nits

- `TestSwitchOffIsNotInverted.CASES` still indexes `CABINET_LIGHT_ENHANCED.entities[0]`, which now silently means the newly inserted switch. Nothing else in the repo indexes these positionally and your own new tests use key lookup — do the same here.
- `_proximity_write` rebuilds the `reverse` dict on every call (`fridge.py:599`); it can be module-level.
- Unrelated blank-line deletion above the "Internal deodorizing filter" comment in `fridge.py`.

## Unrelated to the diff

GitHub reports this PR as `mergeable_state: unstable` with zero check runs — CI doesn't appear to have reported at all. Worth a re-trigger.

---

Jason-Morcos:
Thanks — I pushed 94c1c36 with these cleaned up.

`SensorDesc.options` can now be derived from the live resource snapshot, and the diagnostic uses the advertised `supportedSenseLevels` list while still admitting an unexpected reported value safely. I renamed the two new keys to `welcome_lighting_sense_level` and `night_lighting_schedule`, added the note explaining why the frozen sense field is diagnostics rather than motion detection, and changed the proximity select callback to return raw device values so it matches the documented callable-options contract.

The filter-reset result now lives in `docs/investigations/refrigerator-water-filter-reset.md` and the tautological test is gone. I also replaced the positional switch lookup, moved the reverse proximity map to module scope, and restored the unrelated blank line.

The updated head passes all 1,839 tests, Ruff format/lint, and ty locally. The exact-head validation run also passed pytest, lint/format/type, and Hassfest: https://github.com/Moballo-LLC/localthings/actions/runs/33601320558. Its HACS job only flags fork repository metadata (topics and Issues); the upstream PR workflow still needs approval before it can run.

Issue #433, opened 2026-09-01
Add ME80H2160RAA Support


its-nolan:
### Appliance model

ME80H2160RAA

### What's missing or wrong

New model microwave, most options are not available via Local things.

### Diagnostics file

[config_entry-localthings-01M1FFH8NY15DV2G0T6JJH1TMT-1.json](https://github.com/user-attachments/files/31715111/config_entry-localthings-01M1FFH8NY15DV2G0T6JJH1TMT-1.json)

mbillow:
Opened #434 with local support for this board (a DAWIT 3.0 generation combi that reports cavity/hood state through a completely different set of resources than every other microwave dump on file). Every write is an educated guess based on your dump's own reported ranges/lists — none of it has been confirmed against a live device. @its-nolan, if you're able to try a few things via **Settings → Devices → [the microwave] → Configure → Debug write panel** (or the `localthings.write_resource` service), that would help a lot:

⚠️ **Put a cup of water in the microwave before touching anything under #1–2** — this is a real device, not a simulator.

**1. Cavity controls** (`/oven/status/vs/0`):
```json
{"mode": {"name": "MicroWave", "indexRecipe": "00000000000000"}}
{"microwavePowerLevel": {"unit": "percentage", "setting": 50}}
{"time": {"setting": 30}}
{"childLock": "on"}    // then "off" to restore
```
For power level, pick one of the values the `power_level` select actually offers in HA (read live from the device's own spec) rather than assuming 50 is valid.

**2. Can this generation start a cycle locally?** No older Samsung board in this integration has managed it. One combined write, with `verify_after: 5`:
```json
{"mode": {"name": "MicroWave", "indexRecipe": "00000000000000"},
 "microwavePowerLevel": {"unit": "percentage", "setting": 50},
 "time": {"setting": 10},
 "operation": "run"}
```
If `operation` actually flips and holds instead of reverting back, that's worth knowing — nothing in the integration currently exposes a start button because it's never worked before on this family.

**3. Settings** (`/oven/settings/status/vs/0`) — restore each after:
```json
{"beepSound": "off"}
{"remindBeep": "off"}
{"displayTimeAutoSync": "off"}
```

**4. Hood** (`/hood/status/vs/0`):
```json
{"fanSpeed": "low"}    // then "off"
{"lamp": "medium"}
```

**5. `ddms`** — no idea what this is, couldn't find it documented anywhere. Exploratory only:
```json
[...]

Issue #181, opened 2026-07-29
TP1X_DA-KS-MICROWAVE-01051


dacmcbibs:
### Appliance model

ME21DG6399SRAA

### What's missing or wrong

A few differences between the Smart things app and this intergration:

- Smartthings can determine if the child lock is active or inactive - the integration cannot
- The integration has a child lock toggle - it doesn't work (it does not look like smart things has an equivalent feature)
- Smartthings app has a power level slider (0 - 100%) - the integration does not
- Smartthings app has the following modes { "Microwave", "popcorn", "Reheat", "Defrost", "Sensor Cook" , "Keep Warm"} the integration has {"Microwave", "Keep Warm" , "Auto Cook"}
- Smartthings has a toggle for "Filter Reminder" - Missing from integration
- Smart things has a toggle for "end signal reminder" - Missing from integration 
- Smartthings can send the cook configuration to the Microwave via a "send to microwave" button, I don't see a way to do this with the intrgration
- The light on the smart things app has 3 settings - Low High and off, the integration just has an off on toggle
- 

[config_entry-localthings-01KYKDZAYTWN8ZQ846V22P92MS.json](https://github.com/user-attachments/files/30485683/config_entry-localthings-01KYKDZAYTWN8ZQ846V22P92MS.json)

mbillow:
Thanks for the detailed comparison — this is exactly the kind of report that's easy to act on. Going through the list against your dump:

**Added** (pushed to a branch, will ship in the next release):
- Filter Reminder and End Signal Reminder toggles. Both `FilterRemind_*`/`RemindBeep_*` tokens are already in your dump's options array, and both On/Off values are already confirmed from an earlier dump of this same board (`TP1X_DA-KS-MICROWAVE-01051`/ME7500D), so this was safe to wire straight up the same way Sound/Lamp already work.

**Working as intended, not a bug:**
- The mode list (Microwave/Keep Warm/Auto Cook) is read live from your unit's own `supportedModes` field, not hardcoded — your device only reports those (plus `NoOperation`, which is the idle state). SmartThings' extra modes (Popcorn, Reheat, Defrost, Sensor Cook) are presets the cloud app offers that your unit's firmware doesn't actually expose locally, so there's nothing for the local integration to add without them showing up in that field.

**Need more data before I can safely add these** (don't want to guess a write and have it silently do nothing or the wrong thing on real hardware):
[...]

PR #434, opened 2026-09-02
Add ME80H2160RAA microwave/hood support (issue #433)

mbillow:
## What

Adds coverage for the DAWIT 3.0 generation `TP1X_DA-KS-MICROWAVE-0102X` combi board (model `OT80H30-/AA0`, issue #433). This board reports none of the microwave family's existing hrefs at all — no `/oven/vs/0`, `/mode/vs/0`, `/temperatures/vs/0`, `/doors/vs/0`, `/operational/state/vs/0`, `/hood/fanspeed/vs/0`. Instead:

- `/oven/status/vs/0` — one bare-field resource carrying mode, door, child lock, microwave power level, cook time, machine state
- `/oven/spec/vs/0` — per-mode bounds (time range, power-level list); bare coverage marker, read live by the power-level select
- `/oven/settings/status/vs/0` — beep / end-signal-reminder / display-time-sync toggles
- `/hood/status/vs/0` + `/hood/spec/vs/0` — the built-in vent hood, same split

All 5 previously-unbound hrefs are now bound, added alongside the existing hrefs so both board generations keep working through the same `microwave` registry. Also adds the `oic.d.microwave` → `microwave` row to `_OIC_TYPE_TO_KEY` — this is the first dump with a real `/oic/d` device type for the family (routing already worked via the `MICROWAVE` board token).

Entities added, all read-only: `machine_state`, `cycle_active`, `door_open`, `child_lock`, `cooking_mode`, `power_level`, `cook_time`, `cook_time_remaining`, `cook_finish_time`, `sub_operation` (cavity); `beep`, `remind_beep`, `display_time_auto_sync` (settings); `hood_fan_speed`, `hood_lamp`, `grease_filter_alarm`, `front_vent_open` (hood).

`cooking_mode`, `hood_fan_speed` and `hood_lamp` read their enum options live from the device's own data (`availableModeList`, the sibling spec resource's `fanSpeedList`/`lampStateList`) rather than a hardcoded list, so a different board generation's vocabulary still renders.

**Nothing in this generation is writable.** The branch originally shipped guessed write contracts; issue #433's reporter tested every one and got CoAP 4.05 (Method Not Allowed) from all five resources. The board declares them `oic.if.s` (sensor interface) where all 77 writable hrefs in the fixture corpus declare `oic.if.a`, and control from the SmartThings app still works — so the local API is read-only on this board, not the account. Every `write_fn` was dropped, the same call `common.KIDS_LOCK_VS_FALLBACK` got when #181/#183 turned up the identical 4.05. See the comment thread below for the full reasoning.

## Testing

- `pytest tests/ -q` — 1856 passed
- `ruff check` / `ruff format --check` — clean
- `ty check` — clean
- New fixture (`microwave_me80h2160raa`, scrubbed of serials/MACs), golden state-key regression, and capability unit tests covering the new write contracts and dynamic option lists

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01A8N9XFZSh3vDL99B9PMr2J)_


mbillow:
Went through this critically and reworked the substance. Pushed as `46109ef` (main merged in, so the conflict is gone). The structure was right — the capabilities live in the correct modules, both registries pick them up, the option lists read from the device, the fixture is a verbatim capture. **The write side was wrong, and hardware said so on this very issue.**

## The writes are disproven, not merely unconfirmed

This branch shipped eight controls whose contracts it described as "educated guesses." @its-nolan then tested all of them and reported *"Every command returned The device returned CoAP code 4.05."* Three independent lines say the same thing:

1. **4.05 is Method Not Allowed**, not a payload complaint. A wrong body shape gets 4.00; a wrong path gets 4.04. Getting 4.05 from all five resources is the board saying it doesn't accept writes there at all.
2. **The board declares them read-only.** A live read of `/oven/status/vs/0` on the same unit reports `if: ["oic.if.baseline", "oic.if.s"]` — the OCF *sensor* interface. I checked this against the whole fixture corpus: of the 78 hrefs this registry writes to, **77 declare `oic.if.a`** (actuator). Not one DAWIT 3.0 resource does.
3. **Cloud control still works.** The reporter can drive the unit from the SmartThings app. So the ceiling is this board's local API, not account permissions — which is the interesting finding here and worth recording.

We already have a precedent for exactly this: `common.KIDS_LOCK_VS_FALLBACK`. It was a `SwitchDesc` until #181/#183's reporter confirmed that writing even the *correct* value still 4.05'd, and the answer was to drop the write side and ship a `BinarySensorDesc`. Same call here — a control that always throws is worse than a sensor that reads, because it looks like a bug in Home Assistant rather than a limit of the appliance.

So: **no `write_fn` anywhere in this generation.** Switches → binary sensors, selects and the cook-time number → sensors. A test asserts the whole generation binds zero writable entities, with the evidence in its docstring, so this can't quietly regress.

## What that simplified

Removing the writes took several of the branch's more strained constructs with it:

- **The idle-sentinel union is gone.** `_status_mode_options` unioned `mode.name` into the option list so `NoOperation` wouldn't blank the select. `sensor.py`'s `options` property already admits a live value outside the declared list (PR #341) — so for a sensor, `availableModeList` passes straight through.
- **The `NoOperation` write gap is gone.** The branch listed `NoOperation` as an option but had `write_fn` return `None` for it — a silent no-op. #445's review specifically rejected that shape for the older generation's mode select (it now raises a translated error), so this would have re-introduced the pattern we just removed elsewhere.
- **`power_level` no longer needs the spec resource.** Reading `powerLevelList` only ever existed to bound a write. As a sensor it takes its unit from the device's own `microwavePowerLevel.unit` field, which is a device-nominated unit rather than a guessed one.
- **`COOK_TIME_MIN_S`/`MAX_S` are gone** along with the range validation, and with them the comment explaining why the floor was relaxed from modeSpec's 1 to 0.

## Two things I changed on the read side

- **`cook_time`/`cook_time_remaining` now declare `device_class="duration"`, unit seconds.** The branch left them unitless. The unit is pinned by arithmetic, not a guess: modeSpec caps `time` at **6039 = 99×60 + 99** — the 99:99 ceiling these panels count down from. Same kind of cross-check the dust sensors used for their CAI bands.
- **`child_lock` follows the repo's inverted lock polarity.** As a `BinarySensorDesc` with `device_class="lock"`, On means *unlocked*, matching `common.KIDS_LOCK_GENERIC`. The branch's switch had the opposite sense, which would have read backwards next to every other appliance.

`cooking_mode` maps the board's PascalCase names onto the state keys `select.cooking_mode` already ships, so `KeepWarm` reads "Keep warm" here exactly as it does on the older generation — no new label vocabulary invented.

## Also

Added the missing `seeds_note` to the fixture, recording that the `rt`/`if` arrays are absent from the diagnostics snapshot and that the separate live read is where `oic.if.s` comes from — otherwise the next person to look at this fixture can't see the evidence the whole design now rests on.

1985 passing, lint/format/type clean.

Issue #335, opened 2026-08-09
Samsung 2-in-1 AC sub-unit not discovered on LocalThings 0.20.0


whitesky1224:
### Appliance model

AF19R7573WSN

### What's missing or wrong

Hi mbillow,

I am reporting an issue with a Samsung 2-in-1 Air Conditioner (Model: AF19R7573WSN). LocalThings successfully discovers and registers the main stand unit, but the wall-mounted sub-unit is not discovered or created as an entity.

### Environment

* Home Assistant OS: 18.2
* Home Assistant Core: 2026.7.4
* LocalThings version: 0.20.0
* Model: ARTIK051_FAC_BORA_19K
* Device: AF19R7573WSN

### Details & Observations

* Main stand unit: Successfully discovered and registered
* Wall-mounted sub-unit: Not discovered / no entity created
* The main device exposes a non-empty `subdeviceIdList` through `/subdevices/vs/0`.
* However, the diagnostics show `subdevices: []`.
* `subdevices_skipped` is also empty.
* All entries under `subdevice_probes` return `false`.

For example:

`/<subdevice-id>/...`: `false`

I also independently verified through network diagnostics that both indoor units operate behind a single IP address. There does not appear to be a separate IP address assigned to the wall-mounted sub-unit.

### Related issue #248

I found issue #248, which appears to describe essentially the same problem with the same Samsung model family (`ARTIK051_FAC_BORA_19K`): the main unit is discovered, while the wall-mounted sub-unit is not, despite the sub-device ID being exposed through `/subdevices/vs/0`.

However, I am still experiencing the same behavior with the current LocalThings 0.20.0 release, even though the 2-in-1 sub-device discovery issue appears to have been addressed after the earlier reports.

I also reviewed issue #177 and the related implementation. My understanding is that Samsung 2-in-1 systems can expose multiple logical indoor units behind a single IP address and a single DTLS session, and LocalThings was updated to support multiple sub-device discovery patterns.

My device appears to use the single-IP architecture, but the sub-unit still fails during the probe stage.

### Diagnostics

The attached diagnostics JSON shows the following pattern:

* `subdeviceIdList`: present
* `subdevices`: `[]`
* `subdevices_skipped`: `[]`
* `subdevice_probes`: all `false`

I have attached the full diagnostics JSON for reference.

Thank you for your hard work on LocalThings.

### Diagnostics file

[...]

Issue #282, opened 2026-08-04
[Bug] Missing sub-device entities on ARTIK051 2-in-1 AC (All subdevice_probes return false)


sjdprp0809:
Appliance model:
ARTIK051_FAC_TIME_21K (TizenRT 1.0)
What's missing or wrong:
First of all, thank you so much for the recent patch (related to issue #265). Thanks to the update, the 5-minute timeout issue is completely gone, and the main Stand AC is now added successfully in just a few seconds!
However, the wall-mounted AC (sub-device) is still completely missing from the entity list.
I downloaded the diagnostic JSON and found the exact reason why the sub-device is not being registered. My AC uses the legacy ⁠ARTIK051⁠ module, and it seems to actively reject local probes for the sub-device:
1. The main device does expose the sub-device's UUID correctly via ⁠/subdevices/vs/0⁠.
2. But in the ⁠subdevice_probes⁠ section, every single request (e.g., ⁠/power/0⁠, ⁠/temperature/current/0⁠, ⁠/mode/vs/0⁠ under the sub-device UUID) returned ⁠false⁠.
3. As a result, the ⁠subdevices⁠ array is entirely empty (⁠[]⁠).
It looks like the older ARTIK051 firmware happily reports sub-device data to the SmartThings Cloud, but completely refuses to serve the detailed sub-device resources locally via OCF.
Is there any known workaround or alternative endpoint to fetch sub-device data for this legacy ARTIK051 firmware? Or is this a hard limitation of the firmware itself that makes local sub-device support impossible?
I have attached the diagnostic file for your reference.
Thank you again for your amazing work on this project!
Diagnostics file:

[config_entry-localthings-01KYYPK4A778E1Z7A77NVMWR2E.json](https://github.com/user-attachments/files/30690899/config_entry-localthings-01KYYPK4A778E1Z7A77NVMWR2E.json)

mbillow:
Your diagnostics confirm this isn't a bug: the subdevice's normal capability hrefs (`/power/0`, `/temperature/current/0`, `/mode/vs/0`, etc.) genuinely aren't advertised in `/oic/res` on this board at all — the firmware just doesn't publish sub-device state locally.

[...]

Issue #214, opened 2026-07-30
Air conditioner shows up duplicated


matiaskorhonen:
### Appliance model

AR12NXWXCWKNEU

### What's missing or wrong

My air conditioner now shows up as two connected devices, possibly because of the changes made to address #177 and/or #205. This was probably unintentional for non-2-in-1 devices?

**Devices listing:**

<img width="1513" height="199" alt="Image" src="https://github.com/user-attachments/assets/01b0419d-c390-421e-8e21-5a6986554c3f" />

**Under the parent device:**

<img width="480" height="168" alt="Image" src="https://github.com/user-attachments/assets/28948b86-3b0c-4476-8230-ff987caace6f" />

---

P.S. thanks for the work on LocalThings, I really appreciate it!

### Diagnostics file

[config_entry-localthings-01KYPXECB52AYHHV8P0642AZ0J.json](https://github.com/user-attachments/files/30545814/config_entry-localthings-01KYPXECB52AYHHV8P0642AZ0J.json)

mbillow:
Your device does report an actual subdevice:
```json
"subdevices": [
      {
        "kind": "indexed",
        "key": "1",
        "seed_path": "/device/1",
        "flat_hrefs": [],
        "bound_entity_count": 21,
        "hrefs": [
          "/alarms/vs/1",
          "/diagnosis/vs/1",
          "/energy/consumption/vs/1",
          "/humidity/vs/1",
          "/mode/vs/1",
          "/temperature/current/1"
        ],
        "model": "ARTIK051_KRAC_18K|10193441|60010123001111110100000000000000",
        "resources": {
          "/airflow/vs/0": {},
          "/airflow/0": {},
          "/alarms/vs/0": {},
          "/temperatures/vs/0": {
            "x.com.samsung.da.items": [
              {
                "x.com.samsung.da.id": "0",
                "x.com.samsung.da.description": "Temperature"
              }
            ]
          },
          "/temperature/current/0": {},
          "/temperature/desired/0": {},
          "/diagnosis/vs/0": {},
          "/energy/consumption/vs/0": {
            "x.com.samsung.da.cumulativePower": "117520000",
            "x.com.samsung.da.cumulativeDate": "1785420000"
          },
          "/energy/consumption/0": {},
          "/mode/vs/0": {},
          "/mode/0": {},
          "/power/vs/0": {},
          "/power/0": {},
          "/information/vs/0": {
            ...
          },
          "/file/information/vs/0": {
[...]

Issue #205, opened 2026-07-30
Samsung Home Multi 2-in-1 Air Conditioner (Floor-standing + Wall-mounted)


jhkwon19:
### Appliance model

TP2X_FAC_BORA_21K

### What's missing or wrong

I updated to the latest version (v0.17.0), but I still only get a single Climate entity in Home Assistant.

Based on our previous investigation in issue #177:
- My device uses the UUID-prefixed subdevice structure.
- The UUID successfully resolves to the wall-mounted indoor unit (confirmed by matching the serial number returned from /<UUID>/information/vs/0).
- The diagnostics include the subdevice information.

However, after upgrading to v0.17.0, no additional Climate device/entity is created for the wall-mounted indoor unit. Only the main floor-standing unit appears.

I'm attaching a fresh diagnostics file generated with v0.17.0. Hopefully it helps identify why the subdevice discovery isn't creating the second Climate entity.

### Diagnostics file

[config_entry-localthings-01KYRB2RSA0Y822KV5ZN9YSBS9.json](https://github.com/user-attachments/files/30528052/config_entry-localthings-01KYRB2RSA0Y822KV5ZN9YSBS9.json)

<img width="1449" height="293" alt="Image" src="https://github.com/user-attachments/assets/79666f0f-9998-4e6f-adc6-edc8b6ac7a9a" />

mbillow:
```json
    "subdevice_probes": {
      "/6c2dff6d-ee5c-dad1-6a5e-000000000001/device/0": false,
      "/device/1": false,
      "/device/2": false,
      "/multidevice/vs/0": true
    },
```

Hmm, it doesn’t look like the subdevice exposes a `/device/0` like I assumed. That’s unfortunate, I’m not sure how else we can discover available resources. I guess we could just look for all the same hrefs as the parent device, but that might not catch everything. 

Let me do some thinking.

mbillow:
Can you upgrade to v0.17.1 and see if that creates the sub-device properly?

jhkwon19:
> Can you upgrade to v0.17.1 and see if that creates the sub-device properly?

Wow! Thank you so much! The sub-device has been detected and registered, and it's working perfectly!
<img width="1058" height="198" alt="Image" src="https://github.com/user-attachments/assets/9981cfd0-5751-42e5-a56f-5f19eb508df3" />

mbillow:
[...]

PR #436, opened 2026-09-03
Fix UUID-prefixed subdevice discovery without device collection

onei12:
## Summary
- probe a UUID-prefixed subdevice's identity and highest-priority primary entity resource before its optional device collection
- retain confirmed flat resources when the collection is absent or consumes the discovery budget, while keeping a working collection authoritative
- prioritize climate binding metadata without hardcoding Samsung model names
- cover the ARTIK051 FAC BORA 2-in-1 resource surface and preserve existing collection discovery behavior

## Testing
- 1837 passed
- ruff format --check custom_components tests
- ruff check custom_components tests
- ty check custom_components tests

### DIFF
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -1109,13 +1109,14 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
         self,
         resources: dict[str, dict],
     ) -> tuple[str, ...]:
-        """Return identity, then live primary-entity hrefs in hot/warm-first order.
+        """Return identity, then live primary-entity hrefs in probe order.
 
         A prefixed subdevice without a Collection endpoint has to be probed
         one Property href at a time. Its identity selects its own registry and
-        device metadata, so probe it before the primary resources declared by
-        the master's registry. This remains device-type agnostic, while unknown
-        devices still get a useful first probe before retaining normal order.
+        device metadata, so probe it first; prefer a composite entity's binding
+        href next so one live response can materialize the useful device, then
+        retain the registry's hot/warm-first order. This remains device-type
+        agnostic, while unknown devices still get a useful first probe.
         """
         identity = ("/information/vs/0",) if "/information/vs/0" in resources else ()
         registry = resolve_registry(
@@ -1135,9 +1136,18 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
             ]
             if not primary_caps:
                 continue
-            rank = min(tier_rank.get(capability.poll_tier, 2) for capability in primary_caps)
-            ranked.append((rank, order, href))
-        return identity + tuple(href for _, _, href in sorted(ranked))
+            composite_rank = (
+                0
+                if any(
+                    isinstance(desc, ClimateDesc)
+                    for capability in primary_caps
+                    for desc in capability.entities
+                )
+                else 1
+            )
+            tier = min(tier_rank.get(capability.poll_tier, 2) for capability in primary_caps)
+            ranked.append((composite_rank, tier, order, href))
+        return identity + tuple(href for _, _, _, href in sorted(ranked))
 
     def _enumerate_subdevices_blocking(self, resources: dict[str, dict]) -> dict[str, dict]:
         """One-time (first discovery only) probe for sibling indoor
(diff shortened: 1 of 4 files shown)


mbillow:
Thanks for this — the diagnosis is a sensible one to reach, the branch is clean, tests pass, and it merges without conflict. I went looking for the hardware this would help before reviewing the code, and I came up empty. Laying out what I found, because I'd rather show the data than just say no.

## The boards this targets return nothing under the prefix

I pulled the diagnostics attached to #282 (`ARTIK051_FAC_TIME_21K`) and #335 (`ARTIK051_FAC_BORA_19K`). Between them, 54 UUID-prefixed probes, every one `false` — including the three leaves the preflight promotes:

```
False /<uuid>/information/vs/0
False /<uuid>/power/vs/0
False /<uuid>/mode/vs/0
```

#335's reporter also confirmed by hand that the prefix itself is routable — `/<uuid>/file/list/vs/0` returns 2.05 with real content. So those 4.04s are the appliance genuinely saying "no such resource," not a dead prefix swallowing everything. That was the whole point of that round of manual reads, and it's what makes the result conclusive rather than ambiguous.

Reordering probes can't recover a subdevice when none of the probes answer. On these two boards the outcome is unchanged.

## The one board where a flat probe *does* answer is already handled

Issue #205's unit answers exactly one leaf, `/<uuid>/information/vs/0`, and `test_fac_bora_205_flat_fallback_finds_candidate_but_gate_holds_it_back` already pins that behaviour: the candidate is found, and the materialization gate holds it back because identity binds no primary entity. Your branch keeps that test green, so that board's outcome is unchanged too.

## Where I'd push back hardest

`test_artik051_fac_bora_flat_subdevice_materializes_climate_and_device` builds its board rather than capturing one — it loads the TP2X `fac_bora_2in1` fixture, rewrites `modelNum` to `"ARTIK051_FAC_BORA_20K|TEST"`, and synthesizes the sibling's responses as:

```python
seeds = {f"/{_SUB_UUID}{href}": dict(resources[href]) for href in live_hrefs}
```

That's the master's own reps, copied under the subdevice prefix. It's the aliasing case the comment a few lines above the code you changed already flags as unguarded:

> so this hasn't been guarded against -- the fix would compare a candidate's confirmed reps against the master's own values for the same hrefs.

The test constructs that exact scenario and asserts the subdevice materializes, which makes the known hazard the expected behaviour. And we have real evidence of aliasing on this family: #335's `/<uuid>/file/list/vs/0` returned content byte-for-byte identical to the master's. Broadening reliance on the flat path without that comparison makes a phantom duplicate (issue #214's failure mode) more reachable, not less.

## One thing I got wrong while checking

I first read "all probes logged, none missing" as proof the budget was never exhausted on those boards. That's not right — #282 is v0.19.0 and #335 is v0.20.0, and the shared time budget only landed in v0.21.0, so those runs were unbounded and nothing would have been truncated either way. **Whether starvation actually bites on this firmware is untested.** What holds regardless is the part that matters here: every leaf returns false.

That's also the cheap way to settle it. A fresh diagnostic from either reporter on ≥0.21 would show a truncated `subdevice_probes` list if the Collection probe really is eating the budget, or a complete all-false list if it isn't. Happy to ask for one.

## If the starvation concern is real

I think there's a smaller fix for it. The Collection probe is allowed `min(collection_timeout, remaining)` — 10s of a 15s budget. Clamping it so it can never take more than a fraction of what's left:

```python
timeout = _next_timeout(min(collection_timeout, remaining / 2))
```

gets the same guarantee in one line, with no reordering, no preflight reads discarded when the batch answers, and no new reliance on the unvalidated flat path.

## Smaller notes on the code itself

Not blocking, since I don't think the approach should land as-is, but worth recording:

- `if priority_hrefs and priority_hrefs[0] == "/information/vs/0"` puts an href literal inside `subdevices.py`, which is otherwise deliberately registry-agnostic — the coordinator owns that ordering. `_subdevice_probe_priority` prepending identity is good; the module then re-checking the string couples them.
- `if len(flat_hrefs) == 2` as the "both preflight probes answered" sentinel is positional. It's correct today and breaks quietly if the preflight ever probes a third href.
- Broadening the priority list to every hot/warm capability increases the probe count, which spends more of the budget this is trying to protect.
- When both preflight probes answer on a board whose Collection also works, the whole priority list is read and then discarded.

## Where that leaves it

I don't want to merge a change that reorganizes the probe path without a device it demonstrably helps — this code is hard to reason about and every board it touches is one I can't test against. But the `/information/vs/0`-first idea is sound on its own merits (identity is what selects the sibling's registry), and if a dump turns up showing a truncated probe list, the starvation half becomes concrete and I'd want it fixed.

If you have hardware where the prefixed leaves *do* answer while `/<uuid>/device/0` doesn't, a diagnostics download from it would change this assessment immediately — that's the missing piece, not the code. Genuinely appreciate the time you put in here.

Issue #381, opened 2026-08-16
Duplicate unique_id issue due to same serial number on same models (Device is already configured)


gmlsxhfl:
I have two identical Samsung Air Purifiers (Model: AVT-WW-TP1-23-AXX500) with different IP addresses.
When I try to add the second one, I get a "Device is already configured" error.
I checked core.config_entries and found that both devices return the exact same serial number (BS7SP9AW400114A) from the firmware:

<img width="741" height="103" alt="Image" src="https://github.com/user-attachments/assets/50077ecc-2526-4ff2-a3c0-badb98d8e730" />

1st
> {"created_at":"2026-08-16T05:03:36.774463+00:00","manufacturer":"Samsung Electronics","model":"AVT-WW-TP1-23-AXX500","port":49154,"serial":"BS7SP9AW400114A"},"disabled_by":null,"discovery_keys":{},"domain":"localthings","entry_id":"01M04F8CT6D6X6JBGQFK7BF3YE","minor_version":1,"modified_at":"2026-08-16T05:03:36.774466+00:00","options":{},"pref_disable_new_entities":false,"pref_disable_polling":false,"source":"user","subentries":[],"title":"Samsung Air Purifier (192.168.0.3)","unique_id":"localthings_BS7SP9AW400114A","version":3},

2nd
> {"created_at":"2026-08-16T06:38:41.297408+00:00","manufacturer":"Samsung Electronics","model":"AVT-WW-TP1-23-AXX500","port":49154,"serial":"BS7SP9AW400114A"},"disabled_by":null,"discovery_keys":{},"domain":"localthings","entry_id":"01M04MPFMHWPMWCHH79Y0ZMAAJ","minor_version":1,"modified_at":"2026-08-16T06:38:41.297411+00:00","options":{},"pref_disable_new_entities":false,"pref_disable_polling":false,"source":"user","subentries":[],"title":"Samsung Air Purifier (192.168.0.14)","unique_id":"localthings_BS7SP9AW400114A","version":3}

Because the serials are identical, localthings generates the same unique_id. Could you please add a fallback logic in config_flow to append the MAC address or IP address to the unique_id if the device is a Samsung Air Purifier or if a collision occurs?

mbillow:
Can you provide a device diagnostic from either (or both) of these devices?

gmlsxhfl:
[...]

Issue #189, opened 2026-07-29
Device is already configured - when adding second device


larsbroberg:
I have separate washer and dryer.
I can add each one individually. Both are identified correctly.
But I can only have one device, and not both at the same time.

When adding the second device, using only the IP address I get "Device is already configured".

I've tried turning off the first device, so there is no communication, and adding the second but that doesn't help.

mbillow:
Can you provide dumps of each when added individually, please?

jeison-souza:
Hi, I have he same issue.

Didn't find anything screaming at me on the log when adding the second device. See image.  The IP 181 is a dishwasher and the IP 143 is a range.

In this case I added the dishwasher first, everything seems normal, (Entities are created) then when adding the range I got the message that the Device is already configured. 

If I do on the reversed order, range first then the dishwasher, the same message pops.  

<img width="1852" height="186" alt="Image" src="https://github.com/user-attachments/assets/1571184d-d718-4eee-89b2-473980d04e75" />

If you need any additional log or test o be done, please let me know.

larsbroberg:
Adding dryer:
`2026-07-29 19:21:09.423 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Fetching Samsung cloud UUID from connect-v2.samsungiotcloud.com
2026-07-29 19:21:09.626 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Got UUID: xxxxxxxxx
2026-07-29 19:21:09.626 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Minting leaf cert for UUID xxxxxxxxx
2026-07-29 19:21:10.140 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Leaf cert minted successfully
2026-07-29 19:21:11.858 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] Live DTLS port candidates on 192.168.4.220: [49154, 49155, 49153]
2026-07-29 19:21:24.186 DEBUG (SyncWorker_5) [custom_components.localthings.config_flow] port 49154 failed: DTLS handshake timeout to 192.168.4.220:49154
2026-07-29 19:21:26.393 DEBUG (SyncWorker_6) [custom_components.localthings.coordinator.192.168.4.220] DTLS connected to 192.168.4.220:49155
[...]

Issue #83, opened 2026-07-25
ARTIK051_DONGLE_REF fridge/freezer: type detection, serial-collision (2x), and missing door sensor — root causes + tested fixes


RPaulsson:
<html>
<body>
<p class="font-claude-response-body break-words whitespace-normal" dir="ltr" data-sourcepos="6:1-12:64;150-642">Two physically separate Samsung ARTIK051 appliances (one fridge, one
freezer — each with its own dongle, not a combo unit) on firmware
<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ARTIK051_DONGLE_REF_20190711</code> / <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">TizenRT2.0</code> hit <strong>four</strong> distinct bugs,
all stemming from two root causes on this firmware family. All four are
fixed and verified end-to-end against the real hardware (DTLS handshake,
authenticated <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">/information/vs/0</code> read, full HA add-flow, live entities in
production). Patches below are minimal and backward-compatible.</p>

<p class="font-claude-response-body break-words whitespace-normal" dir="ltr" data-sourcepos="14:1-15:33;644-751">This also directly explains and fixes <strong>#81</strong> (door sensor not created for
<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ARTIK051_REF_17K</code>) — see Bug 3.</p>

<h2 dir="ltr" class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold" data-sourcepos="17:1-17:15;753-767">Environment</h2>

<ul dir="ltr" class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3 print:block print:space-y-1" data-sourcepos="19:1-22:74;769-984">
[...]

PR #443, opened 2026-09-06
Record the OCF device UUID separately from the registry key

mbillow:
First step of the plan in #435. Standalone and useful on its own merits; nothing here depends on any of the credential work being settled.

## Why

`CONF_DEVICE_KEY` is what devices and entities are keyed on. `resolve_device_key` produces it from the first **usable** value in a fallback chain: `/oic/d`'s `di`, then `/oic/p`'s `pi`, then the serialNum, then the host. Three of those four branches exist because firmware families have shipped unusable identity fields — `Nothing(SVC)` (#83), a nil UUID (#189), a serial duplicated across every unit (#381).

That makes it an excellent registry key and poor evidence of OCF identity. `pi` is platform-scoped and shared by every logical device on one board; a serial or an address says nothing about which OCF device answered.

Nothing needs that distinction today. The credential profiles in #435 do: binding a PSK session to the device its credential was issued for means comparing an authenticated `/oic/d.di` against a stored value, and comparing against a platform UUID or a serial instead produces a device-mismatch failure on a correctly configured appliance. Both @Jason-Morcos and @atc722 independently asked for these to be separate values.

## What this does

- `proven_ocf_device_id()` in `registry/identity.py` — deliberately narrower than `ocf_device_key`: `di` or nothing, no fallback, same normalization so the two are comparable when they agree.
- `CONF_OCF_DEVICE_ID` on the entry, written by the config flow at creation (the probe already reads `/oic/d`, so this costs no extra round trip) and backfilled by the coordinator on the first live poll.
- A read reporting no usable `di` leaves a stored value alone rather than erasing it.
- A snapshot replay writes nothing — it never reached the device, the same reason it must not write a device key (#295).
- Exposed in the `identity` block of diagnostics. `di`/`pi` are deliberately unredacted (see `registry/redact.py`), so this adds no identifier the resource dump doesn't already carry.

**No config-entry migration.** Absent is a legitimate state for an entry that has never polled, so this backfills rather than migrates — which also leaves the next version bump free for the carrier/profile model where one is actually earned.

**Store-only.** Nothing acts on a change yet. What a *different* `di` means — and the missing / mismatch / bad-credential distinction — belongs with the credential profiles. Landing the storage first means those arrive with real values already on entries.

## Validation

- 1834 → 1845 tests; 11 new, covering the helper's refusal to fall back to `pi`, the nil-UUID and junk-field cases, config-flow write and non-write, first-poll backfill, no erasure on an unproven read, no write on snapshot replay, and that the registry key and `unique_id` are untouched.
- `ruff format`, `ruff check`, `ty check` all clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_0141wUYNnqLYz1kL5gscWRiq)_

### DIFF
--- a/custom_components/localthings/config_flow.py
+++ b/custom_components/localthings/config_flow.py
@@ -51,6 +51,7 @@ from .const import (
     CONF_LEARN_MODES,
     CONF_MANUFACTURER,
     CONF_MODEL,
+    CONF_OCF_DEVICE_ID,
     CONF_PORT,
     CONF_SERIAL,
     DEFAULT_CLOUD_COURSES_ENABLED,
@@ -664,6 +665,7 @@ def _read_device(sess, host: str, port: int) -> dict:
     from .registry.batch import parse_device0_batch
     from .registry.by_type import resolve as resolve_registry
     from .registry.identity import (
+        proven_ocf_device_id,
         read_identity,
         resolve_device_key,
         resolve_model,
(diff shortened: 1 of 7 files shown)


mbillow:
Closing for #447

Issue #450, opened 2026-09-07
Gas range: cooktopMonitoring is a bitmask of lit burners (NX60T8311SS)


Averyy:
### Appliance model

NX60T8311SS/AA, `TP2X_DA-KS-RANGE-0101X`. Same gas range as #444.

### What's missing or wrong

`/cooktopmonitoring/vs/0` carries a field the integration reads and never exposes:

```json
"x.com.samsung.da.cooktopMonitoring": "31"
```

It's a bitmask of lit burners, one bit per knob. I polled the resource every 2 s over a second DTLS session from a laptop while lighting each burner on its own:

| Burner | Value | Bit |
|---|---|---|
| front-left | 1 | 0 |
| back-left | 2 | 1 |
| center | 4 | 2 |
| back-right | 8 | 3 |
| front-right | 16 | 4 |

Then all five on one at a time: 1, 3, 7, 23, 31. Off one at a time: 29, 25, 24, 8, 0. Turning a burner from high down to its lowest simmer didn't move the value. `cooktopRunningState` reads `Run` whenever the mask is non-zero and `Ready` at zero, and the `SettingPossible_` token in `/mode/vs/0`'s options goes `_0` to `_2` on the same condition.

Every range fixture in `tests/` carries the field, the electric NE63A6511, NE6516A, NE63T8751SG and NE8300D included, but they were all captured idle so each reads `"0"`. #404's gas NX6512A has it too. So this is verified on one gas range and assumed for the rest. I don't have an electric one to check, and I'd guess the boards that also publish `/cooktop/status/vs/0` will show the same burners twice.

### Proposal

On `COOKTOP_MONITORING`, one binary sensor per bit (`Burner 1` to `Burner 5`, on/off) and an `Active burners` count. Five bits because that's what this range has and the resource doesn't say how many burners a board carries. The position table stays out of the code since it's per model; it's here so other owners of this range can rename their entities. PR coming right after this.

Raw monitor log from both passes:

```
10:19:28 START  monitoring=0 running=Ready SettingPossible_0
10:20:23 CHANGE monitoring=1 running=Run SettingPossible_2      front-left on
10:20:36 CHANGE monitoring=1 running=Run SettingPossible_0
10:20:39 CHANGE monitoring=0 running=Ready SettingPossible_0    front-left off
10:20:48 CHANGE monitoring=8 running=Run SettingPossible_2      back-right on (high, then simmer: no change)
[...]

Issue #404, opened 2026-08-21
Sync Time


lhassell:
### Appliance model

Samsung Range TP1X_DA-KS-RANGE-0101X

### What's missing or wrong

I don't even know if it's possible, but the only thing missing from the original app is the ability to sync/set the time of day on the display. I can do it through the smartthings app, but it would be nice if I did not have to use the app at all. Is this doable?

### Diagnostics file

{
  "home_assistant": {
    "arch": "aarch64",
    "dev": false,
    "docker": true,
    "hassio": true,
    "installation_type": "Home Assistant OS",
    "os_name": "Linux",
    "os_version": "6.18.39-haos",
    "python_version": "3.14.6",
    "timezone": "America/Chicago",
    "version": "2026.8.2",
    "virtualenv": false,
    "container_arch": "aarch64",
    "supervisor": "2026.07.5",
    "host_os": "Home Assistant OS 18.2",
    "docker_version": "29.6.2",
    "chassis": "embedded",
    "run_as_root": true
  },
  "custom_components": {
    "material_symbols": {
      "documentation": "https://github.com/beecho01/material-symbols",
      "version": "2026.07.08",
      "requirements": []
    },
    "spook_inverse": {
      "documentation": "https://spook.boo",
      "version": "5.0.0",
      "requirements": []
    },
    "oidc_provider": {
      "documentation": "https://github.com/ganhammar/hass-oidc-server",
      "version": "1.5.2",
      "requirements": [
        "PyJWT>=2.8.0",
        "cryptography>=43.0.0"
      ]
    },
    "lennoxs30": {
      "documentation": "https://github.com/PeteRager/lennoxs30/blob/master/README.MD",
      "version": "2024.8.0",
      "requirements": [
        "lennoxs30api==0.2.15"
      ]
    },
    "localthings": {
      "documentation": "https://github.com/mbillow/localthings",
      "version": "0.23.0",
      "requirements": [
        "cbor2>=5.4.6",
        "pyOpenSSL>=23.0",
        "smartthings-local>=0.1.8"
      ]
    },
    "wundergroundpws": {
      "documentation": "https://github.com/cytech/Home-Assistant-wundergroundpws/",
      "version": "2.0.7",
      "requirements": []
    },
    "meshtastic": {
      "documentation": "https://github.com/kamilzierke/home-assistant-meshtastic",
      "version": "0.8.2",
      "requirements": [
        "pyserial-asyncio-fast>=0.16",
[...]

PR #428, opened 2026-09-01
Add safe range clock synchronization

Jason-Morcos:
## What this changes

Adds a config-category `Sync clock` button for ranges that expose `/configuration/vs/0`.

The timestamp is generated when the button is pressed from Home Assistant's configured local wall clock and sent as `x.com.samsung.da.currentTime` in `YYYY-MM-DDTHH:MM:SS` form. There is no arbitrary date input, and clock sync does not run automatically during setup or reconnect.

On my TP1X range, the POST returned CoAP 2.04 Changed and a follow-up connectivity refresh succeeded. GET continued to return only the resource metadata, so the resource is handled as write-only: the command is not added to the optimistic representation cache, while the normal post-write refresh still runs.

This also matches the TP1X_DA-KS-RANGE-0101X report in #404.

## Validation

- frozen-clock coverage for Home Assistant local-time formatting
- invalid and naive-clock rejection
- resource-presence and range-registry gating
- coordinator coverage proving the write-only value is not cached or guarded as settled state
- full test suite: 1,839 passed
- Ruff format/lint and ty

Fixes #404

PR #451, opened 2026-09-07
Expose the range's lit-burner bitmask as per-burner sensors

Averyy:
Adds a count and one binary sensor per lit burner from `/cooktopmonitoring/vs/0` `cooktopMonitoring`, decoded as a bitmask (issue #450).

## What the value is

Lit one burner at a time on a gas NX60T8311SS/AA (TP2X_DA-KS-RANGE-0101X) and read the resource over DTLS every two seconds: bit 0 front-left, bit 1 back-left, bit 2 center, bit 3 back-right, bit 4 front-right. On/off sweeps in both directions (1, 3, 7, 23, 31 then 29, 25, 24, 8, 0) and turning a burner down to simmer (stays 8) all fit. `cooktopRunningState` is `Run` exactly when the mask is non-zero.

## What gets created

- Boards reporting `Fuel_Gas` in `/mode/vs/0` options[]: `active_burners` (popcount) and `cooktop_burner_1..5`.
- Every other board with the field: one diagnostic sensor, `cooktop_monitoring`, carrying the raw value.
- Boards that also publish a `/cooktop/status/vs/0` `burnerList`: burners come from there and the decoded set stands down.

The gate is the evidence boundary. The bit map was proven on gas; the two gas boards in the corpus (this one and #404's TP1X NX6512A) both carry `Fuel_Gas`, and none of the five electric NE fixtures carry any `Fuel_` token. Their masks have only ever read 0, so on them the raw value is the only entity: no five permanently-off burners on a board nobody has confirmed, and an owner who lights two burners and sees 3 has the number to post on #450. Lifting the gate is then a one-line change.

Width stays at five. The mask can't say how many burners exist, and widening to `MAX_BURNERS` would give the one confirmed unit a permanently-off Burner 6.

## Validation

- New fixture `range_nx60t8311ss` captured with all five burners lit (mask 31), golden added
- `tests/test_range_nx60t8311ss_capabilities.py`: the bit table from the live sweeps; gas gets the decoded set; no fuel token gets the raw value only; a stub `/mode/vs/0` keeps the optimistic carve-out; `burnerList` stands the set down; field absent creates nothing
- Five electric range goldens: the six decoded keys replaced by `cooktop_monitoring`
- `ruff format --check`, `ruff check`, `ty check`, `pytest tests/ -q`: 1943 passed


mbillow:
Reviewed. The investigation behind this is excellent — isolating each bit by lighting one burner at a time, the on/off sweeps in both directions (1,3,7,23,31 → 29,25,24,8,0), and checking that flame level *doesn't* move the value is exactly the evidence needed to call it a bitmask rather than a count or an enum. The code comment disclosing that only the gas unit was seen non-zero, so electric boards remain an assumption, is the right way to record that.

## Blocking, but mechanical: needs a rebase

This currently conflicts with `main`. #428 (range clock sync) landed and added `from datetime import datetime` to the same import block you added `from ..batch import is_stub_rep` to. That's the entire conflict:

```python
from datetime import datetime

from ..batch import is_stub_rep
from ..capability import Capability
```

I resolved it locally against current `main` and everything else is clean — **1917 passed**, `ruff check`/`format` and `ty` all green. So a rebase should be the only thing standing between this and a merge.

## Two things worth considering while you're in there

**1. Duplicate burners on boards carrying both resources.** You called this in the issue ("I'd guess the boards that also publish `/cooktop/status/vs/0` will show the same burners twice"), and I think you're right. `exists_fn` receives `resources` as its second argument, so the mask sensors can stand down when `burnerList` is present:

```python
def _has_burner_mask(rep, resources):
    if (resources.get("/cooktop/status/vs/0") or {}).get("burnerList"):
        return False
    return is_stub_rep(rep) or _burner_mask(rep) is not None
```

Worth noting this is forward-looking rather than observed: I checked the corpus and no fixture carries both today (`range_device.json` has a 4-entry `burnerList` and no mask; the five mask-bearing fixtures have no `burnerList`). So nothing is broken right now — it's cheap insurance against the first board that has both.

**2. `_MASK_BURNERS = 5` vs this file's existing pattern.** `range.py` already hardcodes `MAX_BURNERS = 6` but gates each index with `_burner_exists(i)` against what the device actually reports — the module docstring calls that out as deliberate. The mask genuinely can't tell you how many burners exist, so there's no equivalent gate available from it alone, and I don't think that blocks anything. But it does mean a 4-burner electric range gets a permanently-off "Burner 5", and the goldens show all five mask fixtures picking up six new entities each on boards where the mask has only ever read `0`.

If you'd rather not widen that, gating the whole set on a non-zero mask having been *seen* isn't possible statelessly — but deriving the count from `burnerList` when it's present would at least make the both-resources case exact rather than approximate.

Neither is a correctness defect; the rebase is the only thing I'd call required.

Averyy:
Rebased onto current main (the import block was the whole conflict, resolved the way you showed) and took the first suggestion as a second commit: `_has_burner_mask` returns False when `/cooktop/status/vs/0` carries a non-empty `burnerList`, so all six mask entities stand down on a board with both resources. Two tests cover it, one splicing `range_device.json`'s `burnerList` into the NX60T8311SS capture and one with an empty list where the mask stays.

On the width: with the stand-down in place, a board with `burnerList` never gets mask entities, so deriving the count from it has nothing left to apply to. Widening to `MAX_BURNERS` would give the one confirmed unit a permanently-off Burner 6, and the mask alone can't tell a 4-burner board from a 5-burner one that hasn't lit bit 4. So it stays at five with the reason in the comment above `_MASK_BURNERS`. Happy to change that if you'd rather match the file's six.

1940 passed, `ruff format`/`check` and `ty` clean.

Averyy:
Thought about the width point more and changed my answer. Third commit, 8ea7d08.

The width question sits on top of a bigger one. The bit map was verified on one gas unit, and the PR was creating six entities on every electric board in the corpus anyway, on masks that have only ever read 0. That's the same phantom-entity shape this repo keeps removing elsewhere.

There's a device-reported signal that separates the two. Both gas boards in evidence, my NX9302T and #404's TP1X NX6512A, carry `Fuel_Gas` in `/mode/vs/0` options[]. None of the five electric NE fixtures carry any `Fuel_` token. So:

- `Fuel_Gas`: `active_burners` plus `cooktop_burner_1..5`, as before.
- No `Fuel_Gas`: one diagnostic sensor, `cooktop_monitoring`, with the raw value. No count, no burners.
- `burnerList` present: the decoded set stands down, from the second commit.

The raw sensor is what makes the gate cheap to lift. An electric owner lights two burners, sees 3 in HA, comments on #450, and dropping the gate becomes a one-line change with evidence behind it. Until then there's no permanently-off Burner 5 on a four-element board.

Width stays at five: it's the layout the bits were verified on, and six would give the one confirmed unit a phantom. The stub carve-out follows the existing convention, a `/mode/vs/0` not fetched yet at discovery counts as pass the same way every exists_fn treats its own rep, with a test.

While in there I fixed a wrong docstring from the first commit. It said #404's NX6512A reports the resource without the field. It doesn't; that dump has `cooktopMonitoring: "0"` and `Fuel_Gas`. The field-absent test stays as a defensive case, no longer attributed to #404.

Five electric goldens change from the six decoded keys to `cooktop_monitoring`. 1943 passed, ruff format/check and ty clean. PR description rewritten to match.

mbillow:
Second pass, and this is ready. Verified locally on `pr451` at 8ea7d08: `ruff format --check`/`ruff check` on `custom_components`+`tests` clean, `ty check` clean, **1943 passed**.

What I checked beyond the diff reading:

- **The gate's evidence holds.** #404's dump is `TP1X_DA-KS-RANGE-0101X`, `DeviceType_NX6512A-/AA0`, and it does carry `Fuel_Gas` with `cooktopMonitoring: "0"` — so both gas boards in evidence carry the token, and none of the six non-gas range fixtures carry any `Fuel_` token (I dumped every range fixture's `/mode/vs/0` options[] to confirm). The gate separates exactly the set you say it does.
- **Stub semantics.** `_fuel_is_gas`'s optimistic carve-out matters more than it looks, because it's gating on a *sibling* href rather than the entity's own — but no range fixture ships a stub `/mode/vs/0`, every one carries real options, so the carve-out is the discovery-order safety net and not the common path. `option_value`'s position-0 anchor means `Fuel` can't be matched by some other token's suffix.
- **`burnerList` key.** The stand-down reads the bare `burnerList`, which is what `_burner_exists`/`burner_write` already use and what `range_device.json` actually reports. Consistent.
- **Fixture hygiene.** `serialNum` redacted, no DUID/MAC/SSID, `seeds_note` says what was captured and why. Goldens for the five electric boards swap the six decoded keys for `cooktop_monitoring` as described, and the new golden test follows the file's local-import convention.
- **Translations.** All seven locales, genuinely translated rather than English copies, and `{number}` placeholder wired through `translation_placeholders`.

Taking the width argument as settled — you're right that once `burnerList` stands the set down, there's nothing left to derive a count from, and five is the layout the bits were verified on.

One factual nit, not worth another round trip: the comment says *"Every -0101X/-0102X range dump carries the field"*, but `range_device.json` is a `TP1X_DA-KS-RANGE-0102X` with no `/cooktopmonitoring/vs/0` resource at all (it's the `burnerList` board). Every -0101X/-0102X dump that *has* the resource carries the field, which is the claim you actually need. I'll trim that clause in a follow-up rather than hold the merge.

Merging.

Issue #449, opened 2026-09-07
Samsung NASA Airconditioners - Filter Reset Support


jayrage887:
### What's missing or wrong

I couldn´t find an issue w.r.t. this topic, but the SmartThings app allows to reset the Filter-Usage for Air Conditioners, is there a way to get this implemented in LocalThings?

Currently Available are:

select.bedroom_samsung_airconditioner_bedroom_tp1x_da_ac_rac_01001_0000_filter_alarm_threshold
sensor.bedroom_samsung_airconditioner_bedroom_tp1x_da_ac_rac_01001_0000_filter_status
sensor.bedroom_samsung_airconditioner_bedroom_tp1x_da_ac_rac_01001_0000_filter_usage
sensor.bedroom_samsung_airconditioner_bedroom_tp1x_da_ac_rac_01001_0000_filter_usage_hours

But there is no way to clear that usage back to 0% like there is in SmartThings (e.g. if you wash the Filter and want to reset it via homeassistant).

KR

mbillow:
I just figured out how to reset the water filter on my fridge in #452, so there is a write for you to try in the PR body. Let me know if you need better instructions or if the write works and resets the air filter as well. You could use either the write debug flow in the options menu or the `localthings.write_resource` action.

Either way, to save you some time, here is the write again:

href: `/filter/airdustfilter/vs/0`
body: `{"x.com.samsung.da.filterReset": "On"}`

jayrage887:
> I just figured out how to reset the water filter on my fridge in [#452](https://github.com/mbillow/localthings/pull/452), so there is a write for you to try in the PR body. Let me know if you need better instructions or if the write works and resets the air filter as well. You could use either the write debug flow in the options menu or the `localthings.write_resource` action.
> 
> Either way, to save you some time, here is the write again:
> 
> href: `/filter/airdustfilter/vs/0` body: `{"x.com.samsung.da.filterReset": "On"}`

Already replied in the pull request haha, it worked see my answer there :)

<img width="1440" height="1656" alt="Image" src="https://github.com/user-attachments/assets/f37f5ee7-a596-4a1b-8636-7d5addd0b012" />

mbillow:
Perfect, let me get this wired up for HVAC units as well. Thanks for confirming!
I'll get this out in the next release.

jayrage887:
> Perfect, let me get this wired up for HVAC units as well. Thanks for confirming! I'll get this out in the next release.

[...]

PR #452, opened 2026-09-07
Add a water-filter reset button

mbillow:
## What this changes

Adds a `filter_reset` button to `common.WATER_FILTER`. Pressing it POSTs
`{"x.com.samsung.da.filterReset": "On"}` to `/filter/waterfilter/vs/0`.

Measured on a TP1X_REF_21K (RF29DB9750QLAA, firmware
`A-RFWW-TP1-24-T4-COM_20260617`): `filterUsage` `"100"` → `"0"`,
`filterStatus` `"replace"` → `"normal"`, still zero on a fresh DTLS session
and on the integration's own poll afterwards. The value is case-sensitive —
`"on"` and `"ON"` both fault 5.00.

## How it was found

The field appears nowhere on this device: not in its rep, `/oic/res`, its
`/device/0` batch, or `smartthings-local`. The control that proves it exists
is a near-miss name — `filterResetZZZ` is swallowed with 2.04 like any
unrecognised field, while `filterReset` faults.

The trick is that the 5.00 means the opposite of what it looks like. Every
non-string (`true`, `0`, `1`, `["replaceable"]`) returns 2.04 because
`oc_rep_get_string` fails and the handler never enters the reset block — not
because it was accepted. A string gets in and then misses on the value. So
**5.00 means the right code path was reached**, which turns an unbounded
value space into a short vocabulary of command words.

`docs/investigations/fridge-water-filter-reset.md` records this alongside the
traps: 2.04 is meaningless on this firmware (it ACKs unknown field names), the
POST response body is a verbatim echo with no diagnostic, and
`write_resource`'s `changed` flag only means "payload values present after".

## Scope, and what is not verified

Gated on `filterResetType` naming a reset the device says it supports
(`replaceable`/`washable`; the corpus also carries `notresetable`). That
reaches **four fridge families and three water purifiers**, of which only
TP1X_REF_21K is verified on hardware — the other six are trusted on their own
advertised claim. The full list is pinned in a test so the reach fails visibly
rather than widening unnoticed.

One caveat worth reading before merge: `dishwasher_device.json` reports
`filterReset` as a stored `"00"` on its own `/filter/waterfilter/vs/0`, so the
accepted value evidently varies by family and `"On"` should not be assumed to
travel. No dishwasher gets the button today (`filterStatus` is `notused`, which
`match_fn` rejects, or the rep is empty), but a wrong token would fault 5.00
with nothing surfaced to the user. Narrowing this to the fridge registry is a
reasonable call if you'd rather not ship on an advertised claim alone.

Also documented rather than fixed: the write body carries only the trigger, so
the sensors keep reading pre-reset values for up to 43 s after a press.
Closing that means adding fields to the body, and the body is what goes on the
wire, where only `{"filterReset": "On"}` exactly is verified — and the counter
only climbs, so there is no second reset to re-measure for months.

## Validation

- exact write-payload coverage, and gate coverage against `replaceable` /
  `washable` / `notresetable` / empty / absent
- stub-rep coverage: an explicit `exists_fn` bypasses `entity._is_included`'s
  carve-out, which would drop the button for the life of a config entry
  (#127); a genuinely empty `{}` rep stays excluded
- blast radius pinned across all seven families that advertise a reset
- full suite: 1787 passed. Three `test_config_flow` failures and two
  collection errors under `tests/localthings/` are pre-existing on `main`
  (stale `smartthings-local` missing `smartthings_local.errors`), confirmed by
  stashing this branch
- Ruff check + format clean

## Related

@Jason-Morcos — this closes the question you left open in #429. Your negative
result reproduces exactly here (`filterUsage: "0"` → 2.04, reverted on
readback), and your conclusion was right: `filterResetType` alone does *not*
establish a working reset. What it couldn't show is that an unadvertised
`filterReset` field exists next to it. Worth a look since #429 documents the
opposite guidance and we should land one story — and note the `"00"` value on
the dishwasher fixture above, which supports your caution about generalising.

@jayrage887 — #449 asks for this on air conditioners, which this PR does not
address. But it may be directly testable: `ac-filter-reset.md`'s negative
result for boards that keep the counter in `/filter/airdustfilter/vs/0` only
ever tried `filterUsage` writes and the `/mode/vs/0` token — never a
`filterReset` field, because nobody knew it existed. If you can try
`{"x.com.samsung.da.filterReset": "On"}` against `/filter/airdustfilter/vs/0`
via `localthings.write_resource` and re-read, that would settle it. A 5.00
rather than 2.04 is the encouraging answer, for the reason described above.

https://claude.ai/code/session_01ABdmqZ4jsc7Kjx15hGBReF

### DIFF
--- a/custom_components/localthings/registry/capabilities/common.py
+++ b/custom_components/localthings/registry/capabilities/common.py
@@ -618,6 +618,23 @@ WATER_FILTER = Capability(
             options=("normal", "wash", "replace"),
             value_fn=lambda value: value.lower() if isinstance(value, str) else value,
         ),
+        # filterReset 'On' (case-sensitive), measured on a TP1X_REF_21K:
+        # filterUsage 100 -> 0, filterStatus replace -> normal. A trigger the
+        # board never reports back, so it's gated on filterResetType -- the
+        # device's own claim that it supports a reset -- rather than on
+        # itself. See docs/investigations/fridge-water-filter-reset.md.
+        ButtonDesc(
+            key="filter_reset",
+            field="",
+            payload="On",
+            icon="mdi:restart",
+            entity_category="config",
+            exists_fn=lambda rep, resources: "x.com.samsung.da.filterResetType" in rep,
+            write_fn=lambda p, rep, href=None: (
+                ["filter", "waterfilter", "vs", "0"],
+                {"x.com.samsung.da.filterReset": p},
+            ),
+        ),
     ),
 )
 
--- a/custom_components/localthings/registry/capabilities/common.py
+++ b/custom_components/localthings/registry/capabilities/common.py
@@ -619,17 +619,28 @@ WATER_FILTER = Capability(
             value_fn=lambda value: value.lower() if isinstance(value, str) else value,
         ),
         # filterReset 'On' (case-sensitive), measured on a TP1X_REF_21K:
-        # filterUsage 100 -> 0, filterStatus replace -> normal. A trigger the
-        # board never reports back, so it's gated on filterResetType -- the
-        # device's own claim that it supports a reset -- rather than on
-        # itself. See docs/investigations/fridge-water-filter-reset.md.
+        # filterUsage 100 -> 0, filterStatus replace -> normal. A trigger this
+        # board never reports back, so it can't be gated on itself -- see
+        # docs/investigations/fridge-water-filter-reset.md.
+        #
+        # Gated on filterResetType naming a reset this device says it
+        # supports: the corpus also carries ['notresetable'], which a bare
+        # presence check would read as a yes. is_stub_rep keeps the stub
+        # carve-out an explicit exists_fn would otherwise bypass (same
+        # reasoning as ENERGY_METER above).
         ButtonDesc(
             key="filter_reset",
             field="",
             payload="On",
             icon="mdi:restart",
             entity_category="config",
-            exists_fn=lambda rep, resources: "x.com.samsung.da.filterResetType" in rep,
+            exists_fn=lambda rep, resources: (
+                is_stub_rep(rep)
+                or bool(
+                    {"replaceable", "washable"}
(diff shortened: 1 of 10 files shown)


jayrage887:
I manually did 

[
  {
    "href": "/filter/airdustfilter/vs/0",
    "payload": {
      "x.com.samsung.da.filterReset": "On"
    },
    "settle": 2
  }
]

on my living room ac and it worked as you expected

took around ~1m to actually show up though, probably the same 43 seconds? 

device_id: .....
results:
  - href: /filter/airdustfilter/vs/0
    actual_href: /filter/airdustfilter/vs/0
    code: '2.04'
    raw_code: 68
    accepted: true
    before:
      x.com.samsung.da.filterUsageResolution: '1'
      x.com.samsung.da.filterCapacity: '500'
      x.com.samsung.da.filterCapacityUnit: Hour
      x.com.samsung.da.filterResetType:
        - replaceable
        - washable
      x.com.samsung.da.filterStatus: normal
      x.com.samsung.da.filterUsage: '86'
      x.com.samsung.da.filterDesiredUsage: '500'
      x.com.samsung.da.supportedFilterDesiredUsage:
        - '180'
        - '300'
        - '500'
        - '700'
    after:
      rt:
        - x.com.samsung.da.filter.airdust
      if:
        - oic.if.baseline
        - oic.if.a
      x.com.samsung.da.filterUsageResolution: '1'
      x.com.samsung.da.filterCapacity: '500'
      x.com.samsung.da.filterCapacityUnit: Hour
      x.com.samsung.da.filterResetType:
        - replaceable
        - washable
      x.com.samsung.da.filterStatus: normal
      x.com.samsung.da.filterUsage: '86'
      x.com.samsung.da.filterDesiredUsage: '500'
      x.com.samsung.da.supportedFilterDesiredUsage:
        - '180'
        - '300'
        - '500'
        - '700'
    changed: false


this is what i got back in the developer tools

PR #445, opened 2026-09-06
Fix four idle-state reads on ovens: setpoint, cooking mode, cavity temperature, progress

Averyy:
Fixes #444.

Four idle-state reads on ovens, all from the same "the oven is off" condition. Three commits.

### 1. `oven_setpoint` reads 0 (-18 °C on a metric install)

While the oven is off, the board reports `desired: "0"` on `/temperatures/vs/0`. That's a placeholder for "no setpoint": it's below both minimums in `oven.py` (30 °C, 175 °F), so nothing you can dial in produces it. `oven_setpoint` passed it through as a real value; on a Fahrenheit board HA converts 0 °F for a metric install and the number entity sat at -18 °C. Every range and oven fixture in `tests/` has the same idle `desired: "0"`.

A new `_setpoint()` helper returns `None` when `desired` is `0`. Same treatment `_finish_time` already gives a zero `remainingTime`.

### 2. `oven_mode` shows Unknown while idle

`/mode/vs/0` reports `modes: ["NoOperation"]` when idle but leaves `NoOperation` out of `supportedModes`. The select's option list is that live list, and HA's `SelectEntity.state` returns `None` for a current option that isn't in `options`, so the idle state rendered as Unknown even though the catalog already has `no_operation`. Same shape on every range fixture, #404's range and #277's wall oven; the wall-oven and microwave fixtures include the token, which is why it works there.

`_oven_mode_options()` now prepends `NoOperation` when the live list lacks it. `_oven_mode_write()` still validates against `supportedModes`, so it's display-only. This is the oven case that `learned.py` deliberately keeps its allowlist away from (#327), so the fix lives in the option builder instead.

### 3. `current_temp_c` reads 175 °F while idle

Range firmware floors `current` at Bake's `tempMin` while idle: `175` on every Fahrenheit range fixture and mine, `80` on the Celsius one, matching each unit's `modeSpec`. Wall ovens (#300, #277) report `0`. The sensor forwarded both as measurements, so the history graph is a flat line at 175 with a spike per cook, and with `state_class: measurement` that goes into long-term statistics.

`_current_temp()` returns `None` only when there's no active setpoint **and** `current` is at or below the idle floor for the unit. Both conditions matter, per the #160 rule against broad sentinel collapses: a cool-down after a cook (desired 0, current above the floor) still reads, and a KeepWarm cook at exactly 175 °F (desired 175) does too. Preheat from cold shows the floor, same as the panel.

[...]

PR #455, opened 2026-09-07
Raise a user-facing error when NoOperation is picked as an oven mode

Averyy:
Follow-up to #445, from the review note there.

## What this changes

Listing `NoOperation` in the oven mode select fixed the Unknown-while-idle display but also made it selectable. Picking it hit `_oven_mode_write`'s `None` and the coordinator's `write_fn rejected payload` warning, so the user saw nothing happen.

`SelectDesc` now has the `validate_fn` hook `SwitchDesc` already had. The coordinator already ran it generically ahead of `write_fn`, so no dispatch change was needed. The oven select's `validate_fn` returns `oven_mode_idle_not_selectable` for `NoOperation`, and the coordinator raises `ServiceValidationError` with the translated message:

> NoOperation is what the oven reports while idle. It can't be selected as a mode; cancel the cook on the oven's panel instead.

Added to all seven translation catalogs. The registry stays free of HA imports. Every other option in the list comes from `supportedModes` and passes through unchanged.

## Validation

- `tests/test_oven_capabilities.py`: `validate_fn` returns the key for `NoOperation` and `None` for every supported mode
- `tests/localthings/test_coordinator.py`: selecting `NoOperation` on the real `OVEN_MODE` descriptor raises with domain and key and nothing is posted; `Bake` on the same entity still writes to `/mode/vs/0`
- `ruff format --check`, `ruff check`, `ty check`, `pytest tests/ -q`: 1923 passed

### DIFF
--- a/custom_components/localthings/coordinator.py
+++ b/custom_components/localthings/coordinator.py
@@ -2036,7 +2036,7 @@ class LocalThingsCoordinator(DataUpdateCoordinator[dict[str, Any]]):
         """Write a value to the device. Retries once on a dead session
         (issue #294); raises HomeAssistantError if that retry fails too.
 
-        A description-level validate_fn (SwitchDesc only, currently) rejects
+        A description-level validate_fn (SwitchDesc and SelectDesc) rejects
         a write with a user-facing message ahead of write_fn's silent
         no-op. The remote-control check runs first, unconditionally, unless
         the user opted out via CONF_BYPASS_REMOTE_CONTROL (issue #54: some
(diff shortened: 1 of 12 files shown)


mbillow:
Merged. This is exactly the right shape — thanks for turning it around so quickly.

Verified against current `main`: **1923 passed**, `ruff check`/`format` and `ty` clean, matching your numbers.

Two things I checked rather than assumed, both of which back up your description:

- **The dispatch really is generic.** `coordinator.py:2079` reads `getattr(desc, "validate_fn", None)`, so adding the field to `SelectDesc` is sufficient and your `coordinator.py` diff is correctly just a docstring correction. Nice that this needed no new plumbing.
- **The catalog key resolves.** `oven_mode_idle_not_selectable` lands under `.exceptions`, the same section as `remote_control_disabled`, and is present in all seven catalogs — so the message actually renders instead of falling back to the raw key.

The coordinator test is the part I'd have asked for if it weren't already here: it asserts the domain and key, that nothing was posted, *and* that `Bake` on the same entity still writes. That last assertion is what stops a future over-broad `validate_fn` from silently disabling the whole select.

Registry stays HA-free, which was the constraint that made the #428 version of this awkward. Good outcome.

Issue #183, opened 2026-07-29
Samsung Range (TP1X_DA-KS-RANGE-0101X|40445041|5001011E031511010200000000000000)


dacmcbibs:
### Appliance model

NE63A6511SSAA

### What's missing or wrong

Here are differences between the smartthings app and the integration:
- The smartthings app can determine if the child lock is on or off, the integration cannot (would be awesome if this worked) 
- The oven has a child lock toggle that does not work, it does not look like the app can toggle the child lock (would love if this worked :) )
- Smartthings has a 120hour energy saving toggle - this is not in the integration
- Smartthings has a Cooktop on alert toggle (this one is pretty pointless for the integration since it triggers alerts in the smartthings app,, just sharing for completeness) 
- Integration has a "fast preheat" setting that doesn't exist in the app - does not appear to do anything
- Integration has a "Natural Steam" setting that doesn't exist in the app - does not appear to do anything
- Integration has a "Power" setting that doesn't exist in the app - does not appear to do anything
- Integration has a "setpoint" that may possibly be "temperature" in the app - not sure though since the value in the integration and app don't match (if it's not temperature, temperature is missing) 
- There doesn't appear to be any way to send cook settings to the oven (the app has a start button) 
- If there was a way to enable smart control that would be... interesting. though also possibly a safety issue (the app intentionally doesn't have this feature) 

[config_entry-localthings-01KYKDGFB39BP64HMHDDH4CE3E.json](https://github.com/user-attachments/files/30487569/config_entry-localthings-01KYKDGFB39BP64HMHDDH4CE3E.json)

mbillow:
Thanks for the detailed writeup and diagnostics — went through it point by point:

**Fixed:**
- **"Fast preheat" and "Natural Steam" doing nothing** — confirmed bug, not just on your unit. Both were shipped as an unverified guess and bound unconditionally, so they always showed "off" and any toggle wrote a token your oven's firmware never had in the first place (your dump has neither `fastpreheat_*` nor `NaturalSteam_*` in its options at all). Both now correctly stay hidden when the device doesn't report them.
- **120-hour energy saving toggle** — your dump does carry `EnergySaving_On`, just wasn't wired up. Added as a real switch.
- **Cooktop-on alert** — also present (`BurnerOnAlert_Off`), added too, even though you noted it's mostly for the app's own alerting.

**Already working, no change needed:**
[...]

PR #457, opened 2026-09-07
Stop the oven's finish_time sensor from changing on every push

Averyy:
## What this changes

The oven's `finish_time` sensor produced one logbook entry every 4 to 7 seconds during a 15-minute bake on an NX60T8311SS (TP2X range), flipping between 7:07 and 7:08 PM. Cause: the range pushes `/operational/state/vs/0` each time `progressPercentage` ticks, and its `remainingTime` only has minute resolution with the seconds parked at `:59`, so `oven.py`'s `now() + remaining` drifted a few seconds per push and crossed the minute boundary back and forth.

`operational.py`'s shared `_finish_time` already rounds to the minute and its descriptor carries `hysteresis=True` for exactly this reason (the 3-minute deadband from the options flow). The oven's older copy had neither. It now:

- rounds to the minute the same way,
- sets `hysteresis=True` on the descriptor,
- reports nothing unless the cook is active, so a board that leaves a stale `remainingTime` after Finish reads no phantom time. This range reports `00:00:00` there, so it doesn't change on this unit; it's the same guard the shared version has.

The descriptor moves from `field`/`value_fn` to `rep_fn` so it can see the state. State keys are unchanged, goldens untouched.

Observed on the same cook, for the record: `remainingTime` ticks by whole minutes (`14:59`, `13:59`, ...), `progressPercentage` climbs about one point per 7-10 s, and `/oven/vs/0` goes `Preheat`, `Cooking`, `Finish`. Full notes on #183.

## Validation

- `tests/test_oven_capabilities.py`: rounded and stable across polls with `hysteresis` set; nothing while Ready with a leftover time, while running with `00:00:00`, with the field missing, or unparseable
- `ruff format --check`, `ruff check`, `ty check`, `pytest tests/ -q`: 1925 passed

### DIFF
--- a/custom_components/localthings/registry/capabilities/oven.py
+++ b/custom_components/localthings/registry/capabilities/oven.py
@@ -85,17 +85,31 @@ def _int(v):
         return None
 
 
-def _finish_time(remaining_str):
-    if not remaining_str:
+def _finish_time(rep):
+    """now() + remainingTime while a cook is running, rounded to the minute.
+
+    Same treatment as operational.py's _finish_time, for the same reason:
+    a range pushes /operational/state/vs/0 every few seconds during a cook
+    (progressPercentage ticks about once per 7-10 s), and its remainingTime
+    only has minute resolution with the seconds parked at ':59', so the
+    unrounded sum drifted by a few seconds per push and crossed the minute
+    boundary back and forth. That was one logbook entry per push (NX60T8311SS,
+    a 15-minute bake). The state gate covers boards that leave a stale
+    remainingTime after the cook ends; this one reports 00:00:00 there."""
+    if _to_ocf(rep.get("x.com.samsung.da.state")) != "active":
+        return None
+    remaining = rep.get("x.com.samsung.da.remainingTime")
+    if not remaining:
         return None
     try:
-        h, m, s = remaining_str.split(":")
+        h, m, s = remaining.split(":")
         total_s = int(h) * 3600 + int(m) * 60 + int(s)
-        if total_s == 0:
-            return None
-        return datetime.now(UTC) + timedelta(seconds=total_s)
-    except Exception:
+    except (AttributeError, ValueError):
         return None
+    if total_s == 0:
+        return None
+    finish = datetime.now(UTC) + timedelta(seconds=total_s)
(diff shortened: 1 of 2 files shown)


mbillow:
First pass, and this is right. Verified locally on `pr457` at fff911d: `ruff format --check`/`ruff check`, `ty check`, **1925 passed**.

The diagnosis matches what the fixtures show. `remainingTime` on every oven-family board is minute-resolution, and the range pushing `/operational/state/vs/0` on each `progressPercentage` tick is exactly the condition `operational.py`'s `_finish_time` comment already describes for washers — the oven's copy predates it and never got the same treatment. Rounding plus `hysteresis=True` is the fix that's already proven in the shared version, and the descriptor is now literally the same shape as `operational.py`'s.

On the state gate, which is the part that changes behavior beyond jitter: I checked all 31 fixtures carrying `/operational/state/vs/0`. Every one reports `x.com.samsung.da.state`, so there's no board where the new gate silently kills a finish time that used to work, and every oven/range/microwave fixture reports `00:00:00` remaining while Ready — so on today's corpus the gate is belt-and-braces rather than a change in output. Ranges that leave a stale time will now stay quiet, which is what we want.

Two differences from the shared version I looked at and am fine with:

- You gate on `state` alone, while `operational._is_active` also excludes `progress == 'Finish'`. No oven-family fixture reports `progress` in that resource at all (the oven's Preheat/Cooking/Finish lives in `/oven/vs/0`), so the extra clause would be dead code here.
- Narrowing `except Exception` to `(AttributeError, ValueError)` is a real improvement and the unparseable case is covered.

Dropping `field=` in favor of `rep_fn` doesn't widen creation — the goldens are unchanged and every board with the resource reports `remainingTime`.

One style nit for next time, not a change request: the docstring re-derives reasoning `operational.py` already documents, and the test class docstring says it a third time. `CONTRIBUTING.md` asks for a pointer rather than a re-derivation — `"same rounding + hysteresis as operational._finish_time, for the same reason; observed on NX60T8311SS during a 15-minute bake (#183)"` carries the whole payload. The investigation notes on #183 are the right home for the rest, and you already put them there.

Merging.

Issue #300, opened 2026-08-05
Samsung TP2X_DA-KS-WALLOVEN-000002 not responding to commands


inventor7777:
### Appliance model

TP2X_DA-KS-WALLOVEN-000002

### What's missing or wrong

I enabled Smart Control and the icon on the oven is illuminated, and the HA LocalThings binary sensor for that turns on, but it doesn't respond to commands even when I enable the always-write mode.

In addition, I got the notification from Home Assistant that it is not fully supported.

It does properly create a subdevice for the second cavity, which I really like!

### Diagnostics file

[config_entry-localthings-01KZ94BDN9CS8CFAQ8A979XXDA.json](https://github.com/user-attachments/files/30748045/config_entry-localthings-01KZ94BDN9CS8CFAQ8A979XXDA.json)

mbillow:
Two things going on here, and I've fixed one:

**"Not fully supported" notification** — the one unbound href on your dump, `/diagnosis/vs/0`, is now covered (same shape dishwashers already report). That repair should clear in the next release.

**Doesn't respond to commands** — this one I can't fix blind, but I did find and fix one real bug: your board's `/mode/vs/0` never reports an `UpperLamp_` token at all (yours is a steam-oven-class board — water tank, descale, pyro-free — quite different from the board the oven lamp switch was originally proven against), so the `lamp` switch was showing up and always reading Off with no way for a write to it to ever do anything. That switch will simply stop appearing after the fix, rather than pretending to work.

For the rest (cook mode, setpoint, cook time, start/stop) — our write contracts for this board generation are explicitly unproven in the code (never confirmed against real hardware), so I'd like your help narrowing down whether the problem is on our side or the oven's. Could you try the following through **Settings → Devices & Services → LocalThings → Configure → Debug write**, one at a time, and tell me for each: the response code shown, whether the "new value" readback matches what you asked for, and whether the physical oven's display actually changes?

1. Sanity check (this one *should* work — sound[/`beep`] is a proven on/off token on your board):
   `href: /mode/vs/0`
   `payload: {"x.com.samsung.da.options": ["Sound_Off"]}`
   (then again with `Sound_On` to restore it)

2. Temperature setpoint — this needs the *whole* items array, not just one field:
   `href: /temperatures/vs/0`
[...]

PR #466, opened 2026-09-11
Translate the legacy 8888 envelope into canonical resources

perseus177:
Stage 1a of what you asked for in #168, as code rather than prose: the pure envelope translation and its tests, and nothing else. No transport, no config-flow branch, nothing imports it at runtime. Draft on purpose — it is the half that can be reviewed without any of the coordinator churn a transport would need, so "not this shape" costs you one comment.

*Written by Claude Code acting for @perseus177, who owns the appliance and read this before it went out. The field names and values are off that hardware.*

## The part you asked about, in a form you can check

> If you want to propose a design for an isolated shim that could leverage our existing device registry (I could be convinced to add a field or two) […] I simply don't have enough data to understand how complex or flexible that interface/translation layer needs to be.

**The registry needs no new fields.** The whole translation sits below `adapter.py`. Stage 0 in the issue is the evidence: the shimmed dump resolved to `washer`, bound 16 entities and left `unbound_hrefs` empty with `registry/` untouched.

**And the interface is countable.** Across the component the transport is reached in five files, through two verbs:

| file | what it calls |
|---|---|
| `coordinator.py` | `sess.get` ×6, `sess.post` ×2, `pace` ×3, construct/connect/close, `cbor2.loads` ×6 |
| `config_flow.py` | construct/connect, plus `_read_device`'s one `/device/0` GET |
| `registry/identity.py` | `read_identity`'s three GETs |
| `registry/subdevices.py` | one GET helper + `pace` |
| `observe.py` | one `cbor2.loads`, on the notification path only — HTTP never reaches it |

That seam is *not* in this PR. This is only what sits behind it.

## What is here

`legacy_http.py` — a table and four pure functions. Two mechanical rules carry the translation: a field takes the `x.com.samsung.da.` prefix and keeps its name; a resource maps onto the canonical href its fields belong in. Everything not mechanical is a table row, so an appliance that disagrees costs a row rather than a branch.

The one measured family needs three such rows:

| 8888 resource | canonical href | exception |
|---|---|---|
| `Operation` | `/operational/state/vs/0` | `power` → `/power/vs/0`, `kidsLock` → `/kidslock/vs/0` |
| `Mode` | **`/course/vs/0`** | — (note: *not* `/mode/vs/0`, which this firmware 404s) |
| `Washer` | `/washer/vs/0` | — |
| `Configuration` | `/remotectrl/vs/0` | — |
| `Information` | `/information/vs/0` | `modelID`→`modelNum`, `serialNumber`→`serialNum` |
| `Diagnosis` | `/diagnosis/vs/0` | — |
| `Alarms` | `/alarms/vs/0` | bare JSON list → an `items` array of prefixed maps |

Everything else came out mechanical against the registry as it ships today — `state`, `progress`, `progressPercentage`, `remainingTime`, `waterTemperature`, `rinseCycles`, `spinLevel`, `supported*`, `options`, `supportedOptions`, `kidsLock`, `remoteControlEnabled`, `diagnosisStart` are all the same names your capabilities already read, one prefix apart.

`to_write` takes a **list** of steps and coalesces them into one aggregate body, because on this firmware a cycle is accepted only in the same body as `Operation.state` — a `Course_` token alone is answered `204` and discarded in every shape tried, including a two-token body alongside `LaundryOutTime` that applied the other token in the same call. That is a firmware rule about that field, not a syntax problem, and it isn't exotic to this transport either: `async_raw_write_sequence` exists here (#300) because a wall oven discards settings writes while idle and keeps them only once a cycle is running. Same fact, different family, over CoAP.

`http_status_to_coap` keeps one vocabulary above the seam, so `_coap_accepted`, `_coap_code_str`, diagnostics and the debug services work unchanged. `204 → 2.04` is exact in the unwelcome sense too: accepted is not applied on either transport (an ARTIK051 AC answers `2.04` to `FilterTime_0`, echoes the token, and has dropped it a poll later), which is why the read-back in `_raw_write_blocking` is the right posture for both.

## What is deliberately not here

- **A device-dump fixture.** This appliance leaves the network minutes after it goes idle unless Remote Control is on, and it is off right now, so the binding assertion — run the dump through `by_type.resolve()` and `discover()`, assert the 16 entities and the empty `unbound_hrefs` — is a second commit when it is next running. The test bodies in this PR are hand-built from measured field names and values and say so in the module docstring; they pin the shape of the translation, and nothing in them stands in for a dump.
- **The transport, the config flow, entities.** Staging is in #168. The seam is the part that touches your code, and I would rather implement the one you want than argue for mine.
- **Anything about credentials.** The line you drew doesn't move: the user supplies the CA, and on this family a device token as well; nothing ships either.

If you would rather this didn't ship inside `custom_components/` until something imports it, say so and I will move it under `tests/` as a harness — the tests are the point either way. And if the answer is that the family stays out of scope, that is a fine answer too: the standalone integration works, and closing this costs nothing.

### DIFF
--- a/custom_components/localthings/legacy_http.py
+++ b/custom_components/localthings/legacy_http.py
@@ -1,0 +1,265 @@
+"""Envelope translation for the legacy 8888/HTTPS appliance family (issue #168).
+
+Some 2018-2022 Samsung appliances have nothing on UDP 49152-49160 and only
+TCP 8888: nginx over TLS 1.0 with a mandatory client certificate, serving a
+small REST bridge instead of an OCF server. The OCF URL space is not merely
+unauthorised there, it is absent -- ``/oic/res``, ``/oic/d`` and
+``/mode/vs/0`` all come back as nginx's own HTML 404, while an unknown path
+*inside* the served space answers ``{"errorCode": "0", "errorDescription":
+"Invalid resource"}`` from the application.
+
+What that bridge serves, though, is the same vocabulary in a different
+envelope: the same ``options``/``supportedOptions`` arrays, the same
+``Course_5C`` tokens, the same single-token merge semantics on writes. Only
+the spelling differs -- bare field names instead of ``x.com.samsung.da.*``,
+and ``/devices/0/<resource>`` instead of ``/<resource>/vs/0``.
+
+So this module is a table and four pure functions over it. Given one of
+those appliances' responses it produces exactly the ``{href: rep}`` shape
+``registry.batch.parse_device0_batch`` produces, which is what lets
+``registry/``, ``discovery.py`` and ``adapter.py`` work on such a device
+unchanged -- the expensive part, and the part this repository has already
+solved.
+
+Deliberately pure: no I/O, no Home Assistant, no session. Nothing imports it
+at runtime yet; it is the half of issue #168's proposal that can be reviewed
+and tested without a transport existing, and without any of the coordinator
+churn a transport would need. See that issue for the staging.
+
+Two rules carry the whole translation, and both are mechanical:
+
+* a resource's fields take the ``x.com.samsung.da.`` prefix and keep their
+  names;
+* a resource maps onto the canonical href whose rep those fields belong in.
+
+Everything that is *not* mechanical lives in the table as data, so an
+appliance that disagrees costs a row rather than a branch. On the one
+family measured so far that is three exceptions: ``Operation`` carries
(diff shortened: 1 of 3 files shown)


mbillow:
Closing in favor of #499. The envelope translation from this PR is in there as its own commit, with your authorship, and #499 builds the rest of the 8888 support on top of it.

---

PR #467, opened 2026-09-11
Support the legacy 8888/HTTPS appliance family

perseus177:
The 8888/HTTPS family, working, in six commits — issue #168's proposal built out rather than argued for. It runs a real TP6X_WW6500 through this integration's own registry and comes out with 26 entities; the transcript is at the bottom.

Draft, and yours to throw away. You asked how complex the translation layer would have to be; this is the answer with the code attached, so the size of it is a fact rather than my estimate. #466 is the first commit of it on its own, if a small slice is the more useful thing to look at.

*Written by Claude Code acting for @perseus177, who owns the appliance and read this before it went out. Every number here is off that hardware.*

## What it does

A user types an IP. If TCP 8888 answers, the flow takes the legacy branch — same CA, same leaf, plus the one credential the CoAP path never needs — and the appliance shows up with the entities its family already has in this repository. Nothing in `registry/`, `discovery.py` or `adapter.py` changed, and no new entity description was written.

## The six commits

| | |
|---|---|
| **Put the device connection behind one interface** | `Transport`: connect/close/pace/read/write plus the two OBSERVE calls, with `DtlsTransport` as the only implementation and no behaviour change. Decoding moves below it, so `read` returns a decoded body and `cbor2` disappears from the coordinator entirely. |
| **Translate the legacy 8888 envelope into canonical resources** | The pure table and four functions (this is #466). |
| **Check the translation against the appliance it came from** | The device dump as a fixture, run through `resolve()` and `discover()`. |
| **Assert the registry resolved before reaching into it** | `ty` was right: `resolve()` is Optional. |
| **Speak the 8888 bridge as a second transport** | The HTTP transport itself. |
| **Set up an appliance that has no DTLS to probe** | The config-flow branch and the device-token exchange. |

## What the seam cost

Five files reached a device and each knew it was CoAP over DTLS: `coordinator.py` (6 × `sess.get`, 2 × `sess.post`, 3 × `pace`, plus 6 × `cbor2.loads`), the config flow's probe, `registry.identity`, `registry.subdevices`, and `observe.py`'s notification path. The first four now go through the interface; **`observe.py` needed nothing at all** — `ObserveRefreshTask` only ever calls `refresh_observes`, which the transport forwards.

The visible cost is in the tests: the doubles stood in for `DtlsCoapSession` and now stand in for a `Transport`, which makes most of them shorter, since a fake no longer encodes CBOR to be read back. `tests/test_transport.py` covers `DtlsTransport` itself, which was only ever exercised incidentally before.

**The registry gained no fields.** That was the question in #168 and the answer held up all the way to a running appliance.

## Three things about the firmware that are in the code

* **The response is not valid HTTP** — `X-API-Version : v1.0.0`, space before the colon. `aiohttp` rejects the whole response over it, so this is built on `http.client`, which suits a seam that is blocking anyway.
* **There is no `/device/0`.** The seed read assembles one: the aggregate carries five resources, and only the two it links to cost a request of their own. Three requests per sweep.
* **A cycle is accepted only as part of a start.** `Course_` alone is answered `204` and discarded — measured, including in a body whose *other* token applied, so it is the field and not the request. Same for the washer's temperature/rinses/spin. Those writes are **refused with a log line saying why** rather than sent, because sending them looks like a write that worked. Making them work needs a composite write (`write_fn` returning steps, which `async_raw_write_sequence` already models for the oven that discards settings while idle) — deliberately not in this PR.

## A correction to something I told you earlier in #168

I wrote that this appliance accepts only a SHA-1 leaf, and flagged it as worth one re-test before anyone built on it. I re-tested it: **it is wrong.** A SHA-256 leaf minted exactly the way `_mint_leaf_cert` mints one, same UUID, was answered `HTTP 200` by the appliance, with the known-good SHA-1 leaf as the control in the same run. `_mint_leaf_cert` needs no digest argument and this PR does not touch it. Sorry — that came from following a working reference implementation rather than from measuring.

## Evidence

`ruff format`, `ruff check`, `ty check` clean; **2060 tests pass**.

Set up in a real Home Assistant against the live appliance:

```
Setting up localthings ... 26 entities

select.…_cycle                 5B        <- the dial's actual position
select.…_wash_temperature      60
select.…_spin_speed            1400
select.…_rinse_cycles          2
sensor.…_machine_state         idle
sensor.…_progress              idle
sensor.…_completion_time       269       <- the appliance's own 04:29:00
sensor.…_alarm_code            none
sensor.…_diagnosis_status      ready
binary_sensor.…_smart_control  on
binary_sensor.…_child_lock     on
binary_sensor.…_running        off
switch.…_power                 on
switch.…_addwash_alarm (+ rinse, final_rinse, spin)
button.…_start / _pause / _stop / _start_diagnosis
```

`discover()` leaves **nothing unbound**, and `cycle_options()` decodes the 14 courses out of `supportedOptions` through your own decoder, on a board that publishes no `editCourseList`.

## What is not here

* **No composite write**, so the cycle and settings selects are read-only in effect (they refuse, loudly).
* **One family in the table.** A second either fits the mechanical rule or costs a row; a dryer's `description` would be recorded on the entry either way, so a diagnostics dump from one says which.
* **The credential boundary does not move.** The user supplies the CA, and on this family a device token as well; nothing ships either, and the token exchange only ever runs against the user's own appliance.
* One **`stop` button caveat** worth knowing regardless of this PR: it writes `Operation.state = "Ready"`, which on this firmware cancels from `Run`, does nothing from `Pause`, and on an *idle* appliance moves it to `Pause` and resets the hand-dialled settings.

If the answer is that this family stays out of scope, that is a fine answer and costs you one comment — the standalone integration works and nothing here is wasted on my side.


QuiteYellow:
Your SHA-1 correction matches what I measured on different hardware. I made a mistake here and corrected it in my own repo today, for the two appliances I own at least.

They are a dryer and an oven, both AC14K_M-generation, on CoAP/DTLS. A one-variable-at-a-time factorial with a random-UUID control puts the gate on one field: the leaf's subject UUID, matched against the on-device ACL. Signer, chain, digest and public key all ride along unchecked. Changing only the digest, as you did, an AC14K_M-signed SHA-256 leaf is accepted on both. A leaf signed by a throwaway CA the appliance has never seen reads every data resource; one carrying a random un-ACL'd UUID handshakes and is refused on all of them. Evidence is in QuiteYellow/SmartThings-Local#95.

Different family, different transport, so it may not transfer. But the digest is one of those four variables and you have already run it. If a self-signed leaf carrying the right UUID also gets a 200 on 8888, then the CA the user supplies is doing nothing on that family either, and the same control run would say so.

mbillow:
@perseus177 Can you try a self-signed certificate built to QuiteYellow/SmartThings-Local#95’s spec against your device? You can also look at how it’s implemented on the main branch here as well, since we’re adopting the same pattern @QuiteYellow found as a default, first try.

perseus177:
Ran it. A self-signed leaf works on this appliance — but on this family the UUID
is not the gate either, and the control that should have failed didn't.

*Written by Claude Code acting for @perseus177, who owns the appliance and read this
before it went out. Every number here is off that hardware (`TP6X_WW6500`, 8888/HTTPS,
TLS 1.0, `GET /devices/0`).*

I minted the leaves with your own minters copied verbatim out of `main`
(`_uuid_subject_name` + `_mint_self_signed`), so what was tested is upstream's
default rather than my paraphrase of it. B2 is #95 read strictly instead —
subject and SAN carrying only the UUID, standard EKU, no organisation.

| | client certificate | token | result |
|---|---|---|---|
| A | AC14K_M SHA-1 (my known-good) | valid | **200** (15/15) |
| B1 | **self-signed, your `_mint_self_signed`** | valid | **200** (15/15) |
| B2 | self-signed, #95 strict (SAN+EKU, no `O`/`C`) | valid | **200** (8/8) |
| C | self-signed, **random un-ACL'd UUID** | valid | **200** (15/15) |
| G | self-signed, `CN=totally-unrelated-cert`, **no `uuid:` RDN at all** | valid | **200** (15/15) |
| E | self-signed | wrong | **401** `Token is not valid` |
| F | self-signed | none | **401** `Token is not valid` |
| H | AC14K_M known-good | wrong | **401** `Token is not valid` |
| D | **none** | valid | **400** `No required SSL certificate was sent` (nginx) |

So for the 8888/HTTPS family:

1. **Self-signed authenticates.** Your default works here, no CA needed.
2. **The subject UUID is not the gate.** C is the control that distinguishes your
   families from mine: on your dishwasher and fridge a wrong-UUID leaf handshakes and
   gets `4.01`. Here it reads everything. So does G, which has nothing Samsung about it
   anywhere in the subject. Signer, chain, digest and UUID are all unchecked.
3. **The device token is what authorizes.** Same cert, wrong or absent token → `401`,
   and the original AC14K_M leaf gets `401` just the same.
4. **nginx still wants *a* certificate to be present** — with none it's `400`. It behaves
   like `ssl_verify_client optional_no_ca`: the cert must exist, it just isn't validated.

Which means #467 can drop the CA from the legacy branch entirely and mint self-signed,
same as `main` now does — the CA that branch currently asks the user for is doing nothing
on this hardware. I'll push that unless you'd rather it stayed as-is. It also happens to
take the AC14K_M material out of this family's setup path, which I know you'd prefer.

**One caveat, and I'd rather flag it than let it bite someone.** Two of my early runs
returned `400 The SSL certificate error` — five times inside about two minutes — and it
reads exactly like a cert rejection. It wasn't. The same files then passed 138/138 across
four later runs. I couldn't isolate it: not cert content, not `notBefore` (six leaves from
"now" to −30 days, all 200), not first presentation of an unseen subject (20/20), not
request pacing (60 back-to-back, zero failures). It looks like the same
credential-independent intermittency you named when you retracted the "self-signed doesn't
authenticate" belief — it just surfaces as an nginx 400 here instead of a missing first
flight. My first three runs, read one request at a time, would each have supported a
different wrong conclusion. Worth knowing for anyone testing certs on 8888.

perseus177:
Pushed. The branch is rebased onto `main` and carries the self-signed default I
offered above, plus two things the appliance itself taught me once it was running
in a real Home Assistant.

*Written by Claude Code acting for @perseus177, who owns the appliance and read this
before it went out. Every measurement below is off that hardware.*

**Rebase, and a force-push.** The branch was six commits behind 28 of yours,
including #479. It is now on top of `main`, which is why the first push rewrote it;
everything since is a fast-forward. The merge wasn't free: `Transport.write`
returned a bare code, so the seam was dropping the `response_body` your raw-write
service had meanwhile started returning. On the DTLS boards that is usually empty.
On this one it is the appliance's own account of a refusal (`"Control fail, <...>"`),
which is the only account it ever gives — so `write` now returns `(code, body)`, the
same pair `read` does. `ty` caught the legacy transport still returning an int after
the DTLS one was widened.

**No CA on this branch any more.** `_mint_preferred` picks the AC14K_M CA when an
entry already has one and mints self-signed otherwise, shared with your probe so
neither path decides it alone. There is no `fallback_ca` here: as measured, this
family authenticates nothing about the certificate, so asking for one would be
asking for material that provably does nothing.

## Cycle names: this family is `Table_00`

`cycle_select` reads its course-table id from `/st/washercourse/vs/0`, which this
family does not serve, so every cycle was labelled with its raw code. The id is
declared per family now (a row, not a branch) and derived from nothing — which
codes a board uses is not something its model string states.

The evidence that it is `Table_00` is better than I expected. The 14 codes were
mapped by turning the dial through all 14 panel positions and reading `Course_`
back at each, five anchored semantically (0x5F the only 95C-and-4-rinse program,
0x5E the only 400rpm, 0x5D the only cold, 0x64 the only one with no temperature,
0x63 60C at 400rpm). That map agrees with your existing `washer_cycle_table_00`
catalog on **13 of 14 codes exactly**, including every code the dial walk had
placed only positionally. The fourteenth, 0x6C, was absent from the catalog and is
Denim by elimination; it is added here, reusing the Denim string each locale
already carries for Table_02.

**`supports_observe` was read by nobody.** Both transports declared it and nothing
consulted it, so the subscribe burst ran against the 8888 bridge every cycle and
each href logged a warning for a capability the transport never claimed — four
every ten minutes against a live appliance. `_attempt_observe_mode` now checks it.

## Starting a chosen cycle, and what the appliance did

`Course_` cannot be sent alone here: it is answered `204` and then discarded. So
`Transport.write_many` expresses one action spanning several resources — ordered
writes on DTLS, one `PUT /devices/0` on 8888, which is what `legacy_http.to_write`
was built to produce. `localthings.start_cycle` is the caller.

Then the hardware disagreed with the design. `{Mode.options: [Course_63],
Operation.state: Run}` **loads the programme and does not run it**: the dial moved
to Drum Clean, remaining time went 59 → 119 minutes, and the appliance sat in
`Ready`. A plain `Operation.state = Run` sent afterwards started it — `active`,
progress `wash`. That is the same write your Start button sends, so the service
issues it, and only when the device itself reports it is not running. Never blind.

Two corrections to things I believed before running it:

* I had thought a `Washer` settings block in the start body was what dropped the
  `Run`. This body carried no such block and the `Run` was dropped just the same.
* A `204` from this firmware guarantees nothing. The composite returned 2.04
  Changed while silently discarding half of what it was asked to do. The reliable
  signal that a programme loaded is the completion time jumping to that
  programme's own length.

**On your `stop` button**, which I flagged earlier as writing `Operation.state =
"Ready"`: from `Run` it cancels, as expected. I had also recorded that it does
*not* cancel from `Pause` on this board — that is wrong, or at least not general.
It cancelled from `Pause` twice out of two today, back to `Ready` with the
programme still selected and the clock reset to its full length.

## Evidence

`ruff format`, `ruff check`, `ty check` clean; **2137 tests pass**.

Live, in a real Home Assistant against the appliance: 26 entities, none
unavailable, cycle/temperature/rinses/spin matching the panel; `start_cycle` chose
Drum Clean and started it unaided; Pause paused it; Stop cancelled it. The
self-signed *first-run* path is the one thing that instance could not exercise —
it already had entries, so the flow reused their leaf, as it should.

mbillow:
Heads up that #487 just landed a rework of config-flow discovery, and it moves ground this PR stands on. Nothing here is a request to change the design — the transport seam is untouched and still the interesting part. This is so the rebase isn't a surprise, plus one thing that becomes a real bug the moment this PR works.

## What moved

Everything the config flow used to do before authenticating now lives in a new `custom_components/localthings/probing.py`. The symbols this PR's diff touches went with it:

| was, in `config_flow.py` | now, in `probing.py` |
| --- | --- |
| `_scan_ports` | `look` |
| `_PortScan` | `HostProbe` (gains `host`, `advertised`, `plaintext`, `plaintext_port`, `legacy_http`) |
| `_SweepResult` / `_find_live_ports` / `_sweep_ports` / `_order_candidates` | `SweepResult` / `find_live_ports` / `sweep_ports` / `order_candidates` |
| `_clienthello_probe` | gone — replaced by `probe_dtls_ports` |

`_probe_and_validate` now opens with `scan = probing.look(host)`, and `_handshake_and_read` / `_diagnose_failures` / `_classify_handshake_failure` take a `probing.HostProbe`. The library floor is `>=0.1.18`, because port selection now happens on the port a reply arrived *from* rather than the port dialled (#486).

`_read_device(sess, host, port)` is unchanged, so the `Transport` rename in commit 1 still applies as written.

## Your 8888 detection is already in

`LEGACY_HTTP_PORT = 8888` and a `_legacy_http_open` equivalent shipped with #487 — same one-TCP-connect approach, same reasoning, credited to your measurement that the two lineages are mutually exclusive. It runs concurrently with the plaintext CoAP lookup, so it costs nothing.

**Delete yours and read `HostProbe.legacy_http` instead.** The detection has already happened by the time `_probe_and_validate` returns.

## The branch point is waiting for you

#487 added a placeholder at exactly the spot this PR needs, and built it to be deleted:

```python
# in _probe_and_validate, before any credential work
scan = probing.look(host)
if scan.legacy_http and not scan.candidates:
    raise LegacyHttpFamily(f"{host} serves the legacy bridge on TCP 8888")

# in async_step_user
except LegacyHttpFamily:
    return await self.async_step_legacy_unsupported()
```

`async_step_legacy_unsupported` is a four-line abort that says "support is in progress, see #467". Replacing it with `async_step_legacy_token` is the whole integration: swap the step, delete the `legacy_http_unsupported` key from the translation files, keep the exception and the raise site.

Two details that happen to suit you. `LegacyHttpFamily` is raised **before** minting, so your flow still controls when the leaf is minted — which you need, since the token exchange uses it. And it deliberately does *not* subclass `CannotConnect`, so it never lands in the retryable-form-error path.

## Caveats

**1. A legacy board comes back with no port data.** When `legacy_http` is true and nothing answered CoAP, `look()` short-circuits and returns `candidates=[]`, `confirmed=[]`, `swept=None` — it skips the DTLS scan entirely, since scanning a board with no CoAP server is a guaranteed-empty wait. Don't expect port information on that path.

**2. CoAP wins if both answer.** The short-circuit requires `legacy_http and plaintext_port is None and not advertised`. Belt-and-braces given your mutual-exclusivity evidence — but if any 8888 board turns out to answer plaintext CoAP on 5683 or 49153, it will take the DTLS path and never reach your step. Worth a check against your TP6X if it's cheap.

**3. This one is yours to fix, and it's the reason I'm writing.** `async_step_reconfigure` catches only `(CannotConnect, InvalidCA)`, so `LegacyHttpFamily` falls through to its blanket `except Exception` and the user gets "Unexpected error, check the logs" plus a traceback. Harmless today — reconfiguring an entry onto an unsupported device fails either way, so it's only a bad message. **Once this PR lands it stops being harmless**: reconfiguring a legacy entry onto a new address becomes a legitimate operation that would hit the same blanket handler. It's parked in #487 as known, precisely because this PR is the natural place to close it.

**4. Eight translation files now, not seven.** `sk.json` arrived in #483 while #487 was in flight — easy to miss when adding `legacy_token`'s strings.

**5. Forms now pass defaulted placeholders.** `async_step_user` and `async_step_fallback_ca` always send `model`/`port` defaults, because HA renders `Translation error: …` when a referenced variable is missing rather than leaving the literal. If your token step's error strings carry placeholders, they want the same treatment — #487 shipped that bug on the reconfigure form and had to fix it.

## On the shape of this PR

For what it's worth, nothing in #487 argued against the seam. The interface held up under a fairly invasive change to the layer beneath it: the whole discovery path was rewritten, `_read_device` didn't move, and `observe.py` still needed nothing. That was your claim in the write-up and it survived contact.

mbillow:
Closing in favor of #499, which carries this branch forward.

#499 keeps every commit from this PR with your authorship, merges current `main` into it, and adds fixes on top:
- `DtlsTransport` decodes only success responses, so an accepted write is never sent twice.
- An unmapped family sets up as an unrecognized device instead of borrowing the washer's table.
- The cycle and settings selects hold the choice, and the Start button sends it, in place of the `start_cycle` service.
- A rejected token raises reauth.
- The token exchange runs as a progress step with a sturdier listener.

Thank you for the measurements and the transport work. None of this would exist without them. #499 lists what still needs a live check on the WW6500, and your help there would be very welcome.

---

Issue #465, opened 2026-09-09
Add support for supportedOptions per washer course


wilbiev:
### Appliance model

DA_WM_TP1_21_COMMON

### What's missing or wrong

First of all, thanks for building and maintaining localthings! With the upcoming SmartThings Cloud API changes, moving to local CoAP-DTLS control is essential, and this integration works great.

Currently, washer entities in localthings expose static global option lists (e.g., all water temperatures from Cold to 90°C, all spin speeds from 0 to 1400 RPM). However, Samsung washers restrict available parameters depending on the active cycle (e.g., Eco 40-60 restricts high temperatures and specific spin speeds per energy regulations).

Inspired by the work done for dryers in PR #442, we’ve been looking into implementing dynamic option filtering for washers as well.

Proposed Implementation Overview
1.	Bitmask Matrix Decoding (/course/vs/0): Parse the 20-character hex blocks in x.com.samsung.da.supportedOptions to decode per-course masks:
[Course Hex (2)] [Soil Mask (2)] [Rinse Mask (2)] [Spin Mask (2)] [Temp Mask (2)] [Dry Mask (2)] [Feature Flags (8)]
2.	Dynamic Entity Option Resolution: Wrap the options getters for washer settings (Temperature, Spin Speed, Rinse Cycles, Soil Level, and Dry Level) to resolve against the active course in /st/washercourse/vs/0.
3.	Feature Availability Flags: Use the available property for PreWash, Intensive Wash, and BubbleSoak switches so they automatically gray out in the UI when unsupported by the active cycle.
4.	Listener Invalidation: Invoke coordinator.async_update_listeners() when changing courses to instantly refresh child entity dropdowns and availability states in Home Assistant.

We have analyzed this against real CoAP data dumps from a Samsung washer (WD7000B / Table_02 course matrix) and are preparing the implementation and unit tests.

Before opening a PR, we wanted to check if this is a feature you'd be interested in incorporating into localthings and whether this proposed architectural approach aligns with your vision for the project.

Looking forward to your thoughts!

mbillow:
Yah, of course! Just make sure you test your hypotheses about how to decode option against all of the washer diagnostics in the repo (the golden regressions test corpus) and maybe even a few issues tagged `device:washer` that have different diagnostics attached. This way we can make sure your work helps all users.

wilbiev:
Thanks for your reply. I will run the tests against all washer diagnostics before I create the PR.

wilbiev:
[...]

PR #468, opened 2026-09-12
Narrow the dryer's dry time to the selected course

vmonkey:
First, an apology: I went quiet on #407 and #408 for the last two weeks. Work got very busy and I simply dropped the thread — sorry. Thank you for carrying the review fixes into #425 and the catalog into #442 rather than sitting waiting on me for a push. Picking the agreed sequence back up here.

## Summary

`dry_time` offered every duration its board supports on every course. A write outside what the running course accepts is silently ignored by the appliance — no error, no state change, nothing the user can see. This narrows it with `course_narrowed_options`, the factory #442 was written as so the follow-ups could reuse it.

No new mechanism was needed: the dry-level/dry-time exclusivity falls out of the existing factory.

## Naming `0xE`

Named here rather than in #425, per that PR's own framing — name it in the change that consumes it.

The evidence is the **complement**, not "it decodes against `supportedDryTime`", which on its own is weak since a six-entry list absorbs most masks without complaint. On the DV6800N, the only dump carrying both, every one of the fourteen records holds *both* a `0xD` and a `0xE` group, and the exclusion is carried by an empty mask rather than a missing group:

| courses | `0xD` | `0xE` |
| --- | --- | --- |
| the ten level-dried ones (Cotton, Wool, Iron Dry, Bedding, Delicates, …) | values | `E000` |
| the three timed ones (Cool Air, Warm Air, Time Dry) | `D000` | values |
| Quick Dry 35 | `D000` | `E000` |

A kind unrelated to the dry dial would not go quiet exactly where the dry dial speaks and speak exactly where it goes quiet, across fourteen courses. The one course offering neither is Quick Dry, which takes no dry setting at all — so the complement is total wherever there is anything to complement.

Corroborated the same way the WW6500's record shape was: this dump's `/course/vs/0` options carry a `MostUsed_9AD20EE000` token, byte-identical to course `9A`'s record, which pins the 5-byte width and both groups independently of the header.

A second board carrying `0xE` would settle it outright. Until then this is one board's structure, and the write-up says so rather than claiming more.

## Effect

On the DV6800N, Cotton gives `dry_level = ["1", "2", "3"]` and `dry_time = ["00:00:00"]`; Time Dry inverts it. The dial the selected course does not use collapses to its live value instead of offering settings the appliance will ignore.

Everywhere else nothing changes. The four boards reporting a `supportedDryTime` with no `0xE` group anywhere decode to "no opinion" and keep their full list, and `dryer_dve50a8600` (no `supportedDryTime` at all) stays suppressed by its `exists_fn`.

> [!IMPORTANT]
> Same behaviour change as #442: an automation that was quietly no-opping on a duration the running course refuses will now raise a visible `ServiceValidationError` instead. That is the win, but users will notice it.

## A count correction

The module comment and the investigation doc both said three boards carry an 11- or 13-entry `supportedDryTime` with no group for it. It is four — `dryer_dv80h`'s 13-entry list landed after that count was written, and the doc's enumeration never picked it up. Corrected in both, and the new test loops all four rather than asserting on one, so the claim is pinned rather than restated. If a later dump ever pairs a long list with a `0xE` group, that test is where it surfaces.

## Validation

- 2005 tests pass; `ruff check`, `ruff format --check` and `ty check custom_components tests` all clean
- No golden changes — goldens carry `state_keys`, and narrowing only touches `options`
- Both new tests confirmed to fail when the descriptor wiring is reverted, and to pass when it is restored
- `test_dv6800n_dry_policy_agrees_with_the_other_board_family` extended to pin the complement, and that both groups are structurally present on every record — the exclusion has to stay an empty-mask fact, since that is what lets the factory collapse the unused dial rather than read it as "no opinion"

## Not included

The washer's spin/rinse/temperature, which is the rest of the sequence we agreed. #465 looks to be covering that ground already, so I have left it alone rather than duplicate the work — happy to help there instead if it is useful.

### DIFF
--- a/custom_components/localthings/registry/capabilities/dryer.py
+++ b/custom_components/localthings/registry/capabilities/dryer.py
@@ -14,6 +14,7 @@ from ..entities import SelectDesc, SensorDesc, SwitchDesc
 from .common import diagnosis_status
 from .laundry import (
     OPTION_KIND_DRY,
+    OPTION_KIND_DRY_TIME,
     course_narrowed_options,
     cycle_select,
     drum_clean_cycles_remaining,
@@ -74,12 +75,22 @@ DRYER_SETTINGS = Capability(
             ),
             write_fn=_setting_write("x.com.samsung.da.dryLevel"),
         ),
+        # Narrowed on 0xE, which pairs with dry_level's 0xD rather than
+        # duplicating it: on the DV6800N no course carries values for both,
+        # so a dry-level course collapses this dropdown to its live value and
+        # a timed course collapses dry_level's. The four boards reporting
+        # supportedDryTime without a 0xE group decode to "no opinion" and
+        # keep the full list.
         SelectDesc(
             key="dry_time",
             field="x.com.samsung.da.dryTime",
             icon="mdi:timer",
             entity_category="config",
-            options_field="x.com.samsung.da.supportedDryTime",
+            options=course_narrowed_options(
+                OPTION_KIND_DRY_TIME,
+                "x.com.samsung.da.dryTime",
+                "x.com.samsung.da.supportedDryTime",
+            ),
             exists_fn=lambda rep, resources: bool(rep.get("x.com.samsung.da.supportedDryTime")),
             write_fn=_setting_write("x.com.samsung.da.dryTime"),
         ),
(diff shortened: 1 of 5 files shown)


mbillow:
No need to apologize! Gave this a look over and only really had nit comments. I’ll spare you my pedantry and just merge as is. Thanks for the contribution!

Issue #469, opened 2026-09-13
Feature request: support hostname/mDNS resolution instead of requiring a static IP


mbapepp90:
My situation

I have 3 Samsung air conditioners set up through LocalThings on Home Assistant. My router (AVM FRITZ!Box) has a DHCP static-lease reservation configured for each unit's MAC address ("always assign the same IPv4 address"), yet the IP has changed many times for all three units simultaneously, despite the reservation being correctly in place and the MAC addresses staying identical before and after (confirmed by checking the router's device list each time).

Each time this happens, all three LocalThings config entries stop working — HA can't reach the appliances at their old IP — and the only fix is going into the router's device list, finding the current IP per MAC, and manually updating (or removing/re-adding) each affected integration entry. It's a repetitive, fully manual chore every time it happens.

Root cause on my end still isn't 100% pinned down (possibly router-side Wi-Fi channel/bandwidth auto-adjustment on the mesh node these units connect through, possibly a FRITZ!OS bug with DHCP reservations behind mesh repeaters), but regardless of the underlying router quirk, the integration itself has no resilience to it at all — a changed IP means a full manual teardown/rebuild of the device.

General problem

Currently, adding a device requires entering its IP address, and the config entry stores that IP permanently. If the appliance's IP changes on the LAN (DHCP lease change, router reassigning addresses, etc.), the integration silently stops working — polling/writes fail until the device is manually removed and re-added with the new IP.

This is especially painful for appliances that need CA credentials: removing and re-adding means going through the CA certificate/key setup again (or at least re-supplying the IP and going through the leaf cert minting flow), even though the appliance itself hasn't changed.

Suggested improvement

Would it be feasible for the config flow / coordinator to support resolving the appliance by something more stable than a raw IP — for example:

Resolving by mDNS/hostname if the appliance advertises one on the LAN, instead of (or in addition to) a fixed IP.
Or, failing that, an automatic re-discovery fallback: if the stored IP stops answering on the expected DTLS port range, sweep the local subnet (or a configurable range) for a device that responds with the same OCF device ID / UUID already associated with the config entry, and update the stored IP automatically when a match is found.

Either approach would make the integration resilient to IP churn from router-side DHCP reservation bugs (which do happen — e.g. some AVM FRITZ!Box firmware versions have had issues honoring static DHCP leases behind mesh repeaters) without requiring the user to manually track down the new IP and re-add the device each time.

Thank you.

mbillow:
[...]

PR #472, opened 2026-09-14
Investigation: what these appliances answer on the network, and what they don't

mbillow:
Docs only — no integration code. Split out of #471, which implements the part of #469 that was ready to ship; this is the evidence behind it and the groundwork for what comes next.

Two files: `docs/investigations/network-discovery.md`, and the probe it was measured with, `docs/investigations/ocf_plaintext_probe.py`.

## What it establishes

**Home Assistant's DHCP discovery**, which turns out to be more than packet sniffing: four watchers feeding one function, including an hourly ARP/reverse-DNS sweep of the subnet the `dhcp` integration already performs. That matters for the objection raised in #469 — following a lease needs no scanning on our part, because HA already sweeps. It is also PTR-first, so an appliance behind a router that publishes no reverse names never appears at all.

**What the appliances answer on UDP 5683, with no credentials at all:**

| Resource | Result |
| --- | --- |
| `/oic/d` | the `di` config entries are keyed on (#381), one datagram, seven runs |
| `/oic/p` | manufacturer, model, firmware — enough to name a device before setup |
| `/oic/res?rt=oic.r.doxm` | the DTLS port, 147 bytes, no blockwise |

`/oic/d` and `/oic/p` answer because the device declares them `sec: False` in its own per-link policy, on an appliance that is onboarded and in service — so this is the post-onboarding state, not a window that closes.

**What doesn't work**, which took as long to establish and is the more useful half:

- A hostname cannot address three units that share one, which is exactly #469's install. Ruled out despite being the issue's literal request.
- Multicast discovery cannot cross a router, so a VLAN'd or routed install — the shape a household that segregates IoT devices ends up with — would see it find nothing. Whatever first-time discovery gets built on, it cannot be multicast alone.
- A naive plaintext CoAP reader gets four separate things wrong before any of the above answers: it needs one stable token across a blockwise transfer, has to match the ephemeral source port the device replies from, has to accumulate Block2, and has to retransmit like a CON message.

## On trusting this document

Four claims in here were originally written from what `smartthings-local`'s docstrings and the OCF spec imply, and measurement contradicted every one: that a continuation addressed to 5683 stalls a blockwise read (it was a fresh token), that `/oic/res` advertises `eps` (it advertises none), that the directory therefore cannot name the port (the filtered query does), and that `/oic/res` is unreliable (it was a client that did not retransmit).

The document closes by saying so and pointing at the probe, which is why the probe is vendored rather than described. Run it before trusting a line.

## What it's for

Next steps it makes concrete, none of them in #471: verifying a DHCP sighting by reading `di` at the new address instead of trusting the MAC; a safe OUI matcher for first-time discovery, now that one packet separates an appliance from a Samsung phone; and a one-datagram fast path in front of the config flow's nine-port sweep, which can also name a port outside `49152-49160` that the sweep cannot reach.

Measured on one board — a DW5000-series dishwasher — over seven runs. Still unknown, and recorded as such: whether the ARTIK051-era families answer plaintext at all, and whether any family declares `/oic/d` `sec: True` and so refuses to identify itself.

https://claude.ai/code/session_01Dk5xfT6m3r7JBW9CQqBpjm

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dk5xfT6m3r7JBW9CQqBpjm)_


QuiteYellow (commented):
Item 3 is the one I can add evidence to.

Both stock stacks pick the `/oic/res` representation from the request's Accept option (RT-OCF `utils/rt_data_handler.c:71`, iotivity-lite `api/oc_ri.c:840`), and Accept 60, which your probe sends on every first block, selects the OIC 1.1 form, which has no `eps` key in it. So I measured the other dialect:

| request | dryer, DTLS | oven, DTLS | oven, plaintext |
| --- | --- | --- | --- |
| Accept 60 | 2.05, 1727B, 15 links, 0 `eps` | 2.05, 1629B, 14 links, 0 `eps` | 2.05, 1629B, 14 links, 0 `eps` |
| Accept 10000 | 4.06 | 4.06 | 4.06 |

Accept is the single live variable; adding the content-format-version option (2049) reproduced each row. Both boards refuse the OCF 1.0 dialect outright, and the oven's plaintext and DTLS answers agree to the byte, so the refusal is the device's. The 4.06 is informative in itself: the option parsed and the format was declined, where a 4.00 would mean a malformed request.

Your dishwasher is a different board; one GET with Accept `0x2710` would say whether it answers the same way.

Either way the filtered query is the better default, and it costs no coverage: both stacks honour `rt=` in either dialect, and a doxm link carries whichever form the board speaks. I am reordering `discover_ocf_secure_ports` to match (filtered first, unfiltered directory behind it), so the library will default to the shape item 3 describes.

**Item 1**, the plaintext port. The directory names a port only for links the board marks secure, so the responder port is absent from it, and unicast 5683 is silent on both my appliances. Multicast finds it: the reply's source port is the answer, which is what `ocf_multicast.discover_ocf_responder_ports` returns. It gave 49154 for the dryer and 49153 for the oven, first attempt. Reading them there, unauthenticated:

| | dryer :49154 | oven :49153 |
| --- | --- | --- |
| `/oic/res` | 2.05, 1727B, 2 blocks | 2.05, 1629B, 2 blocks |
| `/oic/res?rt=oic.r.doxm` | 2.05, 149B, 1 block | 2.05, 147B, 1 block |
| `/oic/d` | 2.05, 153B, `di` matching its DTLS read | 2.05, 149B, `di` matching |

Your dishwasher's filtered query is 147 bytes; mine are 147 and 149. So item 1's `di` check and item 3's fast path both hold on two more boards, one datagram each.

Locating that port realistically takes a shared segment, so on a routed install the DHCP sighting still crosses while the port lookup stops at the router. Item 2's OUI matcher inherits it too.

One note on the trust section. Two of the four claims are things `ocf_discovery`'s docstring already agrees with you on: the doxm fallback, and ports "outside a caller's conventional scan set" (`:13-16`, unchanged 0.1.11 to 0.1.17). And on token-versus-port: `_CONTINUATIONS` breaks on the first candidate that answers, so with "stable token, port that answered" first and working, the 5683 and fresh-token variants never went out, which your next bullet already says ("Whether 5683 would also serve them was not established").

PR #483, opened 2026-09-19
Add Slovak translation

OskarSidor:
I've created a new Slovak translation file based on English and Czech translations. No other files were modified.


mbillow:
Thanks! 😊

Issue #176, opened 2026-07-28
Feature Request: Range/Oven Timed cook mode remote start


gonzotek:
Following on as a separate feature request from my initial missing capability report at https://github.com/mbillow/localthings/issues/138#event-28619610925

Via the SmartThing CLI and accompanying API, and with a long lived token taken from before Samsung took those away from end users, I am able to start a timed cooking mode. This would be my ultimate Home Assistant goal: input controls(sliders and/or text fields) for time and set point, and a button to start cooking. I understand if that's simply not a possibility under the constraints this project has to operate with.  Activating this feature is also possible to accomplish from the SmartThings Advanced dashboard at https://my.smartthings.com/advanced/devices

Whether or not this can be done via the local control api you're using is something I can't say; it wouldn't surprise me if it isn't possible due to constraints Samsung/SmartThings enforces, but it would be lovely if it were - I don't want to rely on Samsung's cloud if at all possible.  In my Home Assistant instance, I'm currently using input helpers to control the mode, time and temperature from the Home Assistant UI. A Home Assistant Script uses them as inputs and calls a Rest Command to hit the SmartThings API. That process will probably break in the near future unless I choose to start paying the expected user fee announced to be coming later this year. Rest command example below for illustration purposes(fully understand this probably won't translate to the local api).

```
  oven_start:
    url: "https://api.smartthings.com/v1/devices/ABC-123-54321-ABCDEF-WXYZ/commands"
    content_type: "application/json; charset=utf-8"
    method: POST
    headers:
      authorization: "Bearer ABCDEF-WXYZ-1234"
    payload: |
      [{"component":"main", 
        "capability":"ovenOperatingState",
        "command":"start",
        "arguments":["{{ mode }}",{{ time }},{{ temp }}]
      }]
    # action: rest_command.oven_start
    # data:
    #   temp: 350
    #   mode: Bake
    #   time: 300
```

mbillow:
Could you trigger one of these modes and then take a diagnostic dump? I see where all the modes are defined, but figuring out how to write to start them is going to be the hard part. Hoping seeing one running will help.

gonzotek:
Yes, can do.  Some of the modes (e.g. Broil) act a little differently than others - it doesn't use an integer temp but only a high/low keyword (going by memory at the moment, the actual keyword may be different).  I'll pull together everything I know along with the requested data, but have some personal commitments this week that may make it take a couple of days.  Will definitely followup!

Averyy:
[...]

PR #485, opened 2026-09-20
Record the measured cycle start, and what it does not settle

QuiteYellow:
A cook start has been measured working on one board, so this updates `docs/investigations/oven-cycle-start.md` to record the payload and, at more length, what a single result does not settle.

**The shape.** The cook parameters and the run command go to the device collection in a single write, not to each resource separately:

```
POST /device/0
[
  {"href": "/devices/0"},
  {"href": "/mode/vs/0",  "rep": {"x.com.samsung.da.modes": ["Defrost"]}},
  {"href": "/temperatures/vs/0", "rep": {"x.com.samsung.da.items": [...]}},
  {"href": "/operational/state/vs/0",
   "rep": {"x.com.samsung.da.operationTime": "00:01:00",
           "x.com.samsung.da.state": "Run"}}
]
```

`2.04`, then four seconds later `state: Run`, `/oven/vs/0` reporting `Cooking`, cavity heating. NV7000BS/EUI, single cavity, Remote Control on. Stop is an ordinary single-resource write of `state: "Ready"`, which is why stop has always worked and start has not.

So `Run` is not a separate step on this firmware. It travels inside the collection write, next to the mode and the cook time, and a job is created by that write or not at all. That explains the whole "accepted then discarded" pattern without needing the shadow-rep theory.

**The earlier results are kept, because they still bound the answer. They are also confounded.** Crossing the measurement table against what each board declares in its own `modeSpec`:

| Board | What was tried | Modes it declares startable |
|---|---|---|
| NE9801T-/AA0 (#183) | `state` plus `operationTime`, no mode or setpoint write | 9 of 15 |
| NW9000KD/AA1 (#300) | mode, setpoint and `Run` separately, both orders | no `modeSpec` at all |
| NX9302T-/AA0, gas (#176) | mode, then setpoint, then `Run` | 0 of 8 |

The two boards that received the full sequence are the two that declare nothing startable, so neither result says anything about payload shape. The one that does declare startable modes was never given a mode write. **The separate-write sequence has never been run on a board that can actually start**, so the batch is shown to be sufficient rather than necessary, and the leaf path is untested where it matters rather than refuted.

**Also recorded:** the set duration includes preheat (on my oven). The countdown runs from Start rather than from reaching temperature, so a short cook at a high setpoint is mostly preheat. `progressPercentage` is the same clock at finer resolution (duration/100 against `remainingTime`'s 60 s), which makes it the better field to drive a UI from on any cook under 100 minutes. Both measured on the same board.

Two gotchas. The hour field must be zero-padded: `0:10:00` for a ten-minute cook produced a roughly 609-minute one. And **Remote Control has to be on at the panel, and checked**: with it off the cook parameters still stick and only `state: "Run"` is dropped, both at `2.04`, so a batch that "does not start" says nothing about the payload. I lost a test run to exactly that. It does not survive a power cycle. The doc now says so, because it is the difference between a null result and a void one.

The `/devices/0` marker turned out **not** to be required: the identical batch with that element deleted started a cook first time. The doc records that rather than leaving it open.

Docs only, no code changes here. Two follow-ups I have deliberately left alone as your call:

- `oven.py`'s docstring still says local-OCF cycle start "isn't reproducible on this firmware", which is now wrong for at least one board.
- A start control needs the three cook parameters staged rather than written through, since the appliance will not take them individually from `Ready`. There is a worked version of that in my own bridge if it is useful as a reference; the doc links it.

Happy cooking!


mbillow:
This is awesome! Hopefully this works on the other similar devices: microwaves, ranges, etc. Thanks!

